PR #15, opened 2026-04-13
feat: new-day journal hook (UserPromptSubmit)

brownm09:
## Summary

- Adds `claude/scripts/new-day-journal-check.py` — scans `engineering-journal/sessions/**/` for `*_draft.md` files from prior calendar days
- Wires it as a `UserPromptSubmit` hook in `settings.json` so it runs on every session start
- When a stale draft is detected, outputs a reminder to stdout (injected as context) instructing Claude to run `/journal-compose` before handling the user's request; exits 0 so the prompt is never blocked

## Test plan

- [ ] Start a session with no stale drafts → hook is silent
- [ ] Manually create a `sessions/test/2026-01-01_draft.md`, start a session → hook injects reminder
- [ ] After `/journal-compose` deletes the draft file → hook is silent again on subsequent prompts

### DIFF
--- a/claude/scripts/new-day-journal-check.py
+++ b/claude/scripts/new-day-journal-check.py
@@ -11,12 +11,12 @@ Stdout is injected as context Claude sees before processing the user's message.
 """
 
 import glob
-import json
 import os
 import sys
 from datetime import date
+from pathlib import Path
 
-JOURNAL_REPO = "C:/Users/brown/Git/engineering-journal"
+JOURNAL_REPO = Path.home() / "Git" / "engineering-journal"
 TODAY = date.today().strftime("%Y-%m-%d")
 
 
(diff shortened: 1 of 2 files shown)


claude/scripts/new-day-journal-check.py
```diff
@@ -0,0 +1,55 @@
...
+Stdout is injected as context Claude sees before processing the user's message.
+"""
+
+import glob
+import json
+import os
+import sys
+from datetime import date
+
+JOURNAL_REPO = "C:/Users/brown/Git/engineering-journal"
```
brownm09: Should the repo location be hardcoded?

brownm09: Good catch. Fixed in the follow-up commit — now uses `Path.home() / "Git" / "engineering-journal"`, consistent with the convention in `token-tracker.py` and `token-report.py`. Also dropped an unused `import json` while I was there.

PR #61, opened 2026-04-25
fix: stop false-positive incomplete-journal warnings from stale local draft branches

### LINKED ISSUE #58: Stale local draft/* branches in engineering-journal cause false 'incomplete journal' warnings
## Problem

After a \`draft/YYYY-MM-DD\` branch is squash-merged via the journal-compose workflow, the local branch ref remains in the engineering-journal repo. In future sessions, when Claude encounters \`draft/YYYY-MM-DD\` refs (e.g. from \`git branch\` output), it fires the \"incomplete journal detected\" warning — even though those journals are fully composed and merged.

As of 2026-04-25, the following stale local branches are present in \`engineering-journal\`, all with merged PRs:
- \`draft/2026-04-18\`
- \`draft/2026-04-19\`
- \`draft/2026-04-21\`
- \`draft/2026-04-23\`
- \`draft/2026-04-24\`

## Root cause

Two gaps in the workflow:

1. **Detection rule is too broad.** CLAUDE.md instructs Claude to warn on any \`draft/YYYY-MM-DD\` branch encountered in git output, with no check for whether the branch is still open on the remote. A local-only draft branch means the PR was squash-merged and the local ref was never cleaned up.

2. **No cleanup step after merge.** The journal-compose skill auto-merges the PR but does not delete the local \`draft/YYYY-MM-DD\` branch afterward, leaving stale refs indefinitely.

## Fix

1. **CLAUDE.md — tighten the detection rule:** Only warn if the draft branch exists on the remote (\`git ls-remote --heads origin draft/YYYY-MM-DD\` returns a result). A local-only draft branch should be silently ignored (or pruned).

2. **journal-compose skill — add post-merge cleanup:** After the PR is confirmed merged, delete the local draft branch with \`git branch -D draft/YYYY-MM-DD\`.

brownm09:
## Summary

- **CLAUDE.md**: detection rule now calls `ls-remote` before warning; a local-only `draft/YYYY-MM-DD` branch is silently ignored since it means the PR was already squash-merged
- **journal-compose SKILL.md**: `git branch -d` → `-D` so the post-merge local cleanup actually succeeds (squash-merge leaves commits unreachable from main, causing `-d` to refuse)

Closes #58

## Test plan

- [ ] After merging a journal PR, verify `git branch` in engineering-journal shows no leftover `draft/*` local branch
- [ ] In a fresh session, create a stale local draft branch manually; confirm no warning fires when `ls-remote` returns empty
- [ ] Confirm a genuinely open draft branch (remote exists) still triggers the warning

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -145,7 +145,13 @@ Use that path wherever `sessions/<project>/` appears below.
 
 - **Never compose proactively.** If a `draft/YYYY-MM-DD` branch is encountered during any
   work (e.g., while running `git branch`, checking git status, or reading branch output),
-  emit a single line at the next natural pause:
+  first check whether it exists on the remote before warning:
+  ```bash
+  git -C C:/Users/brown/Git/engineering-journal ls-remote --heads origin draft/YYYY-MM-DD
+  ```
+  A local-only draft branch means the PR was already squash-merged and the local ref was not
+  cleaned up — ignore it silently. Only if the remote branch exists, emit a single line at
+  the next natural pause:
   > "Incomplete journal detected — run `/journal-compose` in a dedicated session."
   Then continue with the user's actual request. Do not read stubs, do not compose, do not
   ask whether to compose.
--- a/claude/skills/journal-compose/SKILL.md
+++ b/claude/skills/journal-compose/SKILL.md
@@ -757,7 +757,7 @@ the stale-branch hook from false-positive firing. Wait for the merge to complete
 delete the local draft branch:
 
 ```bash
-git -C C:/Users/brown/Git/engineering-journal branch -d draft/YYYY-MM-DD 2>/dev/null || true
+git -C C:/Users/brown/Git/engineering-journal branch -D draft/YYYY-MM-DD 2>/dev/null || true
 ```
 
 Tell the user: "Merged: <PR-URL>. Journal published."


brownm09:
## Review: fix: stop false-positive incomplete-journal warnings from stale local draft branches

### Summary

Ready to merge. Both changes are correct and minimal. The `SKILL.md` fix (`-d` → `-D`) directly unblocks the cleanup that was silently failing on every squash-merge. The `CLAUDE.md` detection rule update adds the right gate: `ls-remote` is the authoritative check for whether a branch is still live on origin. No blocking findings. One non-blocking observation and one question.

---

### Non-Blocking Findings

**[reliability]** `claude/CLAUDE.md:148`

The `ls-remote` check is an extra network round-trip that fires whenever any `draft/*` branch is encountered in git output — potentially multiple times per session if the same branch appears in several commands.

It's a low-cost call, but if this check fires frequently it adds latency in proportion to network conditions. The intent is already correctly scoped ("at the next natural pause"), which means it won't block interactive output — this is a note for future awareness if sessions start seeing slowdowns.

**Suggestion:** No action needed now. If it becomes noticeable, cache the result per branch name within a session (e.g., a shell variable check before calling `ls-remote`).

---

### Questions for Author

- The `|| true` on the `-D` command still swallows all errors silently. After the change to `-D`, the only remaining failure case would be if the branch doesn't exist at all (already cleaned up). Is that the full set of expected failure modes, or are there cases where `-D` failing should be visible? If the answer is "any failure is fine to ignore," the `|| true` is correct as-is — just worth confirming the intent is still to suppress all output.

PR #69, opened 2026-04-25
feat: auto-sync dev-env repo on every prompt to prevent stale CLAUDE.md

### LINKED ISSUE #68: Local dev-env repo goes stale — symlinked CLAUDE.md silently serves outdated rules
## Problem

`~/.claude/CLAUDE.md` is a symlink to `C:/Users/brown/Git/dev-env/claude/CLAUDE.md`. When the local dev-env repo falls behind origin/main, every Claude Code session silently reads stale rules — merged changes (new skills, updated workflow rules, etc.) have no effect until someone manually runs `git pull` in the dev-env repo.

Discovered during the 2026-04-25 session: the local repo was 4 PRs behind origin/main, meaning the "stop after PR creation" rule from #60 was never visible to sessions even though the PR had been merged.

## Impact

- Merged CLAUDE.md changes are invisible until a manual pull
- No indication to the user or to Claude that the file is stale
- Silently causes rule non-compliance (as seen with #60 / #67)

## Possible mitigations

1. **Auto-pull hook** — add a `UserPromptSubmit` or session-start hook that runs `git -C C:/Users/brown/Git/dev-env pull --ff-only origin main` (fast-forward only, fails loudly if diverged)
2. **Staleness warning** — hook checks `git -C ... rev-parse HEAD` vs `origin/main` and emits a warning if behind
3. **CLAUDE.md header** — add a comment at the top of the file reminding Claude to check/pull dev-env at session start if the task involves CLAUDE.md changes

Option 1 (auto-pull hook) is the most robust. Option 3 was already added as a linter comment in the file header.

brownm09:
Closes #68

## Summary
- Adds `claude/scripts/dev-env-sync.py`: a `UserPromptSubmit` hook that runs `git fetch` + fast-forward pull on the local dev-env repo at the start of every Claude Code session
- Registers the script as the first `UserPromptSubmit` hook in `claude/settings.json` so tooling is always current before journal-check and turn-count run
- Commits CLAUDE.md workflow rule updates (stop-after-PR, stop-after-merge, journal auto-stub) that were applied to the local file during the #68 incident session but never pushed to origin

## Behavior
- **Up-to-date:** silent — no output, exits 0
- **Behind:** pulls fast-forward, emits a one-liner with commit count and titles (up to 5)
- **Diverged:** emits a warning and exits 0 without attempting a merge
- **Network failure:** silent — never blocks the user's prompt

## Test plan
- [ ] Verify silent exit when repo is current
- [ ] Manually reset local dev-env HEAD behind origin and confirm the hook pulls and reports
- [ ] Confirm diverged scenario emits warning without blocking

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -296,7 +296,8 @@ Subsequent stubs begin directly at `<!-- session: <slug> -->`.
 ### Update triggers
 
 **Project journal** (`sessions/<project>/`):
-- Add to the current session's stub when a PR is merged or a strategic decision is made
+- **Auto-create stub without user prompt on these session-end events:** PR opened, PR merged — follow the Stub file workflow immediately, then stop
+- Add to the current session's stub when a strategic decision is made mid-session
 - Compose and publish the daily document at end of last session of the day
 
 **Meta journal** (`sessions/meta/`):
--- a/claude/scripts/dev-env-sync.py
+++ b/claude/scripts/dev-env-sync.py
@@ -6,6 +6,10 @@ Runs a fast-forward pull at session start so that CLAUDE.md and other
 symlinked tooling always reflect the latest merged changes. Silent on
 success; emits a warning if the repo has diverged and needs manual attention.
 
+Only acts when the repo has `main` checked out — exits silently when a
+feature branch is active so dev-env PR sessions are not spammed with
+false-positive divergence warnings.
+
 Exit 0 always — never block the user's prompt.
 """
 
@@ -33,14 +37,23 @@ def main() -> None:
     except Exception:
         pass
 
+    # Guard: repo must exist at the expected path.
+    if not DEV_ENV_REPO.is_dir():
+        sys.exit(0)
+
+    # Only act when main is checked out — feature branches are intentional.
+    branch = run(["git", "symbolic-ref", "--short", "HEAD"])
+    if branch.returncode != 0 or branch.stdout.strip() != "main":
+        sys.exit(0)
+
     # Fetch quietly so the local remote-tracking ref is current.
     fetch = run(["git", "fetch", "origin", "main", "--quiet"])
     if fetch.returncode != 0:
         # Network issue — don't block, don't spam on every turn.
         sys.exit(0)
 
-    # Compare HEAD to origin/main.
-    rev_local = run(["git", "rev-parse", "HEAD"])
+    # Compare local main to origin/main.
+    rev_local = run(["git", "rev-parse", "refs/heads/main"])
     rev_remote = run(["git", "rev-parse", "origin/main"])
     if rev_local.returncode != 0 or rev_remote.returncode != 0:
         sys.exit(0)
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: feat: auto-sync dev-env repo on every prompt to prevent stale CLAUDE.md

### Summary
Needs one revision before merge. The hook logic is sound for the common case (dev-env repo sitting on `main`), but it produces spurious "diverged" warnings every turn whenever the repo has a non-`main` branch checked out — which happens routinely during dev-env PR development. Since this hook runs on every prompt, that's high-frequency noise during exactly the sessions where you're most actively working. All other aspects — network-failure handling, fast-forward safety, exit-0 guarantee, settings wiring, and the CLAUDE.md update-triggers clarification — are correct and well-structured.

---

### Blocking Findings

**[correctness]** `claude/scripts/dev-env-sync.py:56–63`

The diverged-branch check compares `merge-base HEAD origin/main` against `HEAD`. When the dev-env repo has a feature branch checked out (e.g., during any dev-env PR session), `HEAD` is the feature tip, `base` is the branch point, and `base != local` is always true — so the script prints the "WARNING: local dev-env repo has diverged" message on every single prompt for the duration of that session. This is the exact scenario that occurs when iterating on dev-env itself, so the warning fires most when the user is least confused — and least when the repo is genuinely stale.

**Fix:** Before the comparison block, detect whether `HEAD` is on `main`. The simplest approach: `git symbolic-ref --short HEAD` → if not `"main"`, exit 0 silently (the user is intentionally on another branch; main-tracking is not applicable). Alternatively, compare `refs/heads/main` to `origin/main` directly instead of `HEAD`, so the check is always about main regardless of what's checked out.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/dev-env-sync.py:76`

When more than 5 commits are pulled, only the first 5 are shown with no indication that the list is truncated. A pull of 8 commits prints the same visual weight as a pull of 3.

**Suggestion:** Append `f"  ... and {count - 5} more"` to the output when `count > 5`.

**[maintainability]** `claude/scripts/dev-env-sync.py:15`

`DEV_ENV_REPO` is constructed as `Path.home() / "Git" / "dev-env"`. If the repo doesn't exist at that path (different machine, different layout), all `git` calls return non-zero and the hook exits silently — which is safe but provides no diagnostic. This is fine for the current single-machine setup but will confuse any future portability attempt.

**Suggestion:** Add a one-line existence check at the top of `main()`: `if not DEV_ENV_REPO.is_dir(): sys.exit(0)` — keeps the silent exit, but makes the intent explicit and guards against accidental no-ops.

---

### Questions for Author

- When `sys.stdin.read()` is called, does the Claude Code hook framework close stdin after delivering the payload, or does it leave it open? If it leaves it open, this call blocks until the process is killed. The `except Exception: pass` would never fire in a hang scenario — it only catches errors, not indefinite blocking. Is there a pattern from the other hooks in this repo that should be followed instead (e.g., reading a fixed number of bytes, or using `sys.stdin.readline()`)?

---

### Style Notes

The variable name `l` in the list comprehension on line 73 (`lines = [l for l in log.stdout.strip().splitlines() if l]`) shadows the built-in and is easy to misread as `1`. Rename to `line` to match the outer variable `lines`.

PR #71, opened 2026-04-25
docs: explicitly exclude review sessions from auto-stub creation

brownm09:
## Summary
- Adds a rule to CLAUDE.md that review-only sessions (`/review --post-comment`) do **not** auto-create a journal stub
- Review is not a session boundary — only PR creation and PR merge trigger stubs

## Test plan
- [ ] Verify CLAUDE.md diff is a single added line under Update triggers

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -301,6 +301,7 @@ Subsequent stubs begin directly at `<!-- session: <slug> -->`.
   - PR opened — follow the Stub file workflow immediately, then stop (see Git Workflow → Stop after PR creation)
   - PR merged (including auto-merge) — follow the Stub file workflow immediately, then stop (see Git Workflow → Stop after PR merge)
   - PR closed without merging — follow the Stub file workflow; stopping is optional (see Git Workflow → PR closed without merging)
+- Review-only sessions (`/review --post-comment`) do **not** auto-create a stub — review is not a session boundary
 - Add to the current session's stub when a strategic decision is made mid-session
 - Compose and publish the daily document at end of last session of the day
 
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: docs: explicitly exclude review sessions from auto-stub creation

### Summary

Ready to merge. The change is minimal and unambiguous: it adds two negative-case rules under "Update triggers" that clarify what does *not* create a journal stub. The rules are correctly placed, accurately scoped, and consistent with the existing surrounding conventions. No blocking or reliability concerns. One minor structural note below.

---

### Non-Blocking Findings

**[maintainability]** `claude/CLAUDE.md:304`
The new negative-case block is inserted in the middle of the bullet list for auto-create events, then the two positive triggers ("Add to the current session's stub…" and "Compose and publish…") resume immediately after. A reader skimming the positive trigger list may not notice they've crossed into a different intent category. The negative cases would be easier to distinguish if the "do not auto-create" block were followed by a blank line before the two positive continuation bullets, or if those bullets were rendered as a sibling section header like `**Note:** The following…` rather than a plain list item embedded in the trigger list.

**Suggestion:** Add a blank line between the "do not auto-create" block and the two bullets that follow it, or move those continuation bullets below the negative-case list so the structure reads: positive triggers → negative triggers → other instructions.

---

### Questions for Author

- The second negative case ("Pushing commits to an existing PR…") ends with a conditional: "if the session continues to a merge, the merge creates the stub." Should this also handle the case where the session ends without a merge — i.e., does a push-only session that terminates without a merge produce *no* stub at all? If so, it may be worth a brief note clarifying that outcome explicitly, since it's a new named exception.

---

### Style Notes

The heading `The following do **not** auto-create a stub — they are not session boundaries:` uses a dash with surrounding spaces (` — `) as a separator. The surrounding text uses `**bold**` for emphasis in similar contexts. Consistent — no issue.

PR #73, opened 2026-04-27
feat: add Step 0.5 plan-then-optimize to journal-compose skill

### LINKED ISSUE #72: Add Step 0.5 (plan-then-optimize) to journal-compose skill
## Problem

The plan-then-optimize protocol (CLAUDE.md § Context & Token Efficiency) requires a numbered plan + token-efficiency revision pass before any Agent spawn, skill invocation, or multi-file read. The 2026-04-26 scheduled journal-compose run violated all three requirements:

- No plan stated before acting
- No revision pass
- Redundant stub reads before spawning subagents (manifests already provided enough context)
- Journal preview reads before grep (preview was redundant once grep was used)

Result: ~340–370k tokens instead of an estimated ~220–260k. The rule existed in CLAUDE.md but was still skipped — proving that rules alone are structurally insufficient.

## Root cause

The skill loads and begins executing before the plan-then-optimize protocol is applied. There is no forcing function in the execution path itself — only a rule in CLAUDE.md that must be recalled and applied voluntarily.

## Proposed fix

Add a mandatory **Step 0.5** to `claude/skills/journal-compose/SKILL.md`, immediately before Step 1, that requires a written plan + revision pass before any tool call:

```markdown
## Step 0.5 — Plan-then-optimize (required)

Before reading any file or spawning any agent, write out:
1. The steps you are about to execute (numbered)
2. Which file reads can be skipped (manifest vs. stubs, grep vs. full read)
3. Which agent spawns will be batched in a single message

Do not proceed to Step 1 until this plan is written. No tool calls before the revision pass completes.
```

This creates a forcing function in the skill's execution path rather than relying on the model recalling and applying a CLAUDE.md rule before the skill begins.

## Acceptance criteria

- [ ] Step 0.5 added to `journal-compose/SKILL.md` before Step 1
- [ ] Step 0.5 also added to the subagent prompt template in Multi-project mode (Phase 1), scoped to the subagent's local steps
- [ ] Verified in the next scheduled compose run that a plan is written before the first tool call

brownm09:
Closes #72

Adds a mandatory Step 0.5 to `journal-compose/SKILL.md` that requires a written plan + token-efficiency revision pass before any tool call. The step is added in two places:

1. Main skill flow — between Step 0 (session isolation check) and Step 1 (locate stubs)
2. Multi-project subagent prompt template — before Step 1 of each subagent, scoped to that subagent's local steps

This creates a forcing function in the execution path rather than relying on the model recalling CLAUDE.md rules voluntarily.

### DIFF
--- a/claude/skills/journal-compose/SKILL.md
+++ b/claude/skills/journal-compose/SKILL.md
@@ -26,6 +26,15 @@ If any user turns or assistant responses exist in the conversation before the
 
 Do not proceed to Step 1.
 
+## Step 0.5 — Plan-then-optimize (required)
+
+Before reading any file or spawning any agent, write out:
+1. The steps you are about to execute (numbered)
+2. Which file reads can be skipped (manifest vs. stubs — prefer manifest; grep vs. full read — prefer grep)
+3. Which agent spawns will be batched in a single message with `run_in_background: true`
+
+Do not proceed to Step 1 until this plan is written. No tool calls before the revision pass completes.
+
 ## Step 1 — Locate stubs and acquire compose lock
 
 **Determine the date:**
@@ -137,6 +146,12 @@ You are composing the engineering journal for one project. Follow these steps ex
 **Date:** YYYY-MM-DD
 **Project path:** sessions/<project>/
 **Stub files (in order):** <stub1>, <stub2>, ...
+
+Step 0.5 — Plan-then-optimize (required). Before any tool call, write out:
+  1. The steps you will execute (numbered)
+  2. Which reads can be skipped (manifest vs. stubs — prefer manifest; grep vs. full read)
+  3. Confirm no sequential tool calls exist that could run in parallel
+  Do not proceed until this plan is written.
 
 Step 1 — Acquire compose lock for this project using this project-scoped path:
   C:/Users/brown/Git/engineering-journal/sessions/<project>/.draft-compose.lock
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: feat: add Step 0.5 plan-then-optimize to journal-compose skill

### Summary

Ready to merge with one minor tweak. The change correctly adds a forcing-function plan step before any tool calls, and the gate language ("Do not proceed until...") is well-placed. The single notable issue is that the subagent prompt template inherits a "prefer grep" instruction that conflicts with subagents' actual need to read full stub files during composition.

---

### Non-Blocking Findings

**[reliability]** `claude/skills/journal-compose/SKILL.md:154`
The subagent template's point 2 says "grep vs. full read — prefer grep." Subagents composing the journal must read stub files in full — their entire purpose is to ingest stub content and synthesize it. A subagent that follows this instruction and greps instead of reading will silently produce an incomplete journal entry by missing sections outside the matched lines.
**Suggestion:** Change the subagent template's point 2 to "Which reads can be skipped entirely (e.g., if manifest data is sufficient, skip the stub read)" — emphasizing skip vs. read rather than grep vs. read.

**[maintainability]** `claude/skills/journal-compose/SKILL.md:26,147`
Step 0.5 now exists in two forms with slightly different wording and different item 3 language (main: "agent spawns batched with `run_in_background: true`" vs. subagent: "no sequential tool calls that could run in parallel"). These will diverge silently on future edits.
**Suggestion:** Add a one-line comment above the subagent copy (`# mirrors Step 0.5 in main flow — keep in sync`) so the relationship is explicit.

---

### Questions for Author

- Is the difference in item 3 intentional? The main skill targets agent spawns specifically; the subagent template targets all sequential tool calls. For a subagent that doesn't spawn further agents, the broader framing in the template seems correct — but if so, should the main skill's item 3 be updated to match?

PR #74, opened 2026-04-27
fix: transcript-based context warning replaces unreachable turn-50 threshold

brownm09:
## Summary

- **Root cause**: the old `turn-count-hook.py` warned at 50 user prompts; in practice no session exceeded 10, so the hook never fired
- **Primary signal (new)**: reads the transcript JSONL on each `UserPromptSubmit`, extracts the last assistant turn's `input + cache_read + cache_creation` token count, warns at 40K and every 20K after — stateful via `ctx-warn-<session_id>.txt` scratch files to avoid repeating on the same boundary
- **Secondary signal (lowered threshold)**: prompt count now warns at 8 (was 50), repeats every 4; warning copy explicitly frames it as a weak/less-accurate signal

## Test plan

- [x] Ran against real 76K-token transcript: warns at 40K, escalates to 60K on next prompt, then silent until 80K
- [x] Ran against 22K-token transcript: no output
- [x] Prompt count: silent for prompts 1–7, fires at 8, fires again at 12, silent at 13
- [x] Exit always 0

### DIFF
--- a/claude/scripts/turn-count-hook.py
+++ b/claude/scripts/turn-count-hook.py
@@ -19,6 +19,8 @@ Exit 0 always — never block the user's prompt.
 Stdout is injected as context Claude sees before processing the user's message.
 """
 
+from __future__ import annotations
+
 import json
 import os
 import sys
@@ -40,6 +42,7 @@ COUNTER_MAX_AGE_DAYS = 30
 
 
 def load_prompt_threshold(cwd: str) -> int:
+    """Return turn threshold from project hook-config.json, or the default."""
     path = os.path.join(cwd, CONFIG_FILE)
     try:
         with open(path, encoding="utf-8") as f:
@@ -50,6 +53,7 @@ def load_prompt_threshold(cwd: str) -> int:
 
 
 def cleanup_stale_counters() -> None:
+    """Delete counter and ctx-warn files older than COUNTER_MAX_AGE_DAYS."""
     cutoff = time.time() - COUNTER_MAX_AGE_DAYS * 86400
     try:
         for f in SCRATCH.glob("turn-count-*.txt"):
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: fix: transcript-based context warning replaces unreachable turn-50 threshold

### Summary

Ready to merge. The two-signal design is sound, the state-file approach correctly prevents re-firing on the same boundary, and the test plan covers the key paths. One non-blocking concern (PEP 604 type syntax portability), one UX quirk worth calling out, and one question about the hook payload field name.

---

### Blocking Findings

None.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/turn-count-hook.py:83,108`
Return type annotations use PEP 604 union syntax (`int | None`, `str | None`), which requires Python ≥ 3.10. The current machine runs 3.10.11 so it works today, but the syntax will crash silently at function-definition time on 3.9 if this hook is ever run in a different environment (e.g., CI container, another dev machine).
**Suggestion:** Use `Optional[int]` / `Optional[str]` from `typing` for wider compatibility, or add `from __future__ import annotations` at the top to make all annotations lazy strings — either is a one-line fix.

**[reliability]** `claude/scripts/turn-count-hook.py:121-135`
When context is already above a threshold at the start of a session (e.g., a long transcript loaded on the first prompt), `last_warned` starts at 0, so the hook fires at 40K. On the very next prompt (same token count), `last_warned=40000`, `next_threshold=60000`, still above → fires again at 60K. The user sees two consecutive warnings citing the same token count but different threshold numbers, which can read as contradictory ("was it 40K or 60K?"). The test plan documents this as the expected catch-up behavior, but a user encountering it cold may be confused.
**Suggestion:** Either emit a combined warning when `tokens` exceeds multiple thresholds in a single call (advance `next_threshold` in a loop before returning), or add a clause to the warning text like "caught up from prior boundary" when `last_warned == 0` and `tokens > CONTEXT_WARN_TOKENS`.

---

### Questions for Author

- Is `transcript_path` a documented field in the Claude Code `UserPromptSubmit` hook payload, or was it inferred empirically by inspecting a real hook invocation? If it's undocumented/internal, it may silently disappear in a future Claude Code release and Signal 1 would stop firing with no error — a comment referencing where this field comes from would protect the next reader.

---

### Style Notes

`load_prompt_threshold` and `cleanup_stale_counters` had brief docstrings in the original that were dropped in the rewrite. The functions are short enough that this is fine, but the pattern is now inconsistent — `get_current_context_tokens` and `check_context_tokens` have full docstrings while the helpers don't. Worth either adding one-liners back to the helpers or stripping the long docstrings from the new functions to match.

PR #76, opened 2026-04-27
feat: mandate test verification before PR creation

brownm09:
## Summary
- Adds a **`## Testing` requirement** to every project CLAUDE.md — without it Claude cannot comply with the test mandate and falls back to undocumented manual checks
- Adds a **\"Test before PR\" rule** to the global Git Workflow section, requiring Claude to run the project test command and document results in the PR body before opening a PR
- Adds **`pre-pr-create-check.py`** — a PreToolUse hook that fires a three-point checklist `systemMessage` whenever `gh pr create` appears in a Bash call, surfacing the rule as a runtime reminder
- Hardens the missing-section fallback to a **hard stop** — Claude must ask the user to add `## Testing` before `gh pr create` is allowed to run

## Related
- #79 — add `## Testing` section to dev-env CLAUDE.md itself (follow-on, unblocked after this merges)

## Test plan
- [ ] Hook fires when a Bash call contains `gh pr create` and outputs the checklist systemMessage
- [ ] Hook does not fire for unrelated Bash commands
- [ ] CLAUDE.md changes render correctly (check Per-Project Requirements section and Git Workflow bullet)

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -42,8 +42,28 @@ Before writing a `gh` or other CLI automation script:
 
 ---
 
+## Per-Project CLAUDE.md Requirements
+
+Every project CLAUDE.md **must** include a `## Testing` section specifying the command(s) used to verify a solution before opening a PR. Examples:
+
+```markdown
+## Testing
+Run `npm test` to execute the full test suite.
+Run `npm run build` first if touching compiled output.
+```
+
+```markdown
+## Testing
+Run `pytest` from the repo root. Integration tests require `DATABASE_URL` set.
+```
+
+If the project has no automated tests, the section must say so explicitly and describe the manual verification steps instead. The `## Testing` section is used by the global "Test before PR" rule — without it, Claude cannot comply with that rule and will fall back to an undocumented manual check.
+
+---
+
 ## Git Workflow
 
+- **Test before PR.** Before running `gh pr create`, execute the project's test command defined in `## Testing` in the project CLAUDE.md. Tests must pass (or the failure must be explained and documented). Include what was tested and the outcome in the PR body. If no project test command is defined, state that explicitly and describe any manual verification performed instead.
 - **Never commit directly to `main`.** All changes go through a branch and PR, regardless of repo.
 - **Branch naming:** `feat/`, `fix/`, `config/`, `chore/`, `draft/` prefixes — match the convention already in use in the repo.
 - **PR first, then merge.** Open the PR immediately after pushing the branch; do not prompt the user to run `gh pr create` themselves.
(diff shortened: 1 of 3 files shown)


brownm09:
## Review: feat: mandate test verification before PR creation

### Summary

The PR is close to mergeable but has one reliability gap that could block automated workflows. The hook script is clean and correctly scoped to `gh pr create` Bash calls; the CLAUDE.md changes are well-placed. The main concern is that the "Test before PR" hard-stop rule has no carve-out for the automated journal stub PR workflow, which creates PRs without user interaction and could be stopped by the new rule if the engineering-journal repo lacks a `## Testing` section. Confirm that case is covered before merging.

---

### Blocking Findings

**[reliability]** `claude/CLAUDE.md:55`

The "Test before PR" rule says "If no `## Testing` section exists in the project CLAUDE.md, stop and ask the user to add one — do not open the PR until it is present." The Engineering Journal stub workflow (`/journal-compose` and every session-end PR open) runs `gh pr create` in the engineering-journal repo and is explicitly required to fire "without user prompt." If engineering-journal's CLAUDE.md lacks `## Testing`, these two instructions will conflict: every journal PR will halt and demand user input, breaking the automation guarantee.

**Fix:** Either (a) verify that engineering-journal already has `## Testing` and document that dependency in the PR body, or (b) add an explicit exemption to the rule — e.g., "Exception: `draft/YYYY-MM-DD` branches and repos with no project CLAUDE.md are exempt from this requirement" — so the intent is unambiguous.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/pre-pr-create-check.py:18`

The module docstring says "Does NOT block the PR creation (exit 0). The checklist appears in the Claude Code UI as a hard reminder; the CLAUDE.md testing mandate is the enforcement mechanism." But the CLAUDE.md section uses "must," "hard stop," and "Do not open the PR until." These two descriptions of enforcement level — advisory hook vs. mandatory stop — will confuse a future reader trying to understand what actually happens when tests are missing.

**Suggestion:** Add a single comment in CLAUDE.md or the hook docstring explaining the layered design: the hook reminds, the instruction enforces. Something like "Enforcement relies on instruction-following; the hook is a secondary reminder, not a gate."

---

### Questions for Author

- Does the engineering-journal repo currently have a CLAUDE.md, and if so, does it include `## Testing`? The answer determines whether the blocking finding above is already resolved or still open.
- The regex `_GH_PR_CREATE_RE` doesn't match `| gh pr create --body-file -` (single pipe, not `||`). Is piped input to `gh pr create` used anywhere in the journal or other workflows? If so, the hook would miss those calls.

---

### Style Notes

The Python script is well-structured — clear module docstring, early-exit pattern, and a named regex constant. The CLAUDE.md placement (between CLI Scripting Checklist and Git Workflow) is logical. Minor: the `Per-Project CLAUDE.md Requirements` section heading uses bold inside a `**must**` in prose, which reads fine but mixes with the requirement examples. No action needed.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #81, opened 2026-04-27
fix: remove bash -c wrapper from hook commands

brownm09:
## Summary

- `bash` is not on PATH in Claude Code hook execution environment on Windows
- The `PreToolUse` hook wrapped every Python script in `bash -c '...'`, which exited non-zero and blocked all Bash tool calls
- `UserPromptSubmit`/`PostToolUse`/`Stop` hooks had the same wrapper but failed silently (non-zero exits there do not block)
- Fix: call `python3 <script>` directly

## Root cause

Claude Code resolves hook commands against the system PATH, not the Git Bash PATH. `bash.exe` lives at `C:\Program Files\Git\bin\bash.exe` but is not in the Windows system PATH.

## Test plan

- Verified `python3` is available without the bash wrapper: both PreToolUse scripts exit 0 when invoked as `python3 <script>`
- `bash` confirmed absent from system PATH (`where.exe bash` returns exit 1)

No linked issue -- meta/config change with no meaningful issue scope.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/claude/settings.json
+++ b/claude/settings.json
@@ -7,11 +7,11 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash -c 'python3 C:/Users/brown/.claude/scripts/pre-commit-branch-check.py'"
+            "command": "python3 C:/Users/brown/.claude/scripts/pre-commit-branch-check.py"
           },
           {
             "type": "command",
-            "command": "bash -c 'python3 C:/Users/brown/.claude/scripts/pre-pr-create-check.py'"
+            "command": "python3 C:/Users/brown/.claude/scripts/pre-pr-create-check.py"
           }
         ]
       }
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: fix: remove bash -c wrapper from hook commands

### Summary
Ready to merge. The fix is correct and well-motivated: removing the `bash -c` wrapper from all hook commands resolves the root cause (bash not on the Windows system PATH). One ambiguity worth confirming before merge: `pre-pr-create-check.py` is added to the PreToolUse hooks in this diff but does not appear in the diff itself — since PreToolUse is a blocking hook, a missing script would silently break all relevant tool calls.

---

### Blocking Findings

None.

---

### Non-Blocking Findings

**[reliability]** `claude/settings.json` (new PreToolUse entry)
`pre-pr-create-check.py` is registered as a PreToolUse hook, but its file is not part of this diff. PreToolUse hooks that exit non-zero block the matched tool call — the PR body says "both PreToolUse scripts exit 0," which implies the file exists, but it's not visible here.
**Suggestion:** Either confirm the file was committed in an earlier PR and link that commit/PR, or add it to this diff. A comment in the PR body pointing to the commit where the script was introduced would suffice.

---

### Questions for Author

- Was `pre-pr-create-check.py` committed in a previous PR, or is it a new file that should be in this diff? The test plan references "both PreToolUse scripts," suggesting it exists — but for traceability, it would help to know which PR introduced it.

---

### Style Notes

The PR body is well-structured — root cause, fix, and test plan are all present. No style concerns.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

brownm09:
**Follow-up (post-review check):** The `pre-pr-create-check.py` question from my earlier comment is resolved — PR #76 merged on 2026-04-28 and the script is now on `main`. No merge-order dependency remains. PR #81 is clear to merge.

PR #84, opened 2026-04-28
feat(journal-compose): validate JSONL before composition begins

brownm09:
## Summary
- Adds **Step 0.8** to the `journal-compose` skill, which runs `node scripts/validate-jsonl.js` before any stubs or manifests are read
- Catches malformed manifest/open-prs lines at the earliest possible point — before the compose lock is acquired — so failures surface cleanly with an actionable error message rather than as a mysterious gap in the composed output

## Why before Step 1
The manifest files are the primary input to composition (session count, topics, token data, PR lifecycle). A malformed line in a manifest silently drops or corrupts that session's data. Validating first means composition only runs on confirmed-clean input. It also avoids acquiring the compose lock on broken state, making cleanup trivial (no lock file to clear).

## Related
- Closes no issue (meta/config change — no meaningful issue scope per global CLAUDE.md)
- Companion to `engineering-journal` change adding `## Testing` and `scripts/validate-jsonl.js`
- Addresses the review finding from dev-env#76 (primary gate should be at composition time, not PR time)

### DIFF
--- a/claude/skills/journal-compose/SKILL.md
+++ b/claude/skills/journal-compose/SKILL.md
@@ -40,6 +40,8 @@ Do not proceed to Step 1 until this plan is written. No tool calls before the re
 Before reading any stubs or manifests, run the JSONL validator:
 
 ```bash
+[ -f "C:/Users/brown/Git/engineering-journal/scripts/validate-jsonl.js" ] || \
+  { echo "validate-jsonl.js not found — merge the companion engineering-journal PR first"; exit 1; }
 node "C:/Users/brown/Git/engineering-journal/scripts/validate-jsonl.js"
 ```
 
--- a/claude/skills/journal-compose/SKILL.md
+++ b/claude/skills/journal-compose/SKILL.md
@@ -35,6 +35,19 @@ Before reading any file or spawning any agent, write out:
 
 Do not proceed to Step 1 until this plan is written. No tool calls before the revision pass completes.
 
+## Step 0.8 — Validate JSONL files
+
+Before reading any stubs or manifests, run the JSONL validator:
+
+```bash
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: feat(journal-compose): validate JSONL before composition begins

### Summary

Ready to merge with one suggestion. The change is well-scoped and correctly placed before the lock acquisition step. The single concern is that if the companion `engineering-journal` change (adding `scripts/validate-jsonl.js`) hasn't been merged yet, Step 0.8 fails with a generic Node "Cannot find module" error rather than an actionable one. No blocking issues.

---

### Non-Blocking Findings

**[reliability]** `claude/skills/journal-compose/SKILL.md` — Step 0.8 block
The step runs `node .../validate-jsonl.js` unconditionally. If the companion `engineering-journal` PR hasn't been merged yet, the error will be `Cannot find module '...validate-jsonl.js'` — which points at Node internals rather than the actual problem (missing companion change). Since the two repos are updated independently, this window may be wider than expected.
**Suggestion:** Add a single guard line before the `node` call: `[ -f "C:/Users/brown/Git/engineering-journal/scripts/validate-jsonl.js" ] || { echo "validate-jsonl.js not found — merge the companion engineering-journal PR first"; exit 1; }`. This converts a cryptic module error into an actionable message.

---

### Questions for Author

- Has the companion `engineering-journal` change (adding `scripts/validate-jsonl.js` and the `## Testing` section) already been merged? The PR description says it exists but doesn't link to it — worth confirming before merging this side.

---

### Style Notes

Step 0.8 numbering is consistent with the existing 0.5 slot and the `---` separator matches surrounding steps. No issues.

PR #86, opened 2026-04-28
docs: add outcome-correctness revision pass to plan-then-optimize protocol

### LINKED ISSUE #85: docs: add outcome-correctness revision pass to plan-then-optimize protocol
The current plan-then-optimize protocol has a token-efficiency revision pass but no guard against that revision silently breaking intended behavior. Parallelizing two steps with an implicit ordering dependency, or dropping a file read a downstream step needs, can produce a plan that is cheaper but wrong.

Add a second revision pass (Pass 2 — Outcome correctness) that checks:
- No ordering dependency was broken by parallelization
- No required read was dropped
- No Agent scope was narrowed past what is needed
- Final outputs still match the original intent

Affects: `claude/CLAUDE.md` and `memory/feedback_plan_then_optimize.md`

brownm09:
Adds **Pass 2 — Outcome correctness** to the plan-then-optimize protocol.

The token-efficiency revision (Pass 1) can silently break intended behavior:
- Parallelizing two steps that have an implicit ordering dependency
- Dropping a file read that a downstream step actually needs
- Narrowing Agent scope past what is required for correctness

Pass 2 is an explicit check after Pass 1 that verifies the optimized plan still produces the intended result before any action is taken.

Closes #85

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -126,11 +126,19 @@ Default to Sonnet when uncertain. Never use Opus for tasks a Haiku prompt handle
 
 **Mechanical operations:** If a task is fully scriptable with known inputs, write the script rather than running an interactive session. Candidate operations: stale PR remediation, branch cleanup, rebase-and-merge sequences. Use `~/.claude/scripts/merge-stale-pr.sh` for engineering-journal stale draft PRs.
 
-**Plan-then-optimize before acting:** Any task involving an Agent spawn, a skill invocation, or reads across more than one file requires this protocol. State a numbered plan first, then do one explicit token-efficiency revision pass before taking any action. The revision pass must check:
+**Plan-then-optimize before acting:** Any task involving an Agent spawn, a skill invocation, or reads across more than one file requires this protocol. State a numbered plan first, then apply two explicit revision passes before taking any action.
+
+**Pass 1 — Token efficiency:** check:
 - Sequential tool calls that can be parallelized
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: docs: add outcome-correctness revision pass to plan-then-optimize protocol

### Summary

Ready to merge with a minor clarification. The change is well-scoped, closes a real behavioral gap (Pass 1 optimizations could silently break ordering or data-dependency invariants), and the four Pass 2 checks directly address the failure modes described in the PR body. One question about the PR body's claim regarding `memory/feedback_plan_then_optimize.md` — that file isn't in the diff, and memory files are machine-local by design, so either the PR body is slightly inaccurate or the update happened outside version control.

---

### Questions for Author

- The PR body states "Also updates `memory/feedback_plan_then_optimize.md` to reflect both passes," but `changedFiles` is 1 (only `claude/CLAUDE.md`). Since `~/.claude/projects/.../memory/` is machine-local and not version-controlled, this file can't appear in the diff. Was that update applied as a local memory write during the session, or is the PR body claim an oversight?

---

### Style Notes

The Pass 2 header lead-in is wordier than Pass 1's:
- Pass 1: `**Pass 1 — Token efficiency:** check:`
- Pass 2: `**Pass 2 — Outcome correctness:** after the efficiency revision, verify the optimized plan still produces the intended result:`

Trimming to `**Pass 2 — Outcome correctness:** verify:` would parallel Pass 1's structure. Minor — not a blocker.

PR #91, opened 2026-04-29
fix: extend plan-then-optimize rule to mid-session task switches

### LINKED ISSUE #90: Extend plan-then-optimize rule to mid-session task switches
## Problem

The current rule triggers only at the start of a task (Agent spawn, skill invocation, multi-file reads). When a session pivots mid-stream — e.g., from `/review` output to implementing findings — the protocol is skipped. That pivot is exactly where unplanned sequential tool calls and redundant Agent spawns tend to appear.

## Solution

Extend the trigger condition in `claude/CLAUDE.md` to also fire on mid-session task switches.

brownm09:
## Summary

- Extends the plan-then-optimize trigger in `claude/CLAUDE.md` to also fire when switching to a new task within the same session (e.g., moving from `/review` or another skill output to implementing findings)
- Updates the memory file to match the expanded rule

## Test plan

- [ ] In a future session: run a skill, then pivot to implementation work — verify the plan + revision pass fires at the task-switch boundary before any file writes
- [ ] Memory index reflects the updated trigger description

Closes #90

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -127,7 +127,7 @@ Default to Sonnet when uncertain. Never use Opus for tasks a Haiku prompt handle
 
 **Mechanical operations:** If a task is fully scriptable with known inputs, write the script rather than running an interactive session. Candidate operations: stale PR remediation, branch cleanup, rebase-and-merge sequences. Use `~/.claude/scripts/merge-stale-pr.sh` for engineering-journal stale draft PRs.
 
-**Plan-then-optimize before acting:** Any task involving an Agent spawn, a skill invocation, or reads across more than one file requires this protocol. State a numbered plan first, then apply two explicit revision passes before taking any action.
+**Plan-then-optimize before acting:** Any task involving an Agent spawn, a skill invocation, reads across more than one file, or a switch to a new primary objective within the same session (e.g., moving from a `/review` or other skill output to addressing findings, or from one issue to another) requires this protocol. State a numbered plan first, then apply two explicit revision passes before taking any action.
 
 **Pass 1 — Token efficiency:** check:
 - Sequential tool calls that can be parallelized


brownm09:
## Review: fix: extend plan-then-optimize rule to mid-session task switches

### Summary

Ready to merge. This is a single-line change to a behavioral instruction file — minimal blast radius, correct intent. The new trigger case is well-motivated (the canonical `/review` → implementation pivot is exactly the gap being closed) and the example is clear. No blocking findings. One non-blocking observation about edge-case ambiguity in "a switch to a new task" is worth noting for a future refinement, but does not block this change.

---

### Non-Blocking Findings

**[maintainability]** `claude/CLAUDE.md:130`

The phrase "a switch to a new task within the same session" lacks a boundary definition. The example covers the canonical case (`/review` → implementing findings), but adjacent transitions — e.g., opening a PR and then writing a journal stub, or finishing a bug fix and then running a quick search — will be interpreted inconsistently across sessions. The rule could silently over-trigger (adding overhead to minor pivots) or under-trigger (a session interpreting "task switch" as only skill → implementation transitions).

**Suggestion:** Add a second minimal example or a clarifying qualifier, such as "or a switch to a new primary objective within the same session (e.g., moving from a `/review` or other skill output to addressing findings, or from one issue to another)". "Primary objective" sets a threshold that excludes routine sub-steps.

---

### Questions for Author

- The PR description notes the memory file was also updated, but the diff only shows `claude/CLAUDE.md`. The memory file lives at a machine-local path and is not version-controlled in this repo — is the CLAUDE.md text alone considered the authoritative source of truth, or is there a risk that a fresh environment (or another machine) runs with the old memory rule until the memory is manually rebuilt?

---

### Style Notes

The list restructuring from `"X, Y, or Z"` to `"X, Y, Z, or W"` is correct Oxford comma style. "Addressing findings" in the rule text vs. "implementing findings" in the PR body is a minor inconsistency, but "addressing" is the more general and accurate word — no change needed there.

PR #93, opened 2026-04-29
fix: replace stop-after-PR-creation with compact-then-review in same session

### LINKED ISSUE #92: Replace stop-after-PR-creation with compact-then-review in same session
The current rule stops the session immediately after PR creation, requiring a second session just to run `/review` and merge. This two-session overhead per PR is more friction than the token savings justify. Replace it with a `/compact` + same-session `/review` pattern that preserves most token savings without the session break.

brownm09:
Closes #92

## Summary
- Removes the hard session stop after PR creation
- Replaces it with: run `/compact` after PR, then run `/review <PR-URL> --post-comment` in the same session, address findings, and merge
- Updates the journal update-triggers: PR open is only a session boundary when merge can't happen in the same session (e.g., waiting on CI)
- Merge remains the primary session-end event; stub gets written at merge

## Motivation
The previous rule required two sessions per PR (implement+PR / review+merge). `/compact` reduces implementation context to a small summary before review fires, preserving most token savings without the session break.

## Testing
No automated tests — CLAUDE.md is a config document. Verified diff is correct and coherent with surrounding rules.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -68,7 +68,7 @@ If the project has no automated tests, the section must say so explicitly and de
 - **Never commit directly to `main`.** All changes go through a branch and PR, regardless of repo.
 - **Branch naming:** `feat/`, `fix/`, `config/`, `chore/`, `draft/` prefixes — match the convention already in use in the repo.
 - **PR first, then merge.** Open the PR immediately after pushing the branch; do not prompt the user to run `gh pr create` themselves.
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: fix: replace stop-after-PR-creation with compact-then-review in same session

### Summary
Ready to merge with one minor clarification noted. Both edits are coherent and self-consistent: the Git Workflow rule and the journal update-triggers bullet align correctly (stub on merge for same-session arc; stub on PR-open for CI-blocked arc). No correctness or reliability issues. One non-blocking wording ambiguity around who actually invokes `/compact`.

---

### Non-Blocking Findings

**[maintainability]** `claude/CLAUDE.md:71`
The new rule says "run `/compact`" as a directive to Claude, but `/compact` is a built-in CLI command that the user (or the interface) must invoke — Claude cannot call it as a tool. This session demonstrated exactly that gap: Claude told the user to run it rather than running it directly.
**Suggestion:** Rephrase to "prompt the user to run `/compact`, then immediately run `/review <PR-URL> --post-comment`" so the rule accurately reflects who performs the action and doesn't create confusion on future reads.

---

### Questions for Author
- If `/review` returns blocking findings that require non-trivial implementation work (not just a config tweak), does "address any blocking findings and merge in the same session" apply unconditionally? The rule is silent on whether a complex remediation warrants falling back to a new session. Intentional judgment-call territory, or worth adding an escape hatch?

---

PR #95, opened 2026-04-29
feat: emit worktree list on every prompt when multiple worktrees active

### LINKED ISSUE #94: Add UserPromptSubmit hook to alert on active worktrees
## Problem

When multiple Claude Code sessions are open simultaneously on the same repo (each in a separate git worktree), context switching between sessions creates confusion about which branch/worktree is currently active. This leads to edits and commits landing on the wrong branch.

## Solution

Add a `UserPromptSubmit` hook (`multi-worktree-alert.py`) that runs `git worktree list --porcelain` on every prompt submission. If two or more worktrees are active for the current repo, it emits a `systemMessage` listing all branches and marking the current one — providing ambient context at no manual cost.

If only one worktree is active, the hook exits silently (no noise in normal single-session work).

## Changes
- New script: `claude/scripts/multi-worktree-alert.py`
- Updated: `claude/settings.json` (add to `UserPromptSubmit` hooks)

brownm09:
## Summary
- Adds `claude/scripts/multi-worktree-alert.py` — a `UserPromptSubmit` hook that runs `git worktree list --porcelain` and emits the full list of active branches when ≥2 worktrees are open for the current repo
- Current worktree is marked with `*asterisks*` in the message for quick scanning
- Wires the script into `claude/settings.json` under `UserPromptSubmit`
- Exits silently when only one worktree is active (no noise in normal single-session work)

## Test plan
- [x] Smoke-tested against dev-env (39 active worktrees): emits correct alert with current branch marked
- [x] Smoke-tested against non-git directory: exits silently
- [x] Smoke-tested from subdirectory within a worktree: correctly resolves parent worktree path

Closes #94

### DIFF
--- a/claude/scripts/multi-worktree-alert.py
+++ b/claude/scripts/multi-worktree-alert.py
@@ -30,18 +30,18 @@ def parse_worktree_list(output: str) -> list[dict]:
       branch (str) — short branch name, or '<detached>' for detached-HEAD worktrees
     """
     worktrees: list[dict] = []
-    current: dict = {}
+    current: dict | None = None
     for line in output.splitlines():
         if line.startswith("worktree "):
-            if current:
+            if current is not None:
                 worktrees.append(current)
             current = {"path": line[len("worktree "):].strip(), "branch": ""}
-        elif line.startswith("branch "):
+        elif line.startswith("branch ") and current is not None:
             ref = line[len("branch "):].strip()
             current["branch"] = ref[len("refs/heads/"):] if ref.startswith("refs/heads/") else ref
-        elif line == "detached":
+        elif line == "detached" and current is not None:
             current["branch"] = "<detached>"
-    if current:
+    if current is not None:
         worktrees.append(current)
     return worktrees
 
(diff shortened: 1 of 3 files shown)


brownm09:
﻿## Review: feat: emit worktree list on every prompt when multiple worktrees active

### Summary
Ready to merge with one question resolved. The script is clean, handles all error paths silently, and the `find_current_worktree` path-matching logic is sound. The one thing worth clarifying before merge is a settings format discrepancy that could cause the hook to fail on Windows after the PR lands.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/multi-worktree-alert.py:57-63`
`find_current_worktree` returns `None` when no worktree path matches `cwd`. The caller never checks for `None` — instead, `wt is current` simply never matches, so the message emits without any branch starred. On Windows, if git returns worktree paths with different casing than `Path(cwd).resolve()`, every path comparison raises `ValueError` and the function returns `None` silently.
**Suggestion:** Add a fallback label (e.g., `"(current unknown)"`) when `current is None`, so the message is still informative and the silent case is visible.

**[maintainability]** `claude/scripts/multi-worktree-alert.py:27`
`parse_worktree_list` initializes `current: dict = {}` and relies on `if current:` to avoid appending the initial empty sentinel. The pattern works but requires the reader to trace why the first `worktree` line does not double-append.
**Suggestion:** Use `current: dict | None = None` and check `if current is not None` — makes the sentinel intent explicit.

---

### Questions for Author

- The PR adds the hook entry to `settings.json` using bare `python3 C:/...` (matching the existing repo format). However, the live `~/.claude/settings.json` currently wraps all commands in `bash -c '...'`. If those wrappers were added to fix the Windows Store `python3` stub problem, the new entry will hit the same issue once the PR is synced to the live config. Was that wrapper format intentional, and should the repo's `settings.json` be updated to match before this merges?

---

### Style Notes
The module docstring, function-level docstrings, and inline comments are well-calibrated — enough to explain non-obvious behavior (the path-sort rationale, the `<detached>` case) without narrating the obvious.

PR #97, opened 2026-04-29
docs: add Hook Safety section to CLAUDE.md

### LINKED ISSUE #96: Document hook safety invariants to prevent Bash blocking
## Problem

`pre-pr-create-check.py` blocked all Bash tool calls at some point during its development. The root cause was a timing window where `settings.json` referenced the script before it was committed to the branch the `~/.claude/scripts/` junction resolves against. Python exits with code 2 ("can't open file") when a PreToolUse hook script is missing; Claude Code treats any non-zero exit as a block signal — silently refusing the tool call with no error pointing to the hook.

This has happened more than once (PR #81 addressed a related `bash -c` wrapper failure mode).

## Solution

Add a `## Hook Safety` section to `claude/CLAUDE.md` documenting three invariants:

1. **Atomic commits** — script file and `settings.json` entry must land in the same commit, verified by `python3 <script> < /dev/null; echo "exit: $?"`
2. **Safe-exit guard** — advisory hooks exit 0 on all code paths (empty stdin, malformed JSON, unhandled exception)
3. **No `bash -c` wrappers** — hook commands call `python3 ...` directly, never via `bash -c '...'`

The `settings.json` and script are already deployed correctly. This issue tracks adding the CLAUDE.md documentation to prevent future regressions.

brownm09:
## Summary

- Adds a `## Hook Safety` section to `claude/CLAUDE.md` documenting three invariants that prevent hook scripts from silently blocking all Bash tool calls
- Root cause documented: `settings.json` referencing a script before it's committed to the junction-resolved path causes Python to exit 2, which Claude Code treats as a PreToolUse block

## Three invariants

1. **Atomic commits** — script file + `settings.json` entry in the same commit, verified with `python3 <script> < /dev/null; echo "exit: $?"`
2. **Safe-exit guard** — advisory hooks exit 0 on every code path (empty stdin, malformed JSON, exceptions)
3. **No `bash -c` wrappers** — `python3 ...` directly, never `bash -c 'python3 ...'` (root cause of dev-env#81)

## Test plan

- [ ] Read the new `## Hook Safety` section in `claude/CLAUDE.md` — confirm three invariants are present
- [ ] Verify `claude/scripts/pre-pr-create-check.py` exists and `python3 pre-pr-create-check.py < /dev/null` exits 0
- [ ] Open a new Claude Code session and confirm Bash tool is not blocked

Closes #96

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -106,6 +106,32 @@ If the project has no automated tests, the section must say so explicitly and de
 
 ---
 
+## Hook Safety
+
+PreToolUse hooks that exit non-zero **block the matched tool call silently** — the user sees the
+tool refused with no error pointing to the hook. This has caused complete Bash outages more than
+once. Three invariants prevent recurrence:
+
+1. **Atomic commits.** A `settings.json` hook entry and its script file must land in the **same
+   commit**. Never push a settings.json change that references a script not yet in
+   `claude/scripts/` on main. Verify before committing:
+   ```bash
+   python3 claude/scripts/<new-hook>.py < /dev/null; echo "exit: $?"
+   # Must print "exit: 0"
+   ```
+
+2. **Safe-exit guard.** Advisory hooks (hooks that emit a systemMessage reminder but do not
+   intend to block) must exit 0 on **every** code path — happy path, empty stdin, malformed
+   JSON, and unhandled exception. Wrap stdin reading and JSON parsing in bare `try/except` that
+   calls `sys.exit(0)`. Never add `sys.exit(N)` where N > 0 to an advisory hook.
+
+3. **No `bash -c` wrappers.** Hook commands call the interpreter directly:
+   `python3 C:/Users/brown/.claude/scripts/foo.py` — never `bash -c 'python3 ...'`. The
+   wrapper adds an exit-code-propagation layer and can fail independently on Windows (quoting
+   issues, PATH differences). This was the root cause of dev-env#81.
+
+---
+
 ## Model Selection
 
 Route tasks to the least powerful model that can handle them reliably:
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: docs: add Hook Safety section to CLAUDE.md

### Summary

Ready to merge. This is a clean, focused documentation PR — one file, 26 added lines, no deletions. The three invariants are accurate and the prose is clear. One non-blocking gap: the safe-exit guard invariant says hooks must handle "unhandled exceptions" but the prescription (wrapping only stdin reading and JSON parsing) doesn't fully close that path.

---

### Non-Blocking Findings

**[reliability]** `claude/CLAUDE.md` (Invariant 2 — safe-exit guard)

The invariant says advisory hooks must handle "unhandled exceptions" on every code path, but the prescribed mitigation — "wrap stdin reading and JSON parsing in bare `try/except`" — leaves code outside those two blocks (e.g., `print(json.dumps({...}))`, regex operations, `import` failures) able to raise and exit 1. The existing deployed scripts follow this partial pattern, so the docs accurately describe what's there, but they don't fully satisfy the invariant they claim to enforce.

**Suggestion:** Either extend the prescription to a top-level handler (`try: main() except Exception: sys.exit(0)`), or narrow the invariant's claim to "stdin, JSON parsing, and any path that could raise" and leave the top-level handler as the recommended pattern. The top-level handler is the cleaner fix and would make the deployed scripts fully match the documented invariant.

---

### Questions for Author

- The verification command (`python3 claude/scripts/<new-hook>.py < /dev/null`) confirms the empty-stdin path exits 0, but running from the repo root reads the script from the working tree, not the junction path (`~/.claude/scripts/`). Is it worth noting that the test also implicitly verifies the file exists in `claude/scripts/` before the settings.json entry is pushed — or is the junction-resolution distinction important enough to make explicit in the note?

---

### Style Notes

Cross-reference to `dev-env#81` in Invariant 3 is a good touch — it gives future readers a concrete incident to trace. The `<new-hook>` placeholder in the code example is unambiguous. No style concerns.

---

PR #99, opened 2026-04-29
chore: prune stale claude/* worktrees automatically

### LINKED ISSUE #98: Prune stale Claude session worktrees automatically
## Problem

Claude Code creates a new git worktree for every session. After the session's PR merges, the worktree is never removed. The repo currently has 42 active worktrees (41 Claude session worktrees), all with merged PRs. The \`multi-worktree-alert\` hook fires on every prompt and lists all of them, generating significant noise.

## Solution

1. A script \`claude/scripts/prune-merged-worktrees.py\` that removes \`claude/*\` worktrees whose branches are fully merged into \`origin/main\` and have no uncommitted changes.
2. A weekly scheduled routine \`claude/routines/prune-stale-worktrees/\` (Sunday midnight UTC) that runs the script automatically.
3. Run the script immediately to clear the current backlog of ~40 stale worktrees.

brownm09:
## Summary

- Adds `claude/scripts/prune-merged-worktrees.py`: removes `claude/*` worktrees whose branches are merged into `origin/main` and have no uncommitted changes. Handles both regular merges (git ancestry check) and squash merges (GitHub PR API fallback).
- Adds `claude/routines/prune-stale-worktrees/SKILL.md`: weekly scheduled routine (Sunday midnight UTC) that runs the script automatically.

## Test plan

- [x] Dry run (`--dry-run`) identified 32 correct pruning candidates before execution
- [x] Real run pruned 25 worktrees (42 → 5 active); 7 skipped on permission locks from other open sessions (non-fatal, will clear next week), 2 correctly preserved (`hopeful-williams-011160` has open PR #97, `epic-satoshi-103fa9` has no merged PR)
- [x] `git worktree list` confirms only 5 remain: main, eloquent-haslett (main branch ref), hopeful-williams, epic-satoshi, current session
- [x] `claude/optimistic-lewin-f436b2` (squash-merged PR #61, remote deleted) was correctly pruned via `gh pr list` fallback

Closes #98

### DIFF
--- a/claude/routines/prune-stale-worktrees/SKILL.md
+++ b/claude/routines/prune-stale-worktrees/SKILL.md
@@ -1,0 +1,23 @@
+---
+name: prune-stale-worktrees
+description: Remove Claude session worktrees whose branches have been merged into main.
+schedule: "0 0 * * 0"
+---
+
+Prune stale Claude session worktrees in the dev-env repo. Run fully autonomously — do not ask the user anything.
+
+**Objective:** Remove all `claude/*` worktrees whose branches are fully merged into `origin/main` and have no uncommitted changes. Report the pruned/skipped summary.
+
+**Steps:**
+1. Run the prune script from the dev-env repo root:
+   ```bash
+   python C:/Users/brown/Git/dev-env/claude/scripts/prune-merged-worktrees.py
+   ```
+2. Report the output: how many worktrees were pruned, how many skipped, and the reason for each skip.
+3. If `git worktree list` in dev-env shows any `claude/*` branches that the script skipped due to "not merged" status, list them for manual review.
+
+**Constraints:**
+- Dev-env repo: `C:/Users/brown/Git/dev-env`
+- Script uses `git branch -d` (not `-D`) and `git worktree remove` (no `--force`) — safe by default
+- Never remove the current session's worktree or any non-`claude/*` branch worktree
+- Temp files (if needed) go to `C:/Users/brown/.claude/scratch/`
--- a/claude/routines/prune-stale-worktrees/SKILL.md
+++ b/claude/routines/prune-stale-worktrees/SKILL.md
@@ -14,7 +14,7 @@ Prune stale Claude session worktrees in the dev-env repo. Run fully autonomously
    python C:/Users/brown/Git/dev-env/claude/scripts/prune-merged-worktrees.py
    ```
 2. Report the output: how many worktrees were pruned, how many skipped, and the reason for each skip.
-3. If `git worktree list` in dev-env shows any `claude/*` branches that the script skipped due to "not merged" status, list them for manual review.
+3. If `git worktree list` in dev-env shows any `claude/*` branches that the script skipped due to "not merged" status, list them and send a push notification summarizing the count and branch names so the user can investigate.
 
 **Constraints:**
 - Dev-env repo: `C:/Users/brown/Git/dev-env`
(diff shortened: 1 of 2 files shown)


brownm09:
## Review: chore: prune stale claude/* worktrees automatically

### Summary

Ready to merge with minor changes. The script is well-designed and safe-by-default: it uses `git branch -d` / `git worktree remove` (no force flags), skips dirty worktrees, and has a correct squash-merge fallback via the GitHub API. The most noteworthy finding is that `current_worktree_path()` always resolves against the hardcoded `REPO` constant rather than the actual process CWD, so the "skip current worktree" guard protects the main checkout but not a `claude/*` worktree that might be running the script. This is safe in practice (an active session's branch isn't merged yet, and `git worktree remove` on an active tree fails) but is technically wrong. No correctness or security blockers.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/prune-merged-worktrees.py:59-61`
`current_worktree_path()` calls `git rev-parse --show-toplevel` with `cwd=REPO` (the hardcoded main-repo path), so it always returns the main repo directory regardless of where the script is invoked from. The `path == cwd` guard therefore only skips the main checkout — any `claude/*` worktree running the script is not protected by this check.
In practice this is benign: the running session's branch is never merged (it was just created), and `git worktree remove` on an active tree fails. But the branch-delete step that follows (`git branch -d`) can still execute on any branch that passes `is_merged()`, even if that branch is checked out in another worktree.
**Suggestion:** Derive `cwd` from `os.getcwd()` or pass `cwd=None` to the inner `run()` call so it picks up the actual process working directory, then resolve against that.

**[reliability]** `claude/scripts/prune-merged-worktrees.py:54-56`
`is_dirty()` calls `git status --porcelain` with `check=False`. If `path` no longer exists (e.g., a worktree whose directory was manually deleted but whose git metadata is still registered), the subprocess fails silently and `r.stdout.strip()` returns `""` — treated as "not dirty." `git worktree remove` is then called on the missing path and fails non-fatally, but the branch delete still fires for merged branches.
**Suggestion:** Add an early `Path(path).exists()` check in `is_dirty()` and return `True` (treat as dirty/unsafe) if the path is inaccessible, or add it before the `is_dirty` call in `main()`.

**[maintainability]** `claude/scripts/prune-merged-worktrees.py:15`
`REPO = "C:/Users/brown/Git/dev-env"` is a hardcoded absolute path that ties the script to one machine and one directory layout. The script also uses `REPO` as the default `cwd` in `run()`, so every git command runs against this constant rather than the actual repo root.
**Suggestion:** Derive `REPO` dynamically: `REPO = str(Path(__file__).resolve().parents[2])` (two levels up from `claude/scripts/`) — this works wherever the repo is cloned and doesn't require changes to the rest of the script.

---

### Questions for Author

- The SKILL.md routine instructs Claude to "list them for manual review" for unmerged `claude/*` branches (step 3), but this is a fully autonomous scheduled task. Who sees that output? If there's no notification path, that step may be silently lost. Worth clarifying whether a `PushNotification` or logged output is expected here, or whether the weekly summary in the session transcript is sufficient.

---

### Style Notes

In `main()`, the skipped-message format string for the prefix check renders as `"branch 'X' not in claude/* prefix"` — the `*` is appended manually to `BRANCH_PREFIX` in the f-string. This is a minor inconsistency (the constant itself doesn't contain the glob wildcard), but it's readable and harmless.

---

PR #103, opened 2026-04-29
feat: emit journal-stub reminder after gh pr create

### LINKED ISSUE #102: Hook: emit journal-stub reminder after gh pr create
## Problem

The CLAUDE.md policy requires a journal stub to be written immediately after a PR is opened, before any merge. This isn't being enforced reliably — and when a session is corrupted during \`/compact\`, any session context captured only in Claude's memory is lost permanently.

## Solution

Extend \`claude/scripts/pr-merge-reminder.py\` to also detect top-level \`gh pr create\` invocations and emit a reminder to write the stub immediately. The script is already wired into \`PostToolUse\` hooks, so no settings.json changes are needed.

brownm09:
## Summary

- Extends `pr-merge-reminder.py` to also detect top-level `gh pr create` commands in PostToolUse hooks
- Emits a distinct reminder to write the journal stub immediately — before compaction or session corruption can erase context
- Refactors the shared stack-based shell parser into `_scan_top_level()` so both merge and create detectors reuse the same heredoc/quote/subshell-aware logic

## Test plan

- [x] Empty stdin → exit 0 (safe-exit guard)
- [x] `gh pr create --title "foo"` → exit 2, stub reminder emitted
- [x] `gh pr merge 5 --merge` → exit 2, merge reminder emitted (unchanged)
- [x] `echo "gh pr create"` (inside string) → exit 0 (no false positive)

Closes #102

### DIFF
--- a/claude/scripts/pr-merge-reminder.py
+++ b/claude/scripts/pr-merge-reminder.py
@@ -22,6 +22,7 @@ Exit 2  — gh pr create and/or gh pr merge detected; reminder(s) emitted via st
 import json
 import re
 import sys
+from collections.abc import Callable
 
 # Matches the start of a statement token against `gh pr merge` or `gh pr create`.
 _MERGE_RE = re.compile(r"(?:cd\s+\S+\s+&&\s+)?gh\s+pr\s+merge\b")
@@ -82,7 +83,7 @@ def _find_heredoc_end(cmd: str, start: int) -> int:
     return i
 
 
-def _scan_top_level(command: str, check_fn) -> bool:
+def _scan_top_level(command: str, check_fn: Callable[[str], bool]) -> bool:
     """Return True when *command* contains a top-level statement matched by
     *check_fn* — i.e. not inside a quoted string, $() subshell, or heredoc body.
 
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: feat: emit journal-stub reminder after gh pr create

### Summary

Ready to merge. The refactor is clean and mechanically sound — the stack-based parser body is extracted into `_scan_top_level` without modification, and the new `is_pr_create_command` mirrors the merge path exactly. One non-blocking gap worth noting: neither path guards against a failed `gh pr create` exit code, which could cause the hook to remind Claude to write a stub when no PR was actually opened. No blocking findings.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/pr-merge-reminder.py:192–202`
`main()` evaluates `is_pr_create_command(command)` against the raw command string without checking `tool_response.exitCode`. If `gh pr create` exits non-zero (auth failure, no remote, branch protection error), the hook still emits the stub reminder — instructing Claude to document a PR that was never created. The merge path has the same gap (pre-existing), but the consequence is more concrete for creates: a spurious stub would reference a non-existent PR number.
**Suggestion:** Add `exit_code = data.get("tool_response", {}).get("exitCode", 0)` and `if exit_code != 0: sys.exit(0)` after the `tool_name` guard — mirrors the pattern already in `post-tool-use.py`. Consider doing the same for the merge path in the same commit.

---

### Style Notes

`_scan_top_level`'s `check_fn` parameter has no type annotation while the rest of the file uses standard type hints (e.g., `cmd: str`, `start: int`). Adding `check_fn: Callable[[str], bool]` (with `from collections.abc import Callable`) would keep it consistent — minor, entirely optional.

---

Issue #31, opened 2026-04-19
new-day-journal-check.py: skip fire if composed journal already exists for draft date


brownm09:
## Problem

The stale-draft hook (`new-day-journal-check.py`) fires when a `YYYY-MM-DD_draft.md` file exists on disk. It does not check whether a composed journal (`YYYY-MM-DD-<slug>.md`) already exists for the same date and project. This causes two false positive patterns:

**Pattern 1 — Orphaned untracked draft file:** The journal was already composed and committed, but the draft file was left on disk as an untracked artifact (e.g., via `git rm` removing it from the index without deleting the physical file, or the file being recreated after the rm commit). The hook fires and blocks work; ~$2 spent to determine the journal was already done.

**Pattern 2 — Just-created meta draft:** A meta draft is created by the current session (e.g., during journal-compose for another project). The next prompt in the same session triggers the hook against the newly-created file as if it were stale.

Observed: 2026-04-19 morning session. Hook fired twice consecutively on `sessions/meta/2026-04-18_draft.md` — once after it was created, once on the following prompt.

## Proposed fix

In `new-day-journal-check.py`, before emitting any warning, check two conditions:

**1. Composed journal exists:**
```python
import glob

def composed_journal_exists(sessions_dir, project, date_str):
    pattern = os.path.join(sessions_dir, project, f"{date_str}-*.md")
    return len(glob.glob(pattern)) > 0
```
If a matching composed journal exists, delete the orphaned draft silently and exit 0.

**2. Draft file is too new (created within current session window):**
```python
import time

draft_mtime = os.path.getmtime(draft_path)
age_seconds = time.time() - draft_mtime
if age_seconds < 3600:  # less than 1 hour old
    exit(0)  # not stale — just created
```

## Files to change

- `claude/scripts/new-day-journal-check.py` — add both checks before emitting warning

## Related

Pairs with #30 (once-per-session suppression). The suppression in #30 prevents repeat fires within a session; this issue prevents the false-positive fire entirely.

PR #111, opened 2026-04-30
feat: once-per-session journal hook + Stop hook for session-close reminder

### LINKED ISSUE #30: new-day-journal-check.py: fire once per session + Stop hook for session-close reminder
## Problem

`new-day-journal-check.py` (UserPromptSubmit hook) fires on every prompt in a session once a stale draft branch is detected. This injects repeated context into the conversation, wasting tokens across multiple sessions on the following day.

## Proposed solution

Two-part change:

### 1. Once-per-session suppression in `new-day-journal-check.py`

- Read `session_id` from the stdin JSON payload
- Write a flag file to `C:/Users/brown/.claude/scratch/journal_hook_<session_id>.flag` on first fire
- Skip output on subsequent prompts in the same session
- Clean up flag files older than 24 hours at startup (self-pruning; no session lasts a day)
- Flag is written **only when messages are emitted** — clean sessions stay cheap

### 2. New `Stop` hook: `journal-stop-check.py`

- Fires when a session ends (Stop event)
- Re-checks for stale draft files and unmerged draft branches (same logic as UserPromptSubmit hook)
- If stale condition exists, prints a closing reminder before the session closes
- Register in `settings.json` under the `Stop` hook array alongside `token-tracker.py`

### Result

| Trigger | Fires | Token cost |
|---|---|---|
| Session start (UserPromptSubmit) | Once per session | ~1 reminder |
| Session end (Stop) | Once per session | ~1 reminder |
| Mid-session prompts | Never | Zero |

Two reminders per session (start + end), zero mid-session noise, guaranteed visibility at both natural checkpoints.

## Files to change

- `claude/scripts/new-day-journal-check.py` — add session-scoped suppression + flag cleanup
- `claude/scripts/journal-stop-check.py` — new script
- `claude/settings.json` — add `journal-stop-check.py` to `Stop` hooks array

brownm09:
## Summary

- `new-day-journal-check.py` now fires at most once per session by writing a `~/.claude/scratch/journal_hook_<session_id>.flag` on first emission; flag files older than 24 h are pruned at startup
- New `journal-stop-check.py` Stop hook re-runs the same stale-draft and unmerged-branch checks at session end, plus removes orphaned draft files left as untracked after `git rm` (prevents false positives per #31)
- `settings.json` updated to register the new Stop hook alongside `token-tracker.py`

## Test plan

- [ ] `python3 claude/scripts/new-day-journal-check.py < /dev/null; echo "exit: $?"` → exit: 0
- [ ] `python3 claude/scripts/journal-stop-check.py < /dev/null; echo "exit: $?"` → exit: 0
- [ ] Start a fresh session with a stale draft present: first prompt shows `[journal-hook]` message; second prompt in same session is silent
- [ ] End a session with a stale draft present: closing `[journal-stop-hook]` message appears
- [ ] End a clean session: no output from either hook

Closes #30

### DIFF
--- a/claude/scripts/journal-stop-check.py
+++ b/claude/scripts/journal-stop-check.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+"""
+Stop hook: remind user of stale journal work at session end.
+
+Two checks (same logic as new-day-journal-check.py):
+1. Stale *_draft.md / *.stub.md files from before today
+2. Unmerged remote draft/* branches
+
+Also cleans up orphaned draft files: physical files left on disk as untracked
+after git rm. This prevents new-day-journal-check.py false positives on the
+next session (see dev-env#31).
+
+Exit 0 always — never block session close.
+Stdout is shown to user as a closing message.
+"""
+
+import glob
+import json
+import os
+import subprocess
+import sys
+from datetime import date
+from pathlib import Path
+
+JOURNAL_REPO = Path.home() / "Git" / "engineering-journal"
+TODAY = date.today().strftime("%Y-%m-%d")
+
+
+def stale_draft_artifacts() -> list[str]:
+    """Return paths of *_draft.md or *.stub.md files from before today."""
+    artifacts = []
+    for pattern in (
+        str(JOURNAL_REPO / "sessions" / "**" / "*_draft.md"),
+        str(JOURNAL_REPO / "sessions" / "**" / "????-??-??_*.stub.md"),
+    ):
+        artifacts.extend(glob.glob(pattern, recursive=True))
+    stale = [f for f in artifacts if not os.path.basename(f).startswith(TODAY)]
+    stale.sort(key=lambda f: os.path.basename(f), reverse=True)
+    return stale
+
+
+def composed_dates_on_main() -> set[str]:
+    """Return YYYY-MM-DD dates that have a composed journal file on origin/main."""
+    try:
+        result = subprocess.run(
+            ["git", "ls-tree", "-r", "--name-only", "origin/main", "sessions/"],
+            cwd=JOURNAL_REPO,
+            capture_output=True,
+            text=True,
+            timeout=10,
+        )
+        dates = set()
+        for line in result.stdout.splitlines():
+            fname = line.split("/")[-1]
+            if fname.endswith(".stub.md") or not fname.endswith(".md"):
+                continue
+            if len(fname) >= 10 and fname[4] == "-" and fname[7] == "-":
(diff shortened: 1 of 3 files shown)


brownm09:
## Review: feat: once-per-session journal hook + Stop hook for session-close reminder

### Summary

Merge with minor revision required. The implementation is correct and well-structured — session-scoped flag suppression works, orphan cleanup is precise, and the Stop hook is properly guarded. One blocking issue: `new-day-journal-check.py` is missing the safe-exit guard that the hook safety rules explicitly require, and the new `json.loads(raw)` call added in this PR is the uncovered code path most likely to throw.

---

### Blocking Findings

**[reliability]** `claude/scripts/new-day-journal-check.py:122–131`

The new `json.loads(raw)` call is not wrapped in try/except, and the file's `__main__` block (`if __name__ == "__main__": main()`) has no exception handler. If stdin contains malformed JSON — possible during testing, version mismatch, or hook chain corruption — `JSONDecodeError` propagates uncaught and the process exits non-zero. `journal-stop-check.py` (the new file in this same PR) correctly has the guard; `new-day-journal-check.py` does not.

CLAUDE.md hook safety rule is explicit: "Use a top-level exception handler so no code path escapes."

**Fix:** Change the `__main__` block in `new-day-journal-check.py` to:
```python
if __name__ == "__main__":
    try:
        main()
    except Exception:
        sys.exit(0)
```

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/journal-stop-check.py:119`

`json.loads(raw) if raw else {}` evaluates and discards the result — no assignment, no side effect. The comment acknowledges the data is unused, but an expression-with-no-assignment reads as a logic error (forgotten variable) rather than a deliberate choice.

**Suggestion:** Replace with `_ = json.loads(raw) if raw else {}` or drop the expression entirely and leave only the comment, since the comment already documents the intent.

---

### Questions for Author

- When `session_id` is empty (e.g., `python3 new-day-journal-check.py < /dev/null` in CI/test), the flag logic is bypassed and the hook fires on every prompt for that session. Is silent degradation to the old behavior acceptable, or should there be a fallback (e.g., writing a time-based flag)?

---

### Style Notes

`journal-stop-check.py` contains verbatim copies of `stale_draft_artifacts`, `composed_dates_on_main`, and `unmerged_draft_branches` from `new-day-journal-check.py`. The plan chose this intentionally (no shared module). A one-line marker such as `# keep in sync with new-day-journal-check.py` at the top of each duplicated function would make the copy relationship explicit and prevent silent drift when one is updated without the other.

---

PR #124, opened 2026-04-30
docs: add ExitWorktree-before-merge rule to prevent branch-delete failure

### LINKED ISSUE #123: Document ExitWorktree-before-merge to avoid branch-delete failure
## Problem

When merging a dev-env PR from within the worktree created for that branch, `gh pr merge --squash --delete-branch` reliably fails on the local branch-delete step (step 4 of gh's sequence). Steps 1–3 — squash merge, remote branch delete, prune tracking ref — all succeed, but `git branch -d` errors with "cannot delete branch checked out at …" and gh exits non-zero, making it appear the entire merge failed.

## Fix

Add a rule to the Git Workflow section of CLAUDE.md: call `ExitWorktree` before `gh pr merge`. With the branch no longer checked out in any worktree, all four steps complete cleanly.

brownm09:
## Summary

- Adds one bullet to the Git Workflow section of `CLAUDE.md` documenting that `ExitWorktree` must be called before `gh pr merge` when the PR was developed in a worktree
- Root cause of the recurring error: `gh pr merge --squash --delete-branch` succeeds on steps 1–3 (GitHub merge, remote branch delete, prune tracking ref) but fails on step 4 (`git branch -d`) because git refuses to delete a branch checked out in an active worktree
- `prune-merged-worktrees.py` already handles cleanup correctly as a weekly safety net; this rule prevents the error in the first place

## Test plan

- [ ] Merge a dev-env worktree PR after calling `ExitWorktree` — confirm clean exit with no branch-delete error
- [ ] `python3 -m py_compile claude/scripts/*.py` passes (no script changes, baseline hygiene)

Closes #123


brownm09:
## Review: docs: add ExitWorktree-before-merge rule to prevent branch-delete failure

### Summary
Ready to merge with one clarification addressed. The rule correctly identifies the problem and its fix, but the key claim — that `ExitWorktree` "will find the branch unchecked-out" — is only true if `ExitWorktree` removes the worktree *registration* (not just changes the working directory). Git's branch-delete guard is keyed to registered worktrees, not the active process's cwd. If this is confirmed, the rule is accurate as written; if not, the fix instruction needs to change. One non-blocking gap: no recovery guidance for sessions already stuck in the error state.

---

### Blocking Findings

*(none)*

---

### Non-Blocking Findings

**[reliability]** `claude/CLAUDE.md:85`
The rule documents the prevention path but not the recovery path. A reader who hits this error mid-session won't immediately know whether the merge succeeded — `gh` exited non-zero, so it looks like a full failure. Without guidance they may re-run `gh pr merge`, get "PR is already merged," and be confused.
**Suggestion:** Append one sentence after the main rule: "If you see this error, the merge and remote branch delete have already completed — only the local branch remains and will be collected by the weekly `prune-merged-worktrees.py` run."

---

### Questions for Author

- Does `ExitWorktree` remove the worktree directory and its git registration (so `git branch -d` sees the branch as unchecked-out), or does it only change the active working directory? The rule's correctness depends entirely on the former. If `ExitWorktree` is purely a cwd switch, the fix needs to read `git worktree remove <path>` instead (or in addition).

---

### Style Notes

The new bullet is accurate and the internal explanation ("steps 1–3 of `gh`'s sequence") is a nice touch — it tells the reader exactly how far `gh` got. Consistent with the density of the surrounding bullets.

---

Issue #133, opened 2026-04-30
feat: PostCompact hook auto-invokes /review when open PRs exist


brownm09:
## Problem

After running `/compact`, the PR URL is lost from context and the user must manually type `/review <PR-URL> --post-comment`. This is unnecessary friction — the workflow already mandates this step.

## Solution

Extend `post-compact.py` to read `open-prs.jsonl` from the engineering-journal repo and emit a `systemMessage` with the open PR URL(s) and an imperative instruction to invoke `/review`. Claude sees this in its post-compaction response turn and executes automatically.

Only fires on manual `/compact` (`trigger == "manual"`); auto-compact is unchanged.

PR #138, opened 2026-05-01
fix: emit dev-env-sync divergence warning to stderr

### LINKED ISSUE #137: fix: dev-env-sync.py divergence warning uses stdout (invisible as system message)
## Problem

When the local `dev-env` main branch has diverged from `origin/main`, `dev-env-sync.py` emits a warning via plain `print()` to stdout. UserPromptSubmit hook stdout is injected as a Claude system message — not shown in the UI — so the user never sees it.

This is what caused the PostCompact /review regression: the sync script detected the diverged state and silently warned, but the warning was invisible, so the stale hook kept running unnoticed.

## Fix

Change the two warning `print()` calls (diverged state, failed pull) to `print(..., file=sys.stderr)` so they surface in the terminal output visible to the user.

brownm09:
## Summary

- Changes two `print()` calls in `dev-env-sync.py` to `print(..., file=sys.stderr)` so diverged-main and failed-pull warnings surface in the terminal instead of being silently swallowed as system messages.

## Context

UserPromptSubmit hook **stdout** is injected into Claude's context as a system message — not shown in the UI. The divergence warning was therefore invisible, which allowed stale symlinked tooling to keep running undetected. This is the root cause of the PostCompact `/review` regression from issue #133: local `main` diverged, the warning fired but was never seen, and the old `post-compact.py` (without the `systemMessage` output) kept running.

## What's not changed

The success-pull notification stays on stdout — it's informational and the system-message channel is fine for it.

## Test plan

- [ ] `python3 -m py_compile claude/scripts/dev-env-sync.py` → syntax OK
- [ ] `python3 claude/scripts/dev-env-sync.py < /dev/null; echo "exit: $?"` → exits 0

Closes #137

### DIFF
--- a/claude/scripts/dev-env-sync.py
+++ b/claude/scripts/dev-env-sync.py
@@ -77,7 +77,8 @@ def main() -> None:
         print(
             "[dev-env-sync] WARNING: local dev-env repo has diverged from origin/main.\n"
             "CLAUDE.md and symlinked tooling may be stale. Run `git -C ~/Git/dev-env "
-            "status` to investigate before proceeding."
+            "status` to investigate before proceeding.",
+            file=sys.stderr,
         )
         sys.exit(0)
 
@@ -99,7 +100,8 @@ def main() -> None:
     else:
         print(
             "[dev-env-sync] WARNING: fast-forward pull failed.\n"
-            + pull.stderr.strip()
+            + pull.stderr.strip(),
+            file=sys.stderr,
         )
 
     sys.exit(0)


brownm09:
## Review: fix: emit dev-env-sync divergence warning to stderr

### Summary

Ready to merge. The change is minimal, targeted, and correct: both warning-class `print()` calls now route to `stderr` so they surface in the terminal rather than being silently swallowed as invisible system messages. The success-pull notification intentionally stays on `stdout`. No correctness, reliability, or security concerns.

---

### Blocking Findings

None.

---

### Non-Blocking Findings

None.

---

### Questions for Author

None.

---

### Style Notes

The module docstring (lines 6–13) says "Silent on success; emits a warning if the repo has diverged" — still accurate, but a future reader might benefit from knowing warnings go to stderr specifically. Not worth a separate commit; consider a one-word tweak ("emits a **stderr** warning") if you're in this file for another reason.

---

PR #143, opened 2026-05-01
docs: add cross-session worktree removal fallback to CLAUDE.md

### LINKED ISSUE #141: docs: add cross-session worktree removal fallback to CLAUDE.md
## Problem

The `Merging a PR developed in a worktree` rule tells Claude to call `ExitWorktree` with `action: "remove"` before `gh pr merge`. When merging happens in a *new* session (not the one that created the worktree), the tool returns a no-op:

> No-op: there is no active EnterWorktree session to exit.

The rule already documents the `cannot delete branch checked out at` error case, but is silent on the no-op case. This leaves Claude without guidance — the worktree stays registered and `git branch -d` fails when `gh pr merge` tries to clean up the local branch.

## Fix

Extend the rule with the manual fallback:
```
git -C C:/Users/brown/Git/dev-env worktree remove .claude/worktrees/<name>
git -C C:/Users/brown/Git/dev-env branch -d <branch>
```

Triggered by the exact failure on branch `claude/quizzical-mclaren-da8507`.

brownm09:
## Summary
- `ExitWorktree` with `action: "remove"` is session-scoped — calling it in a new session returns a no-op
- The existing rule handled the `cannot delete branch checked out at` error but not this no-op case
- Adds the manual fallback: `git worktree remove` + `git branch -d` from the main repo before merging

## Test plan
- [x] `grep -n 'date -u' claude/CLAUDE.md` — only match is the testing instruction itself (no stub/branch name context)
- [x] Docs-only change — no scripts modified

Closes #141


brownm09:
## Review: docs: add cross-session worktree removal fallback to CLAUDE.md

### Summary
Nearly ready to merge — one blocking correctness issue with the documented git command. The placement of the new sentence is correct (between the `action: "keep"` explanation and the "cannot delete" error case), and the intent is accurate. Fix the `-d` → `-D` flag before merging.

---

### Blocking Findings

**[correctness]** `claude/CLAUDE.md:85`
The fallback command uses `git branch -d <branch>`, which is the safe-delete form — it refuses to delete a branch that isn't fully merged into its upstream or HEAD. This command runs *before* `gh pr merge`, so the local branch is unmerged at that point. `git branch -d` will exit with `error: The branch '...' is not fully merged` and the cleanup won't complete.
**Fix:** Change `git branch -d <branch>` to `git branch -D <branch>`. The merge happens on GitHub via squash (a commit that never exists locally), so the safe-delete check is irrelevant here. `gh pr merge --squash --delete-branch` will still find the remote branch exists and handle the remote deletion correctly even after the local branch is already gone.

---

PR #144, opened 2026-05-01
fix: suppress hook false positive when draft branch cut from previous day's draft

### LINKED ISSUE #142: Hook false positive when draft branch cut from previous day's draft instead of main
## Problem

`new-day-journal-check.py` and `journal-stop-check.py` detect stale stub files purely by filesystem presence — they do not cross-reference whether the stub's `(project, date)` already has a composed journal file on `origin/main`.

When a new draft branch (`draft/YYYY+1-MM-DD`) is accidentally cut from the previous day's draft branch instead of from `main`, all pre-compose stubs ride along. After the compose squash-merges to main, those stubs remain on the new draft branch. Every subsequent session then sees the hook fire on already-composed content.

## Root cause (2026-05-01 incident)

`draft/2026-05-01` was cut from `draft/2026-04-30` at commit `8218238` (before the compose ran). The compose later deleted all 04-30 stubs and produced composed files (squash-merged as PR #34). `draft/2026-05-01` inherited the pre-deletion state; 29 stale stub/manifest files carried forward across all four projects.

## Fix

1. **Hook scripts**: add `composed_project_dates_on_main()` returning `set[tuple[project, date]]`; filter `stale_draft_artifacts()` to skip stubs whose `(project, date)` is already composed on main.
2. **CLAUDE.md**: add branch validation step after `git checkout -b draft/YYYY-MM-DD` — run `merge-base HEAD origin/main` vs `rev-parse origin/main` and require they match.

## Impact

Fires on every session when lineage is wrong — which can persist for days.

brownm09:
## Summary

- Adds `composed_project_dates_on_main()` to both journal hook scripts, returning `(project, YYYY-MM-DD)` pairs with a composed file on `origin/main`
- Updates `stale_draft_artifacts()` to skip stubs whose `(project, date)` is already composed — these are branch-lineage artifacts carried forward when a new-day branch is cut from the previous day's draft instead of from main
- Adds a branch validation step to CLAUDE.md first-session instructions: after `git checkout -b draft/YYYY-MM-DD`, require `merge-base HEAD origin/main` == `rev-parse origin/main`

## Root cause

`draft/2026-05-01` was cut from `draft/2026-04-30` at a pre-compose commit. The compose later deleted all 04-30 stubs and produced composed files (squash-merged as PR #34). `draft/2026-05-01` inherited the pre-deletion state; 29 stale stub/manifest files across four projects triggered the hook every session.

## Test plan

- [x] `python3 -m py_compile claude/scripts/new-day-journal-check.py claude/scripts/journal-stop-check.py` — syntax OK
- [x] `python3 claude/scripts/new-day-journal-check.py < /dev/null` — exits 0, no output
- [x] Hook fires correctly on the current session (no false positives after 04-30 artifact removal in engineering-journal)

Closes #142

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -320,6 +320,14 @@ operational artifacts (compose lock files, log file timestamps).
 **First session of the day:**
 1. `git -C C:/Users/brown/Git/engineering-journal checkout main && git pull`
 2. `git -C C:/Users/brown/Git/engineering-journal checkout -b draft/YYYY-MM-DD`
+   **Branch validation (required):** Confirm the new branch was cut from `main`, not a previous
+   draft branch. Both commands below must print the same SHA — if they differ, delete the local
+   branch (`git -C C:/Users/brown/Git/engineering-journal branch -D draft/YYYY-MM-DD`) and
+   re-run steps 1–2:
+   ```bash
+   git -C C:/Users/brown/Git/engineering-journal merge-base HEAD origin/main
+   git -C C:/Users/brown/Git/engineering-journal rev-parse origin/main
+   ```
 3. Read `sessions/<project>/open-prs.jsonl` if it exists — include its PR list as session context before starting work.
 4. Create `sessions/<project>/YYYY-MM-DD_HHMMSS.stub.md` (see stub structure below)
 5. Add a `<!-- tokens: input=N output=N cost≈$N -->` comment at the end of the session block
--- a/claude/scripts/journal-stop-check.py
+++ b/claude/scripts/journal-stop-check.py
@@ -26,15 +26,65 @@ JOURNAL_REPO = Path.home() / "Git" / "engineering-journal"
 TODAY = date.today().strftime("%Y-%m-%d")
 
 
+def composed_project_dates_on_main() -> set[tuple[str, str]]:
+    """Return (project, YYYY-MM-DD) pairs that have a composed file on origin/main.
+
+    Finer-grained than composed_dates_on_main(): used to suppress false positives
+    when stubs from a composed date are still present on disk because a new-day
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: fix: suppress hook false positive when draft branch cut from previous day's draft

### Summary

Ready to merge. The fix is correct and well-scoped. `composed_project_dates_on_main()` accurately cross-references stub files against composed output on `origin/main` using the same `git ls-tree` pattern already validated in `composed_dates_on_main()`. The CLAUDE.md branch validation step closes the human-procedure gap that caused the incident. One non-blocking note below.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/new-day-journal-check.py:40-57` and `journal-stop-check.py:29-57`

`composed_project_dates_on_main()` is duplicated verbatim across both scripts. The existing `composed_dates_on_main()` in `new-day-journal-check.py` carries a docstring note: *"reads the local remote-tracking ref (origin/main) without fetching — a stale ref could produce a false positive, but the next fetch will self-correct."* The new function has the same limitation but omits that note. A reader who encounters `composed_project_dates_on_main()` first (e.g., in `journal-stop-check.py`, where it appears before `composed_dates_on_main()`) won't know about the stale-ref behavior.

**Suggestion:** Add the stale-ref caveat to `composed_project_dates_on_main()`'s docstring in both files. Since the scripts are standalone executables with no shared module, duplication itself is acceptable — but the missing note creates a documentation asymmetry.

---

### Questions for Author

- The `project` extraction logic falls back to `project = None` on `StopIteration` or `IndexError`, and a `None` project means the stub is *not* suppressed (conservative behavior). Is there a known path shape that would hit this branch? If so, a log line might help future debugging.

---

PR #146, opened 2026-05-01
fix: replace exit-2 PostToolUse hook with sentinel file approach

### LINKED ISSUE #145: fix: stub-push-archive-reminder exits non-zero without output on Windows stdout buffer flush failure
## Problem

`claude/scripts/stub-push-archive-reminder.py` (PostToolUse:Bash hook) causes a blocking error notification after git push commands. The error reads:

```
[python3 .../stub-push-archive-reminder.py]: No stderr output
```

## Root cause

On Windows + Git Bash, Python's stdout is block-buffered when connected to a pipe (the hook runner). When the trigger conditions are met, the script calls:

```python
print(json.dumps({"systemMessage": msg}))
sys.exit(2)
```

If the flush that Python performs during `sys.exit()` shutdown fails (e.g., BrokenPipeError because the hook runner stopped reading early), the JSON output never reaches the hook runner. The runner then sees a non-zero exit with no output and shows a blocking error instead of the intended systemMessage.

## Fix

Add an explicit `sys.stdout.flush()` before `sys.exit(2)`, wrapped in a try/except so any I/O failure falls back to `sys.exit(0)` rather than exiting with an unexpected non-zero code — consistent with the hook safety invariant that advisory hooks must not block on error paths.

## Hook safety invariant

Per `claude/CLAUDE.md` → Hook Safety: advisory hooks must exit 0 on every non-trigger code path and on any exception.

brownm09:
## Summary

- **Root cause:** PostToolUse hooks do not support the exit-2 `systemMessage` mechanism — any non-zero exit is treated as a blocking error. The original hook called `sys.exit(2)`, causing a spurious "blocking error" notification after every stub push. The same limitation applies to Stop hooks in this Claude Code environment.
- **Fix:** Two-part sentinel approach where all hooks exit 0 always:
  - `stub-push-archive-reminder.py` (PostToolUse) writes `~/.claude/scratch/stub-pushed.flag` when it detects a stub push, then exits 0.
  - `journal-stop-check.py` (Stop) checks for the sentinel at session end, consumes it (deletes it), and emits the archive reminder as a plain stdout closing message, then exits 0.
- Removed dead `composed_dates_on_main()` and wired `composed_project_dates_on_main()` into `unmerged_draft_branches()` — the project-aware variant is more precise (requires `sessions/<project>/<file>` depth).
- Updated `README.md` and `docs/REFERENCE.md` to reflect the new design.

## How it works

```
git push → engineering-journal stub commit
  └─ PostToolUse fires → writes stub-pushed.flag, exits 0 (no error)

Session ends
  └─ Stop hook fires → finds flag, deletes it, prints closing reminder to stdout, exits 0
     └─ User sees: "[journal-stop-hook] Stub committed and pushed ... Archive this session now."
```

The sentinel is consumed before the message is emitted so the reminder fires exactly once per push.

## Test plan

- [x] `python3 -m py_compile claude/scripts/*.py` — all pass
- [x] `python3 claude/scripts/stub-push-archive-reminder.py < /dev/null` → `exit: 0`
- [x] `python3 claude/scripts/journal-stop-check.py < /dev/null` (no sentinel) → `exit: 0`, no output
- [x] Write sentinel (`echo 1 > ~/.claude/scratch/stub-pushed.flag`), run stop-check → prints `[journal-stop-hook] Stub committed and pushed ...`, `exit: 0`, sentinel deleted

Closes #145

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,9 @@ All hooks are advisory — none block tool execution.
 | PostToolUse (Bash) | `pr-merge-reminder.py` | Reminds to write a journal stub after `gh pr create` or `gh pr merge` |
 | PostToolUse (Bash) | `post-tool-use.py` | Auto-adds issues/PRs to configured GitHub Project |
 | PostToolUse (Bash) | `post-pr-merge-pull.py` | Fast-forwards local `main` after `gh pr merge` |
-| PostToolUse (Bash) | `stub-push-archive-reminder.py` | Prompts to archive session after a stub is pushed to engineering-journal |
+| PostToolUse (Bash) | `stub-push-archive-reminder.py` | Writes a sentinel flag after a stub is pushed to engineering-journal; Stop hook consumes it to remind Claude to archive |
 | Stop | `token-tracker.py` | Aggregates session token usage to `scratch/token-sessions.jsonl` |
-| Stop | `journal-stop-check.py` | Checks for stale open journal stubs at session end |
+| Stop | `journal-stop-check.py` | Checks sentinel flag and stale open journal stubs at session end; emits closing reminder if stub was pushed this session |
 | PostCompact | `post-compact.py` | Emits compaction status line (trigger type + remaining tokens) |
 | Git pre-push | `hooks/pre-push` | Warns when branch merge base diverges from `origin/main` in squash-merge repos |
 
--- a/claude/scripts/journal-stop-check.py
+++ b/claude/scripts/journal-stop-check.py
@@ -1,17 +1,20 @@
 #!/usr/bin/env python3
 """
-Stop hook: remind user of stale journal work at session end.
+Stop hook: remind Claude to archive after a stub push, and remind the user
+of stale journal work at session end.
 
-Two checks (same logic as new-day-journal-check.py):
+Check 1 — stub-push sentinel:
+  If stub-push-archive-reminder.py wrote a sentinel flag (meaning a stub was
+  pushed to engineering-journal this session), consume the flag and emit a
+  closing message reminding the user to call ccd_session_mgmt__archive_session.
+
+Checks 2–3:
 1. Stale *_draft.md / *.stub.md files from before today
 2. Unmerged remote draft/* branches
 
 Also cleans up orphaned draft files: physical files left on disk as untracked
 after git rm. This prevents new-day-journal-check.py false positives on the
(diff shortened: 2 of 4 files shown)


brownm09:
## Review: fix: flush stdout before sys.exit(2) in stub-push-archive-reminder

### Summary

Ready to merge with one non-blocking fix recommended. The core change is correct: `stub-push-archive-reminder.py` now writes a sentinel file and exits 0 on every code path; `journal-stop-check.py` consumes the sentinel at session end and emits the reminder as a plain stdout message with exit 0. This is the right pattern for advisory hooks in this Claude Code environment. One dead-code function (`composed_project_dates_on_main`) was introduced but never wired in — it should be removed or connected before merge. The PR title and body describe an intermediate iteration (exit-2 Stop hook approach) that was abandoned; the reference docs (`REFERENCE.md`) are correct, but the PR body is misleading.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/journal-stop-check.py` — `composed_project_dates_on_main()`
`composed_project_dates_on_main()` is added in this diff but never called. `unmerged_draft_branches()` still calls `composed_dates_on_main()` (the simpler variant). The docstring claims it suppresses false positives from new-day branches cut from prior draft branches — if that bug is real, the wiring step was omitted. If it's speculative, the function shouldn't land yet.
**Suggestion:** Either wire it into `unmerged_draft_branches()` in place of `composed_dates_on_main()`, or remove it. Dead code in a hook script inflates maintenance surface without benefit.

---

### Questions for Author

- Was `composed_project_dates_on_main()` intended to replace the `composed_dates_on_main()` call inside `unmerged_draft_branches()`? The docstring suggests a specific false-positive scenario — is that an observed problem or speculative?
- The PR body describes the exit-2 Stop hook design (the intermediate iteration that was abandoned), and the test plan checklist says "prints `{"systemMessage": "..."}`, `exit: 2`" — but the actual implementation exits 0 with plain stdout. Should the PR body be updated to reflect what actually landed, for future reference?

---

### Style Notes

The PR title ("fix: flush stdout before sys.exit(2)") no longer describes the fix — the final approach removed `sys.exit(2)` entirely and adopted a sentinel file. Worth updating the title to something like "fix: replace exit-2 PostToolUse pattern with sentinel file approach" before merge, so the git log is accurate.

---

PR #148, opened 2026-05-01
feat: reduce token expenditure in /cover-letter skill

### LINKED ISSUE #147: Reduce token expenditure in /cover-letter skill
## Problem

The \`/cover-letter\` skill spawns a draft subagent (Step 6) that receives all context files verbatim (~2,000 tokens per letter), even though the main agent already holds that content in context. For batch runs this cost compounds per letter without proportional quality gain.

Four concrete inefficiencies identified:
1. Draft subagent spawned even when Sonnet is selected — the re-pass is redundant, the main agent can draft inline
2. Full \`accomplishments.md\` (~540 tokens) passed to subagent even though only 4–8 rows are relevant per JD
3. Stable context files (\`style_rules.md\`, \`accomplishments.md\`, \`VOICE_SYNOPSIS.md\`) re-read on every letter in a session rather than reused
4. 400-word ceiling in the skill contradicts the canonical 475-word ceiling in \`CLAUDE.md\` and \`style_rules.md\`

## Proposed fixes

- Default draft model to Sonnet (was Opus); add inline drafting path for Sonnet that skips the subagent and self-check subagent entirely
- Add Step 5c to filter accomplishments to relevant rows before passing to Opus subagent
- Add session-cache notes to Steps 3, 5, 5b so files aren't re-read on letters 2-N in a batch
- Fix word ceiling: 400 → 475 in Steps 6 and 9
- Write ADR 009 documenting the analysis

## Projected savings

| Scenario | Before | After | Saved |
|---|---|---|---|
| Single Sonnet letter | ~4,400 tokens | ~2,450 tokens | ~44% |
| Batch of 5 Sonnet letters | ~22,000 tokens | ~7,300 tokens | ~67% |
| Single Opus letter | ~4,400 tokens | ~4,050 tokens | ~8% |
| Batch of 5 Opus letters | ~22,000 tokens | ~15,750 tokens | ~28% |

brownm09:
## Summary

- Sonnet is now the default draft model; letters draft inline in the main agent (no subagent spawn), eliminating ~2,520 tokens per Sonnet letter
- Opus path adds Step 5c to filter accomplishments to relevant rows before subagent pass (~200–350 tokens saved)
- Session-cache notes on Steps 3, 5, 5b skip re-reads on letters 2-N in a batch (~1,170 tokens saved per subsequent letter)
- Word-count ceiling corrected 400 → 475 in Steps 6 and 9 (authoring error vs. canonical ceiling)
- ADR 009 documents the full token-profile analysis and decision rationale

## Projected savings

| Scenario | Before | After | Saved |
|---|---|---|---|
| Single Sonnet letter | ~4,400 tokens | ~2,450 tokens | ~44% |
| Batch of 5 Sonnet letters | ~22,000 tokens | ~7,300 tokens | ~67% |
| Single Opus letter | ~4,400 tokens | ~4,050 tokens | ~8% |
| Batch of 5 Opus letters | ~22,000 tokens | ~15,750 tokens | ~28% |

## Test plan

- [ ] Run `/cover-letter` with a real JD, select Sonnet: confirm no Agent subagent spawned, self-check runs inline, word ceiling is 475
- [ ] Run `/cover-letter` a second time in the same session: confirm Steps 3, 5, 5b skip file reads (session cache)
- [ ] Run `/cover-letter`, select Opus: confirm Opus subagent spawned, receives filtered accomplishments, Haiku self-check follows
- [ ] Verify `docs/adr/INDEX.md` lists ADR 009 and the file exists

Closes #147

### DIFF
--- a/claude/skills/cover-letter/SKILL.md
+++ b/claude/skills/cover-letter/SKILL.md
@@ -71,6 +71,14 @@ If PROCEED, continue.
 
 Otherwise: Read `C:/Users/brown/Git/job-search/context/style_rules.md` in full. Internalize all rules before drafting. Do not proceed until read.
 
+## Step 3b — Read universal style rules
+
+**Session cache:** If `prose_style.md` is already in context from a previous letter this session, skip this file read and re-use the content already loaded.
+
+Otherwise: Read `C:/Users/brown/Git/job-search/context/prose_style.md` in full.
+
+The `## Universal Self-Check` section from this file will be used in Step 6 (Sonnet inline check) or Step 7 (Haiku subagent) to verify the draft against universal prose rules that are not duplicated in the cover-letter-specific check.
+
 ## Step 4 — Select model letter
 
 Read `C:/Users/brown/Git/job-search/models/INDEX.md`.
@@ -119,7 +127,12 @@ Constraints:
 - Body word ceiling: 475 words (body paragraphs only)
 - Output is Markdown only
 
-After drafting, run a self-check inline — do not spawn a separate subagent. Using the "Cover-Letter-Specific Self-Check" section from `style_rules.md` (already in context from Step 3), check the draft for every violation. Fix any violations before continuing. Report "PASS" or list each fix made.
+After drafting, run a self-check inline — do not spawn a separate subagent. Run both checks in sequence:
+
+1. **Universal Self-Check** — from `prose_style.md` `## Universal Self-Check` (already in context from Step 3b). Applies to all external-facing content: em-dashes, filler phrases, AI-tell patterns, rhythm, passive voice, etc.
+2. **Cover-Letter-Specific Self-Check** — from `style_rules.md` `## Cover-Letter-Specific Self-Check` (already in context from Step 3). Cover-letter extensions only: word count, HQ address, markdown format, closing construction, leadership philosophy presence, etc.
+
+Fix any violations before continuing. Report "PASS" or list each fix made with the offending text.
 
 Collect the final draft as DRAFT. Skip Step 7 and proceed to Step 8.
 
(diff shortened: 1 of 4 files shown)


brownm09:
## Review: feat: reduce token expenditure in /cover-letter skill

### Summary

Ready to merge with one question answered. The optimization logic is sound: the branching is clean, the session-cache guard correctly evaluates to true/false based on actual context (so a `/compact` between letters would correctly trigger a re-read), and the 400→475 correction is accurate. One non-blocking correctness gap pre-dates this PR but the inline Sonnet self-check makes it more visible: the self-check instruction targets only the `## Cover-Letter-Specific Self-Check` section, which explicitly does *not* duplicate the Universal Self-Check items. Those universal rules are not verified at the self-check step. Whether this is intentional is the single question worth answering before landing.

---

### Non-Blocking Findings

**[correctness]** `SKILL.md` — Step 6 Sonnet path and Step 7 Opus path

The self-check in both paths targets the `## Cover-Letter-Specific Self-Check` section from `style_rules.md`. That section's header says: "Run *after* the universal self-check from `prose_style.md` — items below are cover-letter-specific *extensions*, not duplicates." `prose_style.md` is never read by the skill, so the 23-item Universal Self-Check (em-dash ban, filler phrases, AI-tell patterns, rhythm checks, etc.) is applied during *drafting* but not explicitly verified at self-check time. The Sonnet inline path makes this more visible than the prior Haiku subagent path, because the inline check is now the only safety net. Pre-existing gap — this PR did not introduce it.

**Suggestion:** In the Sonnet Step 6 self-check instruction, add: "Also apply the Universal Self-Check rules from prose_style.md — they are not repeated in the Cover-Letter-Specific section." This works without reading the file because the LLM has the rules in context from training and from style_rules.md's header. For the Haiku Opus path in Step 7, add the same line to the subagent prompt.

---

**[maintainability]** `SKILL.md` — Steps 3, 5, 5b

The session-cache guard appears three times with near-identical boilerplate. If the caching strategy ever needs to change (e.g., a "force-refresh" flag, or a different condition), three places need updating.

**Suggestion:** Not worth a follow-up PR on its own — document the pattern in a comment at the top of Step 3 so the next editor knows the three steps are intentionally parallel.

---

### Questions for Author

- Is it intentional that the self-check (both Sonnet inline and Haiku Opus) covers only the cover-letter-specific extensions and not the Universal Self-Check items? If yes, the gap is accepted by design — add a note in the ADR's Consequences section. If no, the non-blocking finding above is actually blocking.

---

### Style Notes

The ADR References section mentions `job-search/CLAUDE.md` — the local repo name. The GitHub remote is `brownm09/career-playbook`. A reader following the reference on GitHub would have to know the mapping. Consider `job-search/CLAUDE.md (brownm09/career-playbook)` for clarity.

---

PR #150, opened 2026-05-01
fix: remove ExitWorktree from pre-merge workflow

### LINKED ISSUE #149: Remove ExitWorktree from pre-merge workflow — session-bound, always no-op after compact
## Problem

The documented merge procedure calls `ExitWorktree action: remove` before `gh pr merge`. ExitWorktree only works in the session that called `EnterWorktree` — that state is in-memory only and is lost on every `/compact`. So ExitWorktree is always a no-op in post-compact sessions (the common case for longer work).

When ExitWorktree is a no-op, the fallback in CLAUDE.md (manual `git worktree remove` + `git branch -D`) is frequently skipped because the compact summary doesn't preserve the detailed protocol. Claude improvises and runs `gh pr merge` directly, triggering a cascade:

- `gh` fails to checkout main locally (worktree blocks it)
- `gh` fails to delete local branch (branch is checked out)
- Rebase/recovery attempts follow
- Eventually: merge completed — because the *remote* merge always succeeds

**Token waste:** ~3,500–4,500 tokens (6–8 extra tool calls + recovery prose) per worktree merge in a post-compact session.

## Root Cause

`ExitWorktree` is session-bound. The two-path conditional defending against this state loss fires on *every* compact — the fallback is the common case, not the edge case.

## Fix

Remove the ExitWorktree pre-merge step. `gh pr merge --squash --delete-branch` always completes the remote work (merge + remote branch delete) regardless of local state. Local worktree and branch cleanup already runs weekly via `prune-merged-worktrees.py`. The pre-merge cleanup is purely cosmetic and not worth the complexity.

brownm09:
## Summary

- `ExitWorktree action: remove` is session-bound — it only works in the session that called `EnterWorktree`. After `/compact`, that state is gone, so it is always a no-op in post-compact sessions (the common case for longer work).
- The two-path fallback (ExitWorktree → if no-op, manual `git worktree remove`) was frequently skipped after compact because the compact summary doesn't preserve the detailed protocol, causing a 6–8 tool call recovery cascade (~4k wasted tokens) on every worktree merge.
- `gh pr merge --squash --delete-branch` always completes the remote merge and remote branch delete regardless of local state. Local errors are benign. Weekly prune handles local cleanup.

## Test plan

- [x] `python3 -m py_compile claude/scripts/*.py` — passes (no Python files changed)
- [x] Read line 85 of `claude/CLAUDE.md` — ExitWorktree instruction and manual-git fallback removed; single unconditional instruction in place
- [ ] Smoke test: next worktree merge, confirm the recovery chain no longer appears

Closes #149


brownm09:
## Review: fix: remove ExitWorktree from pre-merge workflow

### Summary

Ready to merge. The change is a 1-line doc fix that correctly eliminates a session-bound precondition that was always a no-op after `/compact`. The PR description clearly documents the root cause, the tradeoff (local artifacts linger up to ~7 days instead of being cleaned pre-merge), and the acceptance condition. No blocking findings.

---

### Non-Blocking Findings

**[maintainability]** `claude/CLAUDE.md:85`
The new text says `gh` "will fail to check out main locally and **may** fail to delete the local branch." In practice, running `gh pr merge` from within a worktree will always fail both of these local steps — the worktree holds the branch checked out, so checkout and branch-delete are deterministically blocked, not probabilistic. The hedge introduces mild ambiguity about when to expect these errors.
**Suggestion:** Replace "may fail to delete the local branch" with "will fail to delete the local branch" for accuracy.

---

### Questions for Author

- The global `CLAUDE.md` states: "The design decisions behind the rules in this file are recorded in `docs/adr/`." Is there an existing ADR covering the ExitWorktree-before-merge workflow (or the decision to remove it) that should be created or updated? The plan file documents the rationale well — the question is whether it should be promoted to a durable ADR rather than living only in the plan file and this PR body.

---

PR #152, opened 2026-05-02
config: allow Write to scratch dir without permission prompt

### LINKED ISSUE #151: Allow Write to scratch dir without permission prompt
When executing plans with bypass permissions, temp file writes to `C:/Users/brown/.claude/scratch/` still trigger a permission prompt because the directory is outside the project root. Add a global `permissions.allow` rule for `Write(C:/Users/brown/.claude/scratch/**)` to `claude/settings.json` so scratch writes are silently permitted in all sessions.

brownm09:
## Summary
- Adds `permissions.allow` rule for `Write(C:/Users/brown/.claude/scratch/**)` to `claude/settings.json`
- Eliminates permission prompts when I write temp files to the scratch directory during plan execution
- One-time global fix — no per-repo boilerplate needed

## Test plan
- [ ] Verify `python3 -m py_compile claude/scripts/*.py` passes (docs/config-only change, no scripts modified)
- [ ] Start a new session with bypass permissions active; confirm scratch dir writes (e.g. `gh` JSON output, temp parsing files) proceed without a prompt

Closes #151

### DIFF
--- a/claude/settings.json
+++ b/claude/settings.json
@@ -1,5 +1,10 @@
 {
   "model": "claude-sonnet-4-6",
+  "permissions": {
+    "allow": [
+      "Write(C:/Users/brown/.claude/scratch/**)"
+    ]
+  },
   "hooks": {
     "PreToolUse": [
       {
--- a/claude/settings.json
+++ b/claude/settings.json
@@ -2,7 +2,8 @@
   "model": "claude-sonnet-4-6",
   "permissions": {
     "allow": [
-      "Write(C:/Users/brown/.claude/scratch/**)"
+      "Write(C:/Users/brown/.claude/scratch/**)",
+      "Edit(C:/Users/brown/.claude/scratch/**)"
     ]
   },
   "hooks": {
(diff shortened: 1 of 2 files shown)


brownm09:
## Review: config: allow Write to scratch dir without permission prompt

### Summary
Ready to merge. This is a minimal, well-scoped config change — 5 added lines, zero deleted, one file. The `permissions.allow` rule is correctly placed in `settings.json` and the glob pattern matches the canonical scratch path. No correctness, security, or reliability issues. One non-blocking note on path portability.

---

### Non-Blocking Findings

**[maintainability]** `claude/settings.json:4`
The path `C:/Users/brown/.claude/scratch/**` hardcodes the Windows username. The rule is functionally correct for this machine, but if `dev-env` is ever bootstrapped on a second machine (or a new Windows account), the allow rule will silently fail to match and the permission prompts will return — with no obvious indication why.
**Suggestion:** Add a comment above the rule (or in the PR description / a nearby doc) noting that this path is machine-local and must be updated on re-bootstrap. Alternatively, track the expected path in `docs/REFERENCE.md` under a "Machine-local settings" note so it surfaces during onboarding.

---

### Questions for Author
- The rule covers `Write(...)`. Do any scratch-dir operations use the `Edit` tool (modifying an existing file in place) rather than `Write`? If so, an `Edit(C:/Users/brown/.claude/scratch/**)` entry may also be needed — though most temp-file patterns (write, read, delete) would only touch `Write`.

---

PR #156, opened 2026-05-02
fix: add Bash(TMPFILE=*) allow rule for review temp file writes

### LINKED ISSUE #155: fix: review skill temp file write triggers permission prompt
## Problem

The `/review` skill's Step 9 writes the review body to a temp file using a Bash heredoc (`cat > "$TMPFILE" << 'EOF'`). The full Bash command string starts with `TMPFILE=`, so no existing allow rule in `permissions.allow` matches it — a permission prompt fires even in bypass mode.

## Fix

Replace the Bash heredoc write with an explicit instruction to use the **Write tool** for the temp file step. `Write(C:/Users/brown/.claude/scratch/**)` is already in the allow list, so no new permission rule is needed. Pass the scratch path directly to `gh pr comment --body-file` in the subsequent Bash call (which is already covered by `Bash(gh pr comment *)`).

brownm09:
## Summary
- The `/review` skill's Step 9 writes the review body to a temp file using a Bash heredoc (`TMPFILE=... cat > "$TMPFILE" << 'EOF'`). The full command string starts with `TMPFILE=`, so no existing allow rule matched it — a permission prompt fired in bypass mode.
- Fix: add `Bash(TMPFILE=*)` to `permissions.allow`. This covers any Bash command that starts with a `TMPFILE=` variable assignment, which is the recognizable idiom for scratch temp file setup.
- Restore the explicit heredoc pattern in SKILL.md Step 9 (previously a comment placeholder).
- Document the decision and the rejected Write tool alternative in ADR-010.

## Test plan
- [ ] Run `/review <PR-URL>` in a session with bypass permissions mode active
- [ ] Confirm no permission prompt appears for the `TMPFILE=` heredoc step or the `gh pr comment` step

Closes #155


brownm09:
## Review: fix: use Write tool for review temp file to avoid permission prompt

### Summary
Ready to merge. The change correctly fixes the `/review` skill's Step 9 permission prompt by adding `Bash(TMPFILE=*)` to `permissions.allow` and restoring the full heredoc pattern in the skill spec. ADR-010 documents both the decision and why the Write tool alternative was rejected. The one non-blocking issue is that the PR title and body are stale — they describe the initial Write tool approach, not the `Bash(TMPFILE=*)` allow rule that was actually implemented. This doesn't affect correctness; the ADR references section explicitly calls it out.

---

### Non-Blocking Findings

**[maintainability]** `PR title and body`
The PR title ("fix: use Write tool for review temp file to avoid permission prompt") and the PR body ("Fix: instruct the skill to use the Write tool… No new allow rules required") describe the first approach, which was rejected mid-session. Anyone reading the git log or PR history will see a description that contradicts the actual diff.
The ADR's References section does call this out ("initial approach; revised to allow rule in same PR"), but that requires reading the ADR to reconcile the discrepancy.
**Suggestion:** Update the PR title to something like "fix: add Bash(TMPFILE=*) allow rule for review temp file writes" and revise the PR body to describe the allow rule approach before merging, so the squash commit message accurately reflects what landed.

---

PR #161, opened 2026-05-03
feat: add nightly-research routine

### LINKED ISSUE #160: Add nightly-research routine
Add a scheduled Claude agent that researches topics overnight using the 5-hour usage window.

## What

- New routine: `claude/routines/nightly-research/SKILL.md`
- New local repo: `C:/Users/brown/Git/research-notes/` (queue + output notes)
- Reference doc updates: `README.md` Routines table + `docs/REFERENCE.md` Routines section

## Behavior

- Reads `research-queue.md` for pending topics (checkbox format, one per line)
- Processes as many topics as fit in a 5-hour wall-clock budget (~15–20 min/topic)
- Writes structured markdown notes to `notes/YYYY-MM-DD/<slug>.md`
- Marks completed topics in Done; keeps failed topics in Pending with an annotation
- Commits to the local research-notes repo; no remote push during testing phase
- Fires at 3:00 AM local time (`0 8 * * *` UTC CDT; `0 9 * * *` CST in winter)

brownm09:
## Summary

- Adds `claude/routines/nightly-research/SKILL.md` — a scheduled agent that researches pending topics overnight using the 5-hour usage window
- Updates `README.md` Routines table with the new routine
- Updates `docs/REFERENCE.md` Routines section with full details

## Behavior

The routine fires at 3:00 AM local time (`0 8 * * *` UTC CDT). It reads `research-queue.md`, processes as many pending topics as fit in the 5-hour budget (~15–20 min/topic), writes one structured markdown note per topic to `notes/YYYY-MM-DD/<slug>.md`, marks completed items in Done, keeps failed items in Pending with an annotation, and commits locally. No remote push during testing phase.

## Test plan

- [x] `python3 -m py_compile claude/scripts/*.py` — SYNTAX OK (no new scripts; baseline confirmed)
- [ ] Register the routine via the `scheduled-tasks` MCP tool, add 1–2 test topics to the queue, and trigger on-demand to verify: note written, queue updated, git committed
- [ ] Confirm cron fires at correct local time the first night

Closes #160

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Autonomous scheduled agents in `claude/routines/`. No user interaction.
 |---|---|---|
 | Daily midnight UTC | `daily-journal-compose` | Assembles stub files into canonical journal entries and opens PRs |
 | Sunday midnight UTC | `prune-stale-worktrees` | Removes merged `claude/*` worktrees |
+| Nightly 3:00 AM local | `nightly-research` | Researches pending topics from the queue and writes structured markdown notes to `research-notes/` |
 
 ## Adding new configs
 
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Autonomous scheduled agents in `claude/routines/`. No user interaction.
 |---|---|---|
 | Daily midnight UTC | `daily-journal-compose` | Assembles stub files into canonical journal entries and opens PRs |
 | Sunday midnight UTC | `prune-stale-worktrees` | Removes merged `claude/*` worktrees |
-| Nightly 3:00 AM local | `nightly-research` | Researches pending topics from the queue and writes structured markdown notes to `research-notes/` |
+| Nightly 8:00 UTC (3 AM CDT) | `nightly-research` | Researches pending topics from the queue and writes structured markdown notes to `research-notes/` |
 
 ## Adding new configs
 
--- a/claude/routines/nightly-research/SKILL.md
+++ b/claude/routines/nightly-research/SKILL.md
@@ -2,6 +2,8 @@
 name: nightly-research
 description: Research pending topics from the queue overnight, write structured markdown notes, and update the queue with completion status.
 schedule: "0 8 * * *"
+# CDT (Apr–Nov): 0 8 * * *   → 3:00 AM local
+# CST (Nov–Mar): 0 9 * * *   → 3:00 AM local
 ---
 
 Research pending topics from the queue file. Run fully autonomously — never call AskUserQuestion.
@@ -145,7 +147,7 @@ Generate the slug from the title:
 
 Example: `"B-Tree vs LSM-Tree Storage Engines"` → `b-tree-vs-lsm-tree-storage-engines`
 
-Write the note to `${NOTES_DIR}/<slug>.md`. If a file at that path already exists (same topic run twice on the same day), append `-2` to the slug.
+Write the note to `${NOTES_DIR}/<slug>.md`. If a file at that path already exists, increment a counter suffix until a free name is found: `<slug>-2.md`, `<slug>-3.md`, etc.
 
 ---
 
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: feat: add nightly-research routine

### Summary

Ready to merge with two minor fixes. The routine is well-structured and the constraints section is thorough. Two reliability gaps need addressing before the first live run: a cleanup glob that never matches anything (harmless but misleading), and an all-failures edge case that produces a confusing error notification. Neither is a data-loss risk, but both will show up on the first nightly run and should be tightened now while the code is fresh.

---

### Blocking Findings

**[reliability]** `claude/routines/nightly-research/SKILL.md:230`

The Step 5 cleanup `rm -f "${SCRATCH}/nightly-research-"*"-$$."*` uses `$$` (current shell PID), but each Bash tool invocation runs in a separate process — the PID here will never match the PID that was in scope when the scratch files were created. The update script is already deleted inline (`node "$UPDATE_SCRIPT" && rm -f "$UPDATE_SCRIPT"`), so this glob would need to match *that* deletion to be belt-and-suspenders — but it can't, because `$$` differs.

In practice no scratch files will accumulate from the update script (it's cleaned up immediately), but any agent that adds a scratch file in the middle of the run and relies on this cleanup will silently leave a file behind. The pattern as written is a no-op and will never clean up anything.

**Fix:** Replace the glob with explicit `rm -f` calls for each file the agent creates, listed by name (not by PID pattern). Since `UPDATE_SCRIPT` is already deleted inline, Step 5 can simply confirm `rm -f "${SCRATCH}/nightly-research-update-"*.js` as a safety net, without relying on `$$`.

---

**[reliability]** `claude/routines/nightly-research/SKILL.md:193`

If every topic in a run fails (zero primary sources found for all), `notes/${RUN_DATE}/` is never written to and remains empty. `git add notes/${RUN_DATE}/` on an empty directory silently stages nothing (git doesn't track empty directories), so `git commit` exits nonzero with "nothing to commit." The spec's failure handler then fires: *"nightly-research: git commit failed — notes written to ${NOTES_DIR} but not committed"* — but no notes were written, making the notification actively misleading.

**Fix:** Before the `git add` in Step 4, check whether any note files were written:
```bash
if [ -z "$(ls -A "$NOTES_DIR" 2>/dev/null)" ]; then
  # No notes written — skip commit, only update queue if any annotations were added
fi
```
Adjust the notification text to distinguish "nothing to commit because all topics failed" from "commit command itself failed."

---

### Non-Blocking Findings

**[maintainability]** `claude/routines/nightly-research/SKILL.md:148`

Slug collision handling only appends `-2`. If the same topic appears twice in the queue (or is manually re-queued after completion), the first repeat gets `-2.md`, but a third run would try to write `-2.md` again and overwrite it — the spec doesn't mention checking for `-2` before writing.

**Suggestion:** Use an incrementing counter: check for `<slug>.md`, then `<slug>-2.md`, `<slug>-3.md` and pick the first available. In practice this is unlikely to matter before the Obsidian migration, but the current spec creates a silent overwrite case.

---

**[maintainability]** `claude/routines/nightly-research/SKILL.md:8` and `docs/REFERENCE.md:253`

The CDT/CST seasonal schedule adjustment is mentioned in two places (SKILL.md line 8 and REFERENCE.md) but requires a manual frontmatter edit twice a year. There's no reminder mechanism and it's easy to miss, causing the routine to fire at 2 AM or 4 AM for half the year.

**Suggestion:** Add a comment inside the frontmatter itself so it's visible when the file is opened for any reason:
```yaml
# CDT (Apr–Nov): 0 8 * * *   → 3:00 AM local
# CST (Nov–Mar): 0 9 * * *   → 3:00 AM local
schedule: "0 8 * * *"
```

---

### Questions for Author

- When a topic has a `<!-- attempted YYYY-MM-DD, no sources found -->` annotation from a prior failed run, the checkbox is still `- [ ]` so the agent will attempt it again on the next run. Is that the intended behavior (auto-retry until the user removes it), or should the agent skip annotated entries until the annotation is manually cleared?

---

### Style Notes

The README.md Routines table entry uses "Nightly 3:00 AM local" while the two existing entries use UTC ("Daily midnight UTC", "Sunday midnight UTC"). This breaks the column convention and makes the table harder to scan at a glance. Consider "Nightly 8:00 UTC (3 AM CDT)" to match the existing style.

---

PR #168, opened 2026-05-04
Add ADR-warrant check to global Git Workflow

### LINKED ISSUE #167: Add ADR-warrant check to global CLAUDE.md workflow
## Problem

There is no rule that prompts Claude (or me) to consider whether a change warrants an ADR. ADRs are written when I happen to remember to write one, which means decisions with non-obvious rationale can land in `git log` without a durable record. Six months later, the *why* is gone and a future change can innocently reverse the decision.

This came up while addressing review findings on [career-playbook#89](https://github.com/brownm09/career-playbook/pull/89): the multi-artifact writing-system decisions (universal/specific style split, per-artifact skills, project-level skill location, regression-detection requirement) had no ADR until prompted. Three of the four would have been hard to recover from `git log` alone.

## Decision needed

Add an "ADR-warrant check" rule to global `claude/CLAUDE.md` that fires at three explicit checkpoints:

1. Immediately after a plan is approved (post-`ExitPlanMode`).
2. Immediately after `gh pr create` returns.
3. Immediately before `gh pr merge`.

Criteria for "warrants an ADR":

- Changes a non-negotiable rule, hook, skill, or settings value documented anywhere in `claude/`.
- Introduces or restructures a directory under `claude/` (skills, hooks, scripts, routines).
- Establishes or changes a workflow rule that other CLAUDE.md files reference.
- The rationale would be hard to recover from `git log` alone six months later.

If warranted and not yet written, write the ADR before merging — never merge a qualifying change with no ADR. Pair with the existing INDEX.md update.

## Related

- Project-level version landed in [career-playbook#89 / `## Testing`](https://github.com/brownm09/career-playbook/blob/main/CLAUDE.md) — this issue is the global-instructions counterpart.
- See [`docs/adr/INDEX.md`](https://github.com/brownm09/dev-env/blob/main/docs/adr/INDEX.md) for the existing ADR catalog.

brownm09:
## Summary

- Adds an ADR-warrant check rule to `claude/CLAUDE.md` ## Git Workflow that fires at three explicit checkpoints (post plan-approval, post `gh pr create`, pre `gh pr merge`) with concrete criteria for when an ADR is warranted.
- ADR 011 records the decision itself (meta — exactly the kind of structural workflow rule that warrants an ADR per its own criteria).
- Project-level counterpart already landed in [career-playbook#89](https://github.com/brownm09/career-playbook/pull/89) `## Testing` section — this is the global-instructions equivalent.

## Test plan

- [x] `python3 -m py_compile claude/scripts/*.py` (syntax check, no scripts changed but verified untouched)
- [x] No `date -u` references introduced in stub-filename or branch-name contexts
- [x] ADR 011 added to docs/adr/INDEX.md

Closes #167


brownm09:
## Review: Add ADR-warrant check to global Git Workflow

### Summary

Merge with minor changes. This is a clean docs-only PR: a new CLAUDE.md workflow rule, a self-referential ADR capturing the decision, and an INDEX.md update. The writing is clear and the ADR format is consistent with existing entries. The single noteworthy gap is that checkpoint (1) — "post-`ExitPlanMode`" — has undefined behavior for the majority of sessions that never enter plan mode, creating a silent skip path for the earliest and cheapest capture point.

---

### Non-Blocking Findings

**[maintainability]** `claude/CLAUDE.md` (ADR-warrant check bullet)
The first checkpoint is defined as "immediately after a plan is approved (post-`ExitPlanMode`)" — but most sessions don't invoke plan mode (quick tasks, single-turn edits, conversational changes). There's no fallback: if `ExitPlanMode` never fires, checkpoint (1) silently doesn't happen, and the earliest/cheapest capture point is lost.
**Suggestion:** Add a fallback clause, e.g., "for sessions without an explicit planning phase, evaluate at the start of the first file edit instead." This closes the gap without requiring plan mode for every session.

**[maintainability]** `docs/adr/011-adr-warrant-check.md:64` (Verification section)
The Verification section says the project-level counterpart landed in `career-playbook CLAUDE.md ## Testing`. The PR body says the same. But ADR-warrant check is a workflow rule, not a testing rule — `## Git Workflow` would be the natural home. If the career-playbook placement really is `## Testing`, that's likely a misplacement in that repo.
**Suggestion:** Verify which section the career-playbook change landed in and correct the reference if it was misfiled.

---

### Questions for Author

- Criterion 1 ("changes a non-negotiable rule, hook, skill, or settings value documented in `claude/`") applies to all three of CLAUDE.md's sections — Git Workflow, Model Selection, Engineering Journal, etc. Does "non-negotiable" have a working definition, or is it intended as a smell test? If the latter, the ADR's own bloat-risk mitigation ("the other three criteria are concrete; criterion 4 is a backstop") actually applies to criterion 1 too, which weakens the argument that the criteria are "concrete enough to evaluate against."

---

### Style Notes

The new CLAUDE.md bullet is one of the longer ones in the section (~120 words), but it's consistent with the existing density of the Git Workflow bullets. No change needed.

---

PR #176, opened 2026-05-05
config: add git -C read-only allowlist entries to reduce routine permission prompts

### LINKED ISSUE #175: Add git -C * read-only allowlist entries to reduce routine permission prompts
Scheduled routines that operate on non-CWD repos (career-playbook, engineering-journal, lifting-logbook) use `git -C <path>` for read-only operations. These are not auto-allowed by Claude Code even though the equivalent plain `git` subcommands are, causing permission prompts that disrupt unattended routine runs.

Add allowlist entries for `git -C * show *`, `git -C * fetch *`, `git -C * ls-remote *`, `git -C * log *`, `git -C * status *`, `git -C * merge-base *`, `git -C * diff *`, `git -C * remote *`, `git -C * branch --show-current`, and `python3 -m py_compile *` to the global `claude/settings.json`.

brownm09:
## Summary
- Adds 10 allowlist entries to `claude/settings.json` for `git -C <path>` read-only subcommands that Claude Code does not auto-allow (unlike plain `git` equivalents)
- Covers the patterns routines use most: fetch, show, ls-remote, log, status, merge-base, diff, remote, branch --show-current, and python3 -m py_compile
- Skipped broad git -C * branch * (branch -D appears in transcripts), node -e * (arbitrary code execution), gh project * (mutating)

Closes #175

## Test plan
- [x] python3 -m py_compile claude/scripts/*.py passes
- [x] Settings JSON is valid
- [ ] Verify routines run without permission prompts on next scheduled execution

Generated with Claude Code

### DIFF
--- a/claude/settings.json
+++ b/claude/settings.json
@@ -13,7 +13,8 @@
       "Bash(git -C * status *)",
       "Bash(git -C * merge-base *)",
       "Bash(git -C * diff *)",
-      "Bash(git -C * remote *)",
+      "Bash(git -C * remote -v)",
+      "Bash(git -C * remote get-url *)",
       "Bash(git -C * branch --show-current)",
       "Bash(python3 -m py_compile *)"
     ]
--- a/claude/settings.json
+++ b/claude/settings.json
@@ -5,7 +5,17 @@
       "Write(C:/Users/brown/.claude/scratch/**)",
       "Edit(C:/Users/brown/.claude/scratch/**)",
       "Bash(gh pr comment *)",
-      "Bash(TMPFILE=*)"
+      "Bash(TMPFILE=*)",
+      "Bash(git -C * fetch *)",
+      "Bash(git -C * show *)",
+      "Bash(git -C * ls-remote *)",
+      "Bash(git -C * log *)",
+      "Bash(git -C * status *)",
+      "Bash(git -C * merge-base *)",
+      "Bash(git -C * diff *)",
+      "Bash(git -C * remote *)",
+      "Bash(git -C * branch --show-current)",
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: config: add git -C read-only allowlist entries to reduce routine permission prompts

### Summary
Ready to merge with one minor change. The entry selection is thoughtful — using `git -C * branch --show-current` instead of the broad `branch *`, and skipping `node -e *` for arbitrary execution safety. One finding: `git -C * remote *` is overly broad and admits mutating subcommands (`remote add`, `remote remove`, `remote set-url`) that the routines don't need.

---

### Non-Blocking Findings

**[security]** `claude/settings.json:13`
`Bash(git -C * remote *)` matches any `git remote` subcommand, including `remote add <name> <url>`, `remote remove <name>`, and `remote set-url <name> <url>`. These mutate the repository's remote configuration — they don't just list remotes.
The routines appear to use `git remote` for read-only inspection (`-v`, `get-url`). The allowlisting intent was read-only, but this pattern grants more than intended.
**Suggestion:** If routines only list remotes, narrow to `Bash(git -C * remote -v)` and add `Bash(git -C * remote get-url *)` if needed. Otherwise document that the mutating subcommands are accepted as a known tradeoff.

---

### Questions for Author
- Which specific `git remote` invocation do the scheduled routines use? That determines whether narrowing the pattern is worth a follow-up PR or can be noted as a known-accepted tradeoff.

---

### Style Notes
JSON formatting is consistent with the file's existing style. The trailing comma added to the `TMPFILE` line is correct.

---

PR #197, opened 2026-05-07
feat: detect and reconcile accidentally-resurrected merged draft branches

### LINKED ISSUE #196: Detect and reconcile accidentally-resurrected merged draft branches
## Problem

When a session pushes to a `draft/YYYY-MM-DD` branch that was already squash-merged, GitHub recreates the remote branch. This causes the stale-branch hook to fire on every subsequent prompt with a misleading message ("needs a PR to main" — but a PR was already merged).

The late commits contain real stub files from subsequent sessions that need to be incorporated into the journal.

## Solution

Three changes:

1. **`pre-push` hook** — block pushes to `draft/` branches that already have a merged PR. Emit a helpful error naming the correct target branch.

2. **`reconcile-late-stubs.py`** — when resurrection has already occurred, this script moves the orphaned stubs to the earliest unmerged draft branch (or today's), then deletes the stale remote branch.

3. **`new-day-journal-check.py`** — distinguish between two stale-branch cases:
   - Branch never had a PR → existing compose message
   - Branch had a merged PR but was resurrected → emit reconcile message with script invocation

## Acceptance criteria
- Pushing to an already-merged draft branch is blocked before it reaches remote
- `reconcile-late-stubs.py draft/2026-05-06` successfully moves the late stubs to `draft/2026-05-07` and deletes the stale branch
- Hook no longer fires for `draft/2026-05-06` after reconciliation

brownm09:
## Summary
- **pre-push hook**: blocks pushes to `draft/YYYY-MM-DD` branches in engineering-journal that already have a merged PR — prevents the resurrection at the source
- **`reconcile-late-stubs.py`**: moves orphaned stub files from a resurrected branch to the earliest unmerged draft branch (or today's), then deletes the stale source branch
- **`new-day-journal-check.py`**: adds `resurrected_draft_branches()` to distinguish "never composed" from "merged but resurrected", emitting the correct actionable message for each case

## Test plan
- [ ] `python3 -m py_compile` passes on all changed scripts
- [ ] `python3 ~/.claude/scripts/reconcile-late-stubs.py draft/2026-05-06` moves late stubs to `draft/2026-05-07` and deletes stale branch
- [ ] Hook no longer fires for `draft/2026-05-06` after reconciliation

## Testing
Run `python3 -m py_compile claude/scripts/*.py` — passes.

Closes #196

### DIFF
--- a/claude/hooks/pre-push
+++ b/claude/hooks/pre-push
@@ -37,7 +37,7 @@ while read local_ref local_sha remote_ref remote_sha; do
           echo "Check your current branch — did you mean a different draft branch?" >&2
           echo "To reconcile orphaned commits after an accidental push:" >&2
           echo "  python3 ~/.claude/scripts/reconcile-late-stubs.py $remote_branch" >&2
-          exit 1
+          exit 1  # intentionally skips per-repo hook chain — push is fully blocked
         fi
       fi
     fi
--- a/claude/hooks/pre-push
+++ b/claude/hooks/pre-push
@@ -20,6 +20,28 @@ while read local_ref local_sha remote_ref remote_sha; do
     echo "If this repo uses squash merges, the PR may target the wrong base." >&2
     echo "Consider cherry-picking onto a fresh branch from origin/main." >&2
   fi
+
+  # Block pushes to already-merged draft/YYYY-MM-DD branches in engineering-journal.
+  # Pushing to a merged branch resurrects it, causing the stale-branch hook to fire
+  # on every subsequent prompt until orphaned commits are reconciled.
+  remote_branch="${remote_ref#refs/heads/}"
+  if echo "$remote_branch" | grep -qE '^draft/[0-9]{4}-[0-9]{2}-[0-9]{2}$'; then
+    remote_url=$(git remote get-url origin 2>/dev/null || echo "")
+    if echo "$remote_url" | grep -q "engineering-journal"; then
+      if command -v gh >/dev/null 2>&1; then
+        merged=$(gh pr list --repo brownm09/engineering-journal \
+          --state merged --head "$remote_branch" --json number 2>/dev/null || echo "[]")
+        if [ "$merged" != "[]" ] && [ -n "$merged" ]; then
+          echo "ERROR: $remote_branch already has a merged PR." >&2
+          echo "Pushing to this branch would resurrect it and cause stale-branch hook noise." >&2
+          echo "Check your current branch — did you mean a different draft branch?" >&2
+          echo "To reconcile orphaned commits after an accidental push:" >&2
+          echo "  python3 ~/.claude/scripts/reconcile-late-stubs.py $remote_branch" >&2
+          exit 1
+        fi
+      fi
+    fi
+  fi
 done
 
 # Chain to per-repo hook if one exists (preserves Husky, git-secrets, corporate hooks, etc.)
--- a/claude/scripts/new-day-journal-check.py
+++ b/claude/scripts/new-day-journal-check.py
@@ -175,6 +175,44 @@ def unmerged_draft_branches() -> list[str]:
         return []
 
 
+def resurrected_draft_branches() -> list[str]:
+    """Return remote draft/* branch names that were already merged but still exist.
+
+    A branch is resurrected when its date has a composed journal file on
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: feat: detect and reconcile accidentally-resurrected merged draft branches

### Summary
Needs one revision before merge. The pre-push hook and the new `resurrected_draft_branches()` detection are correct and well-guarded. The reconcile script's core flow — fetch, verify merged PR, extract late commits, copy stubs to target, delete source — is sound. However, `open-prs.jsonl` is included in the file-move set without checking whether the target branch already has its own copy. Overwriting the target's version is the most likely outcome and would corrupt PR tracking state. Fix is small and isolated to `get_stub_files_in_commits`. Three non-blocking findings below.

---

### Blocking Findings

**[correctness]** `claude/scripts/reconcile-late-stubs.py:71-76`

`get_stub_files_in_commits` includes `open-prs.jsonl` in the list of files to copy to the target branch. Later in `main()`, these files are written into a worktree of the target and committed without first checking whether the target already has `open-prs.jsonl`. The source branch's `open-prs.jsonl` reflects the state at the moment of the accidental push — it may predate PR merges that the target's sessions already recorded. Overwriting target's file with the source's re-introduces closed PRs as "open" or erases PR entries that were added by correct sessions. In the specific `draft/2026-05-06` → `draft/2026-05-07` scenario, `draft/2026-05-07` has been accumulating its own `open-prs.jsonl` updates all day; the source's version is older.

**Fix:** In `get_stub_files_in_commits`, exclude `open-prs.jsonl` from the collected paths. The stubs and manifests are what must move; `open-prs.jsonl` on the target branch is already the authoritative live state. If a PR was opened in the accidental session and missing from the target's file, the user can reconcile it manually — that edge case is far less harmful than silently corrupting the target's file. Alternatively, after copying files into the worktree, check whether `open-prs.jsonl` already exists (via `git show origin/<target>:path`) and skip the write if it does.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/new-day-journal-check.py:178-213`

`resurrected_draft_branches()` duplicates the `git ls-remote --heads origin refs/heads/draft/*` call and the remote-date-extraction loop from `unmerged_draft_branches()`. The two functions differ only in the final set predicate (`d not in merged` vs. `d in merged`). If the ls-remote timeout, error handling, or date-parsing logic needs to change, both copies must be updated together.

**Suggestion:** Extract a `remote_draft_dates() -> set[str]` helper that runs ls-remote and returns the set of date strings, then call it from both functions.

**[reliability]** `claude/hooks/pre-push:27-44`

The `exit 1` inside the `while read` loop exits the entire hook script, skipping the per-repo hook chain (`$REPO_HOOK "$@"`) at the bottom. For repos that have a local `.git/hooks/pre-push` (e.g., Husky, git-secrets), blocking a bad push to engineering-journal silently skips their checks for all branches in that push batch, not just the bad one.

**Suggestion:** This is acceptable for blocking a single-branch push, which is the normal case. For multi-branch pushes it's a minor gap. Add a comment on the `exit 1` line noting that the chain is intentionally skipped on block, so a future reader doesn't try to "fix" it by moving the check after the loop.

**[maintainability]** `claude/scripts/reconcile-late-stubs.py:38`

`dict | None` union syntax in the return annotation requires Python 3.10+. The sibling script `new-day-journal-check.py` uses `set[tuple[str, str]]` and `list[str]` (3.9+ syntax) but not the `|` union operator. If the machine's Python is 3.9, this script will fail to import with a `TypeError` at annotation evaluation time (only matters when `from __future__ import annotations` is absent, which it is here).

**Suggestion:** Change to `Optional[dict]` with `from typing import Optional`, or verify Python 3.10+ is the floor for this machine and add a version comment.

---

### Questions for Author

- In the pre-push hook, `--json number` is used (just checking existence), while `reconcile-late-stubs.py` uses `--json number,mergedAt,url`. That asymmetry is intentional and correct — the hook only needs a boolean. Worth a one-line comment so a future editor doesn't "complete" the field list?
- If `find_target_branch` returns `draft/TODAY` as the fallback (no unmerged branch exists after the source date), `ensure_remote_branch_exists` creates it from `origin/main`. Is this the desired behavior, or should the script prompt the user before creating a new branch they may not have intended?

---

PR #199, opened 2026-05-07
feat(#198): suppress hook noise in Claude-managed worktree sessions

### LINKED ISSUE #198: Suppress hook noise in Claude-managed worktree sessions
## Problem

When multiple Claude Code sessions run in parallel, `new-day-journal-check.py` and `multi-worktree-alert.py` inject shared state into every session's context window — including Claude-managed worktree sessions where the warnings are not actionable.

- `new-day-journal-check.py` fires a `git ls-remote` call to engineering-journal and emits draft-branch / resurrected-stub warnings in every session
- `multi-worktree-alert.py` emits the full sibling worktree list as a `systemMessage` in every session, even though each worktree session already knows its own location from the system prompt

## Solution

Add a guard to both hooks: if `cwd` is inside a Claude-managed worktree (`.claude/worktrees/` in the path), exit silently. Also add guidance to `CLAUDE.md` on when a separate local clone is preferable to a worktree for fully independent parallel sessions.

brownm09:
## Summary

- Adds a worktree guard to `new-day-journal-check.py` and `multi-worktree-alert.py`: both hooks now exit silently when `cwd` is inside `.claude/worktrees/`
- Adds CLAUDE.md guidance on when a separate local clone is preferable to a worktree for fully independent parallel sessions
- Adds ADR-015 recording the `.claude/worktrees/` detection pattern for future hook authors; updates REFERENCE.md hook descriptions

Closes #198

## Test plan

- [x] `python3 -m py_compile claude/scripts/new-day-journal-check.py claude/scripts/multi-worktree-alert.py` — passes clean
- [ ] Open a second Claude session in a Claude-managed worktree — confirm no `[journal-hook]` or `[worktree-alert]` output on first prompt
- [ ] Submit a prompt in a main-checkout session with stale draft branches present — confirm both hooks still fire


brownm09:
## Review: feat(#198): suppress hook noise in Claude-managed worktree sessions

### Summary

Ready to merge with minor non-blocking observations. No blocking findings. The guards are correctly placed before all subprocess I/O in both hooks, the safe-exit invariant is maintained, and ADR-015 documents the pattern clearly for future authors. The single noteworthy issue is a prose/code mismatch in the ADR: it describes the guard as checking "consecutive path components" but the actual code only checks set membership — a non-critical discrepancy but worth aligning.

---

### Non-Blocking Findings

**[maintainability]** `docs/adr/015-suppress-hook-noise-in-claude-worktrees.md:26` and `claude/scripts/*.py`  
The ADR Decision section says "consecutive path components" but the guard pattern in both scripts (and in the Consequences code snippet) only checks `".claude" in _parts and "worktrees" in _parts` — set membership, not adjacency. A path like `C:/Users/worktrees/project/.claude/bar` would trigger the guard even though it's not a Claude-managed worktree.  
**Suggestion:** Either update the ADR prose to say "as path components" (removing "consecutive"), or tighten the guard to verify adjacency: `any(_parts[i] == ".claude" and i+1 < len(_parts) and _parts[i+1] == "worktrees" for i in range(len(_parts)))`. Given how unlikely a false-positive path is in practice, updating the prose is the cheaper fix.

**[maintainability]** `claude/scripts/new-day-journal-check.py` guard placement  
The worktree guard fires before `cleanup_stale_flags()`, so stale sentinel files are never cleaned up during worktree sessions. This is probably fine — stale flags are created by main-checkout sessions and cleaned up there — but it's an implicit assumption rather than an explicit decision.  
**Suggestion:** Add a one-line comment noting the cleanup skip is intentional, or move `cleanup_stale_flags()` above the guard if cleanup should still run in worktrees.

---

### Questions for Author

- Is skipping `cleanup_stale_flags()` in worktree sessions intentional? Or should flag cleanup still run there even when the rest of the hook is suppressed?

---

### Style Notes

The CLAUDE.md bash example hardcodes `C:/Users/brown/Git/<repo>`. The file already acknowledges machine-specific paths in the machine-local permissions section, so this is consistent with the existing pattern — but the path could use `<your-repos-dir>/<repo>` to signal it's a placeholder rather than a literal path on any machine.

---

PR #201, opened 2026-05-08
[chore] Document single-select option mutation hazard

### LINKED ISSUE #200: [chore] Document single-select option mutation hazard in global CLAUDE.md
## Problem

`updateProjectV2Field` with `singleSelectOptions` is a full replacement: option IDs regenerate on every mutation even when the unchanged options are passed through, and every project item that referenced an option by ID loses its assignment. Validated twice on the lifting-logbook project — most recently 2026-05-08 when adding an "Observability" Epic option, which silently wiped 89 items' Epic assignments.

The previous saved-memory guidance ("pass the full list to preserve assignments") is wrong and should be replaced with a backup-and-restore procedure that any project's CLAUDE.md can reference.

## Proposed Solution

Add a global note to dev-env `claude/CLAUDE.md` § GitHub Project that:

1. States `updateProjectV2Field` is always destructive
2. Requires a git-tracked snapshot to `.claude/backups/` before any single-select mutation on any project
3. Requires capturing new option IDs from the mutation response and updating that project's CLAUDE.md option-ID table in the same PR
4. Requires restoring assignments via per-item `gh project item-edit` calls keyed by the snapshot's epic-name → new option-ID mapping
5. Requires aborting and recovering from snapshot if a mutation runs without a prior snapshot commit

## Acceptance Criteria

- [ ] dev-env `claude/CLAUDE.md` includes the hazard note and procedure under § GitHub Project
- [ ] The note is positioned where it applies to any project, not just dev-env
- [ ] lifting-logbook CLAUDE.md cross-references or duplicates the procedure for the Epic field specifically

brownm09:
## Summary

- `updateProjectV2Field` with `singleSelectOptions` is always destructive — passing the existing options unchanged still regenerates option IDs and drops every item's prior assignment.
- Adds a global note and backup-and-restore procedure under § GitHub Project so the rule applies to any project, not just dev-env.
- Validated on lifting-logbook 2026-05-08: adding an Observability Epic option wiped Epic assignments on all 89 project items. Recovered 42 by reconstruction; remainder tracked in [lifting-logbook#203](https://github.com/brownm09/lifting-logbook/issues/203).

## Test plan

- [x] `grep -n 'date -u' claude/CLAUDE.md` — single match, in the testing-rule meta-text (not a stub filename or branch name)
- [x] Manual review of inserted text for the procedure being repo-agnostic (uses "any project" language, not lifting-logbook-specific paths)

Closes #200

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -146,7 +146,17 @@ All new dev-env issues must be added to the **Dev Env** project and given an Imp
 | Impact | `PVTSSF_lAHOAjEKvM4BWKFezhRgkNc` | High=`08de2558`, Medium=`6320e8a6`, Low=`d8a85c2f` |
 | Why | `PVTF_lAHOAjEKvM4BWKFezhRgkN0` | (text) |
 
-> **Single-select option mutation hazard (applies to every project, not just dev-env).** Running `updateProjectV2Field` with `singleSelectOptions` is a full replacement — passing the existing options unchanged still produces new option IDs and drops every item's prior assignment for that field. Before adding, removing, or renaming any single-select option on any project, snapshot the current per-item assignments to a git-tracked file under the project repo's `.claude/backups/`, commit it, run the mutation, then restore by re-issuing `gh project item-edit` for each item using the new option IDs. Update the project's CLAUDE.md option-ID table in the same PR. If a mutation runs without a prior snapshot commit, stop and recover from the latest snapshot before any other work. Validated against lifting-logbook on 2026-05-08 — assignments were lost despite passing the full option list.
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [chore] Document single-select option mutation hazard

### Summary
Ship it. Two-line docs-only addition that captures a validated incident (lifting-logbook 89-item Epic wipe on 2026-05-08) as a global hazard rule. Content is accurate, repo-agnostic, and correctly placed under § GitHub Project so it surfaces alongside the field-ID tables every project's automation references. No blocking findings.

---

### Non-Blocking Findings

**[maintainability]** `claude/CLAUDE.md:149`
The hazard note is a single ~90-word sentence inside one blockquote. The 4-step procedure (snapshot → commit → mutate → restore) is the actionable core, but it's buried mid-sentence — a reader scanning under stress (i.e., right after they realize they just wiped a field) has to parse prose to recover the steps.
**Suggestion:** Keep the lead sentence as-is, then break the procedure into a numbered list. The lifting-logbook CLAUDE.md already uses this shape — mirroring it here keeps the two docs consistent.

---

### Questions for Author
- Should this rule cross-link to lifting-logbook#203 (or a generalized "post-incident recovery" issue) so future readers can see what the recovery actually looked like in practice? Useful precedent, but adds a project-specific outbound link from a global doc — judgment call.

---

PR #203, opened 2026-05-10
[feat] Auto-install npm packages in Claude-managed worktrees

### LINKED ISSUE #202: feat: auto-install npm packages in Claude-managed worktrees
## Problem

Claude Code worktrees are created at `<repo>/.claude/worktrees/<name>/`. Each worktree has an independent working directory — `node_modules` is NOT shared with the main repo. Worktrees are created without running `npm install`, so the first prompt in a new worktree often produces test failures like:

> pre-existing failures due to missing npm packages [...] in this worktree — unrelated to my changes

This is noise that wastes time and requires a manual `npm install` before work can begin.

## Solution

Add a `UserPromptSubmit` hook (`worktree-npm-install.py`) that:
1. Detects if the cwd is a Claude-managed worktree (`.claude/worktrees` in the path)
2. Checks that `package.json` exists (npm repo)
3. Checks that `node_modules` is absent
4. If all three: runs `npm ci` (or `npm install` if no lockfile) and emits a systemMessage

The `node_modules` check doubles as the sentinel — once installed, the hook is a no-op for all subsequent prompts in that worktree.

## Scope

Global hook in dev-env — applies to all npm repos using Claude Code worktrees. No-op in non-npm repos.

brownm09:
## Summary

- Adds `worktree-npm-install.py` — a `UserPromptSubmit` hook that detects Claude-managed worktrees and auto-runs `npm ci` (or `npm install`) when `node_modules` is absent
- Registers the hook in `claude/settings.json`
- Documents the decision in ADR-016

## Problem

Fresh Claude-managed worktrees have no `node_modules`. The first test run produces failures like _"pre-existing failures due to missing npm packages — unrelated to my changes"_, requiring a manual `npm install` before real work can begin.

## How it works

On every `UserPromptSubmit`, the hook:
1. Checks that `.claude/worktrees` appears in the cwd path (is a Claude-managed worktree)
2. Checks that `package.json` exists (npm repo)
3. Exits silently if `node_modules` already exists (sentinel — installed, nothing to do)
4. Runs `npm ci` if `package-lock.json` exists, otherwise `npm install`
5. Emits a \`systemMessage\` confirming success or explaining failure; always exits 0

## Scope

Global hook — applies to all npm repos using Claude Code worktrees. No-op in non-npm repos, non-worktree sessions, and already-installed worktrees.

## Test plan

- [ ] Open a new Claude session in a fresh lifting-logbook worktree (no \`node_modules\`): first prompt should trigger install and emit a systemMessage
- [ ] Second prompt in the same worktree: hook is a no-op
- [ ] Session in a non-npm repo: hook exits silently
- [ ] Session in the main repo checkout (not a worktree): hook exits silently
- [ ] \`python3 -m py_compile claude/scripts/worktree-npm-install.py\` — passes ✓

Closes #202

### DIFF
--- a/claude/scripts/worktree-npm-install.py
+++ b/claude/scripts/worktree-npm-install.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+"""Claude Code UserPromptSubmit hook — auto-installs npm packages in Claude-managed
+worktrees when node_modules is absent.
+
+Claude-managed worktrees share the git object store but have independent working
+directories. node_modules is never present on first use, causing spurious test
+failures unrelated to the current change. This hook detects that condition and
+runs npm ci (or npm install if no lockfile) before Claude starts working.
+
+The node_modules directory check doubles as the sentinel — once installed, the
+hook exits silently for all subsequent prompts in the same worktree.
+
+Fires on every user prompt; exits silently when not applicable.
+
+Stdin JSON shape (UserPromptSubmit):
+  {
+    "hook_event_name": "UserPromptSubmit",
+    "session_id": "...",
+    "cwd": "..."
+  }
+
+Exit 0 always — advisory only, never blocks.
+"""
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+
+def main() -> None:
+    raw = sys.stdin.read().strip()
+    cwd = ""
+    if raw:
+        try:
+            cwd = json.loads(raw).get("cwd", "")
+        except json.JSONDecodeError:
+            pass
+
+    if not cwd:
+        sys.exit(0)
+
+    cwd_path = Path(cwd)
+    parts = cwd_path.parts
+
+    # Only run in Claude-managed worktrees (.claude/worktrees/<name> path structure).
+    if ".claude" not in parts or "worktrees" not in parts:
+        sys.exit(0)
+
+    # Verify .claude and worktrees appear as consecutive components.
+    try:
+        claude_idx = parts.index(".claude")
+        if claude_idx + 1 >= len(parts) or parts[claude_idx + 1] != "worktrees":
+            sys.exit(0)
+    except ValueError:
+        sys.exit(0)
+
+    # Only run in npm repos.
+    if not (cwd_path / "package.json").exists():
+        sys.exit(0)
+
+    # node_modules presence is the sentinel — already installed, nothing to do.
+    if (cwd_path / "node_modules").exists():
+        sys.exit(0)
+
+    # Choose npm ci (reproducible) when a lockfile exists, otherwise npm install.
+    has_lockfile = (cwd_path / "package-lock.json").exists()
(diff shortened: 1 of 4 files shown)


brownm09:
## Review: [feat] Auto-install npm packages in Claude-managed worktrees

### Summary

Ready to merge with one non-blocking reliability note. The hook is well-structured: correct hook event choice, minimal footprint (three `Path.exists()` checks on every prompt when not applicable), clean systemMessage output, and always-exit-0 safety. The ADR is thorough. One concern worth addressing before this runs in production: `subprocess.run(["npm", ...])` without `shell=True` may silently fail to locate `npm.cmd` on Windows — the `FileNotFoundError` catch swallows the failure without emitting the warning systemMessage, making the hook a no-op rather than an informative failure.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/worktree-npm-install.py:67–77`
`subprocess.run(["npm", "ci"], ...)` is called without `shell=True`. On Windows, `npm` is installed as `npm.cmd` — a batch file that Windows `CreateProcess` does not locate when the executable name has no extension. Python's subprocess on Windows does not expand PATHEXT automatically in non-shell mode. If Claude Code's process inherits a PATH where `npm.cmd` is the only form of npm, this raises `FileNotFoundError`, which the `except` on line 77 catches and turns into a silent `sys.exit(0)` — no warning systemMessage, no install, hook appears to succeed but did nothing.
**Suggestion:** Add `shell=True` to the `subprocess.run` call (and drop the list form: `cmd = "npm ci" if has_lockfile else "npm install"`), or use `"npm.cmd"` as the executable on `sys.platform == "win32"`. The shell=True approach is simpler and consistent with how Git Bash-targeted scripts in this environment typically resolve PATH.

---

**[maintainability]** `claude/scripts/worktree-npm-install.py:46–55`
The membership pre-check (`".claude" not in parts or "worktrees" not in parts`) on lines 46–47 is redundant with the consecutive-components check on lines 50–54. If `.claude` isn't in `parts`, `parts.index(".claude")` would raise `ValueError` — but the pre-check already exits before reaching it, so the `except ValueError` on line 54 is dead code. The two checks do the same work twice with overlapping coverage.
**Suggestion:** Remove the pre-check and the `try/except`, and use a simple linear scan or a single `".claude/worktrees" in str(cwd_path).replace("\\", "/")` idiom. Or keep the pre-check and drop the `try/except`. Either way, one check is sufficient.

---

### Questions for Author

- The ADR notes 30–120 second latency for large monorepos. Is there a plan to surface progress during that wait (e.g., a systemMessage *before* the install starts like "Installing packages — this may take a minute…")? Currently Claude receives no feedback until the install completes, which might look like a hang on first prompt in a large repo.

---

### Style Notes

The module-level docstring is thorough and well-placed. Minor: the "Stdin JSON shape" block in the docstring documents `session_id` but the script never uses it — not incorrect, but worth omitting from the documented shape to avoid implying the field is consumed.

---

PR #211, opened 2026-05-10
feat(hooks): reconcile open-prs.jsonl against GitHub state at session start

### LINKED ISSUE #209: Auto-reconcile open-prs.jsonl against GitHub PR state on session start
## Problem

`open-prs.jsonl` drifts from reality when sessions end before the PR-merge removal step runs. Stale entries propagate into `post-compact.py` reminders and Claude's session context, causing wasted turns cleaning up discrepancies.

Confirmed instance: dev-env #170 (ADR 012) is open but has no entry; an older session left it behind.

## Solution

A new `UserPromptSubmit` hook (`reconcile-open-prs.py`) that:
- Runs once per session (per-session sentinel in scratch/)
- Calls `gh pr view` for each entry in every `sessions/*/open-prs.jsonl`
- Removes entries whose PRs are MERGED or CLOSED
- Emits a `systemMessage` with surviving open PRs so Claude has correct session context from turn 1

The hook does **not** commit — modified files are picked up by the next stub commit.

## Also investigated

Audited all 20+ existing scripts; the open-prs reconciliation is the only gap with clear evidence of repeated token waste. Other recurring activities are already well-covered.

brownm09:
## Summary

- Adds `reconcile-open-prs.py` (`UserPromptSubmit` hook) that calls `gh pr view` for every entry across all `sessions/*/open-prs.jsonl` files in engineering-journal, removes stale entries whose PRs are MERGED or CLOSED, and emits a `systemMessage` with surviving open PRs so Claude has correct context from turn 1 without a manual file read.
- Runs once per session via a per-session sentinel in `scratch/`; always exits 0; fails safe on `gh` errors (leaves entry intact).
- Does **not** commit — modified files are left dirty and picked up by the next stub commit.
- Includes ADR 018 and reference doc updates.

## Test plan

- [x] `python3 -m py_compile claude/scripts/reconcile-open-prs.py` — passes
- [x] Manual run found and removed 6 stale merged entries (dev-env#152, dev-env#205, career-playbook#149, lifting-logbook#214, and 2 others), correctly preserved 3 open PRs
- [x] Second run with same session_id produced no output (idempotent)

Closes #209

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ All hooks are advisory — none block tool execution.
 | UserPromptSubmit | `journal-onboard-check.py` | Warns when the active project has no journal home in engineering-journal |
 | UserPromptSubmit | `turn-count-hook.py` | Warns when session context token count exceeds threshold |
 | UserPromptSubmit | `multi-worktree-alert.py` | Lists active worktrees in `repo:branch` format when ≥2 are open |
+| UserPromptSubmit | `reconcile-open-prs.py` | Removes stale entries from `open-prs.jsonl` whose PRs are now merged/closed; emits surviving open PRs as session context |
 | PreToolUse (Bash) | `pre-commit-branch-check.py` | Emits current branch as a checkpoint before `git commit` |
 | PreToolUse (Bash) | `pre-pr-create-check.py` | Emits test-verification checklist before `gh pr create` |
 | PostToolUse (Bash) | `pr-merge-reminder.py` | Reminds to write a journal stub after `gh pr create` or `gh pr merge` |
--- a/claude/scripts/get-project-item.sh
+++ b/claude/scripts/get-project-item.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# get-project-item.sh — resolve a GitHub Project item ID from an issue/PR number
+#
+# Usage: get-project-item.sh <issue-or-pr-number> [project-number] [owner]
+#
+# Outputs the project item node ID to stdout so callers can capture it:
+#   ITEM_ID=$(bash get-project-item.sh 42)
+#
+# Defaults: project 3, owner brownm09 (dev-env project board).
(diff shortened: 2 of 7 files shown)


brownm09:
## Review: feat(hooks): reconcile open-prs.jsonl against GitHub state at session start

### Summary

Ready to merge with minor changes. The implementation is clean, the sentinel pattern is applied correctly, and the fail-safe design (leave intact on `gh` error, always exit 0) is solid. One non-blocking issue worth addressing before merge: the `"unknown"` sentinel fallback creates a shared sentinel key that will cause the second session with a malformed hook payload to skip reconciliation entirely. Consequence is low-probability but worth closing.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/reconcile-open-prs.py:140`
`session_id` defaults to `"unknown"` when `json.loads` of stdin fails or when the payload has no `session_id` key. The sentinel file for this case (`open-prs-reconciled-unknown.flag`) is shared across all such sessions. The first session with a malformed payload writes the sentinel; every subsequent session in the same 30-day window that also receives a bad payload silently skips reconciliation — they all see the same flag and bail immediately.
**Suggestion:** Use a combination of `"unknown"` + a timestamp suffix so each malformed-payload session gets its own sentinel: `session_id = data.get("session_id") or f"unknown-{int(time.time())}"`. The sentinel still gets cleaned up after 30 days.

---

### Questions for Author

- `reconcile_file` raises no exception on a missing `url` field (the `repo_from_url` returns `None`, and the entry is preserved). Is that the intended behavior for malformed entries — keep them forever? A one-time log to stderr would make stale malformed entries visible without blocking.

---

### Style Notes

The `elif not all_removed: return` branch at the bottom of `main()` carries a comment "nothing to report — no files exist" but the condition also triggers when all files are empty (every JSONL file has zero parseable lines). The comment is slightly misleading; "nothing found or tracked" would be more accurate.

---

PR #214, opened 2026-05-10
fix: extend journal-compose today-guard to all paths; allow same-day draft push

### LINKED ISSUE #213: fix: extend journal-compose today-guard to all date paths; allow same-day push after accidental compose
Two related gaps in the mid-day compose protection introduced by PR #205:

1. **Pre-push hook blocks same-day recovery pushes.** After an accidental mid-day compose squash-merges `draft/YYYY-MM-DD`, the hook blocks all further pushes to that branch — even on the same calendar day when a recovery push is legitimate. The hook should skip the merged-PR check when the branch date matches today.

2. **journal-compose today-guard has an explicit-date bypass.** The guard added in PR #205 only fires on the auto-detected-branch path. Passing `/journal-compose YYYY-MM-DD` with today's date bypasses it entirely, making accidental same-day compose possible via explicit invocation. The guard should cover all date inputs; a `--force` flag provides the escape hatch for intentional same-day compose (manual end-of-day) without requiring interactive prompts.

The nightly `daily-journal-compose` routine is unaffected: it runs at midnight UTC and always passes yesterday's local date — never today's — so the extended guard never fires for it.

brownm09:
## Summary

- **Pre-push hook**: skip merged-PR check when branch date == today — same-day recovery pushes after an accidental mid-day compose are now unblocked (prior-day branches still blocked)
- **journal-compose guard**: extended from branch-detection path only to all date inputs; `--force` flag required for intentional same-day compose (end-of-day manual run); past-date invocations unaffected
- **CLAUDE.md**: end-of-day instruction updated to `/journal-compose --force`
- **ADR 017**: updated to document the extended scope, `--force` mechanism, and pre-push same-day exception

The nightly `daily-journal-compose` routine passes yesterday's UTC date — never today's — so it is unaffected by the guard change and requires no `--force`.

## Test plan

- [x] `python3 -m py_compile claude/scripts/*.py` — passes
- [x] `grep -n 'date -u' claude/CLAUDE.md` — single match in testing instructions paragraph (not a stub filename or branch name context)

Closes #213

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -422,8 +422,10 @@ Schema — one JSON line per open PR:
 - `/journal-compose` preserves this file unchanged in the merge-to-main commit so it carries forward to the next day.
 
 **End of day (last session):**
-1. Run `/journal-compose` — it discovers all stubs via manifest (or glob fallback), merges them,
-   produces the canonical 11-section document, and auto-merges the PR
+1. Run `/journal-compose --force` — it discovers all stubs via manifest (or glob fallback), merges
+   them, produces the canonical 11-section document, and auto-merges the PR. `--force` is required
+   when composing today's branch; past-date composition (`/journal-compose YYYY-MM-DD` for a prior
+   day) does not need the flag
 
 ---
 
--- a/claude/hooks/pre-push
+++ b/claude/hooks/pre-push
@@ -24,8 +24,12 @@ while read local_ref local_sha remote_ref remote_sha; do
   # Block pushes to already-merged draft/YYYY-MM-DD branches in engineering-journal.
   # Pushing to a merged branch resurrects it, causing the stale-branch hook to fire
   # on every subsequent prompt until orphaned commits are reconciled.
+  # Exception: same-day branches are allowed — recovery pushes after an accidental
+  # mid-day compose are legitimate. Prior-day branches are still blocked (stale-branch noise).
   remote_branch="${remote_ref#refs/heads/}"
   if echo "$remote_branch" | grep -qE '^draft/[0-9]{4}-[0-9]{2}-[0-9]{2}$'; then
+    branch_date=$(echo "$remote_branch" | grep -oE '[0-9]{4}-[0-9]{2}-[0-9]{2}')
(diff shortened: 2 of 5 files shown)


brownm09:
## Review: fix: extend journal-compose today-guard to all paths; allow same-day draft push

### Summary

Ready to merge. The logic is correct — the pre-push date comparison is sound, the `--force` guard covers all invocation paths, and the nightly routine remains unaffected. One non-blocking gap: the "Draft branch recovery" procedure in `CLAUDE.md` still instructs `Run /journal-compose` without `--force`; following it verbatim on the same calendar day will hit the new guard. The other finding is a stale description string in the SKILL.md frontmatter.

---

### Non-Blocking Findings

**[maintainability]** `claude/CLAUDE.md` — Draft branch recovery section, Step 5

Step 5 of the "Draft branch recovery" block reads: "When running `/journal-compose`, ensure the engineering-journal working tree is on `draft/YYYY-MM-DD-recovery`." The recovery scenario is, by definition, a same-day compose (the original branch was merged mid-day and you're composing at end of that same day). With the new guard, the resolved date will equal today and the skill will halt unless `--force` is passed. A user following the recovery procedure verbatim will hit the guard without any hint of why.

**Suggestion:** Change Step 5 to `Run /journal-compose --force` and add a one-line note: "`--force` required because the target date is still today."

---

**[maintainability]** `claude/skills/journal-compose/SKILL.md` — frontmatter `description` field

The `argument-hint` was correctly updated to `[YYYY-MM-DD] [--force]`, but the `description` value still ends with "Invoke as /journal-compose [YYYY-MM-DD]." — the `--force` option is invisible to anything that reads the description for help text or indexing.

**Suggestion:** Append `[--force]` to the inline invocation hint in the description string, or remove the invocation example entirely (the `argument-hint` field is the canonical place for this).

---

### Questions for Author

- In the pre-push hook, the same-day `continue` fires before the `engineering-journal` URL check, so it applies to any repo whose branch matches `draft/YYYY-MM-DD` — not just engineering-journal. Is that intentional scope, or should the exception also require the remote URL to match `engineering-journal`? (In practice it's harmless since no merged-PR check would fire for non-EJ repos anyway, but the intent isn't documented.)

---

PR #216, opened 2026-05-10
fix(post-compact): tell user to reply to trigger /review after compact

### LINKED ISSUE #215: PostCompact hook: user sees review prompt but doesn't know to reply
After `/compact`, the PostCompact hook outputs a `systemMessage` instructing Claude to invoke `/review`. However, `systemMessage` is context injection — it only activates on the **next user-initiated turn**. Since Claude passively waits after compaction, the message displays in the output block but nothing happens.

**Fix:** add a stderr status line telling the user to reply to trigger the review. The existing `systemMessage` already has the right instruction; users just need to know to type something.

brownm09:
## Summary

- `systemMessage` from a PostCompact hook is context injection, not an auto-trigger — Claude only acts on the **next user-initiated turn**, so after `/compact` it waits silently
- Add a stderr status line (`open PR #N — reply to this message to trigger /review`) so users know to type something
- The existing `systemMessage` already contains the right instruction; users just needed a visible cue

## Test plan

- [ ] Run `/compact` in a session with an open PR in `open-prs.jsonl`
- [ ] Confirm new stderr line appears alongside the output block
- [ ] Type any message — Claude should immediately invoke `/review <url> --post-comment`
- [ ] `python3 -m py_compile claude/scripts/post-compact.py` passes (verified)

Closes #215

### DIFF
--- a/claude/scripts/post-compact.py
+++ b/claude/scripts/post-compact.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """PostCompact hook — emit a status line and, for manual compactions with open PRs,
-inject a systemMessage so Claude auto-invokes /review without user input."""
+inject a systemMessage so Claude invokes /review on the next user reply."""
 import json
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


brownm09:
Shelving — the underlying mechanism (systemMessage) requires user input to trigger; no path to fully automatic without a hook system change. Closing without merge.

PR #220, opened 2026-05-11
feat(review): add Step 2c documentation coverage check

### LINKED ISSUE #219: Add context/ index-hygiene validator for career-playbook (and pattern for other repos)
## Problem

PR brownm09/career-playbook#168 added \`context/sentence_density.md\` — a load-bearing file referenced by the new modular cover-letter workflow's M-B.2 step — without updating \`context/README.md\` (the index) or \`context/letter_writer_briefing.md\`'s source-file header. The miss surfaced two weeks later via PR brownm09/career-playbook#170 (post-mortem: ADR 016).

A briefing regeneration in that window would have silently dropped sentence-density guidance, and the README is the entry point new sessions use to orient on the directory.

## Outcome

A validator that fails when a file under \`career-playbook/context/\` is added, renamed, or removed without a corresponding change to \`career-playbook/context/README.md\` in the same diff. Mirrors the doc-reconciliation enforcement pattern from ADR 019 (dev-env).

## Scoping decisions

1. **Placement:** career-playbook \`scripts/validate.sh\` extension (catches devs who run validate), pre-push hook (catches everyone), or both? Pre-push is the stronger guarantee but adds a hook to maintain.
2. **Pattern reuse:** the check is generic enough — \"directory X has a README that must mention every file in X\" — that it could ship as a reusable helper in dev-env's \`claude/scripts/\` and be invoked from any repo's \`validate.sh\`. Decide whether to scope this issue to career-playbook only or generalize.
3. **README naming convention:** require an exact filename match (`sentence_density.md` literal substring), or accept any reference to the file in the README? Exact match is stricter but catches typos in references; substring match is more forgiving.

## Why now (rationale, not just symptom)

The modular workflow ships entirely through `context/` and `.claude/skills/` files. Any future file added there is at risk of the same miss. The cost of the miss is a workflow regression that may not surface for weeks. The cost of the validator is small (one shell function, one CI/hook hookup).

## Companion work

- career-playbook PR #170 (this issue's trigger) — already addresses the immediate README index gap manually and adds an "Index hygiene rule" pointer to a future validator.
- career-playbook ADR 016 — documents the rationale for the hardening that surfaced this gap.

brownm09:
## Summary
- Adds **Step 2c — Documentation Coverage Check** to the `/review` skill: LLM-judged check that fires when files are added, deleted, renamed, or significantly rewritten, walks ancestor directories up to depth 3, and flags missing README updates.
- Classifies a missing update to an existing README as **blocking**; a missing README in a directory that would benefit from one as **non-blocking suggestion**.
- Rationale recorded in [ADR 020](docs/adr/020-doc-coverage-in-review.md). Replaces the career-playbook-specific mechanical validator originally proposed in the issue with a repo-agnostic semantic check.

## Why this design (instead of the mechanical script in the issue)
A path-only script must choose between (a) flagging every structural change in a directory with a README (false positives) or (b) hardcoding known index directories per repo (does not generalize, requires maintenance). LLM judgment sidesteps both. Full discussion in ADR 020 § Rationale.

## Files
- `claude/skills/review/SKILL.md` — new Step 2c; Step 6 blocking/non-blocking buckets updated to acknowledge Step 2c findings.
- `docs/REFERENCE.md` — `/review` description rewritten to describe both Step 2b and Step 2c.
- `docs/adr/020-doc-coverage-in-review.md` — new ADR.
- `docs/adr/INDEX.md` — ADR 020 row added.

## Test plan
- [x] `python3 -m py_compile claude/scripts/*.py` — passes (no Python files changed).
- [ ] After merge, run `/review <next-PR>` on a PR that adds a new file to a directory with a README; confirm the review surfaces a blocking documentation finding.
- [ ] Run `/review` on a content-only edit PR; confirm no spurious documentation findings.

Closes #219

### DIFF
--- a/claude/skills/review/SKILL.md
+++ b/claude/skills/review/SKILL.md
@@ -98,6 +98,57 @@ If the table exists (dev-env and any repo that adopts the pattern):
    > **Fix:** Per the Documentation Maintenance table in `CLAUDE.md`, update the relevant
    > section(s) of `README.md` and/or `docs/REFERENCE.md` in this PR.
 
+Proceed to Step 2c.
+
+---
+
+## Step 2c — Documentation Coverage Check
+
+Applies to both PR_URL mode and DIFF_MODE.
+
+For every file that was **added, deleted, or renamed** in the diff (structural changes),
+and for every file that was **significantly rewritten** (purpose or functionality changed,
+not a minor fix or content update), apply the following check:
+
+1. Identify the file's directory and all ancestor directories up to depth 3 from the
+   changed file (stop at repo root).
+2. For each ancestor level, check whether a `README.md` exists at that level.
+   - In PR_URL mode: `git show origin/<headRefName>:<dir>/README.md 2>/dev/null`
+   - In DIFF_MODE: infer from the diff whether a README is present at that level.
+3. Use judgment to determine whether this specific change warrants a documentation update
+   at any of those levels. Consider:
+   - **New file in a directory that has a README index** → the index likely needs a new entry.
+   - **Deleted or renamed file in a directory with a README** → the index reference may need removal or updating.
+   - **File rewritten to serve a different purpose** → the README description may be stale.
+   - **Minor bug fix, typo correction, or content-only update to an existing file** → no README change needed.
+   - **Cross-tree impact**: a skill, hook, or workflow change referenced in a parent README or
+     in a related doc outside the immediate directory tree → flag the relevant parent or sibling
+     doc even if it is not in the same directory as the changed file.
+4. For each directory level where a README **exists** and the change **warrants** an update
+   and the README **does not appear in the diff**: record a **blocking Documentation finding**.
+5. For each directory level where no README exists but one **would add navigational value**
+   (e.g., a directory now has several files with no index): record a **non-blocking
+   Documentation finding** suggesting creation.
+
+If no ancestor directories have READMEs and no cross-tree impact is identified, note
+"No README coverage gaps found." and proceed to Step 3.
+
+**Example findings:**
+
+> **[documentation]** `context/sentence_density.md` added without README update
+> `context/README.md` is an index of all files in this directory. Adding a new file here
+> without adding an entry leaves sessions that orient from `context/README.md` unaware of
+> the new file — the career-playbook#168 incident that prompted issue #219.
+> **Fix:** Add a `### \`sentence_density.md\`` entry to `context/README.md` describing
+> its purpose and when to load it.
+
+> **[documentation]** `claude/skills/new-skill/` added; top-level README not updated
+> `README.md` contains a Skills table that lists all available skills. Adding a skill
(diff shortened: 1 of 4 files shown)


brownm09:
## Review: feat(review): add Step 2c documentation coverage check

### Summary
Merge with one minor revision. The Step 2c design is sound, the ADR rationale is well-developed, and the Step 2b → 2c → 3 control flow is clean. The single blocking finding is a correctness gap in the DIFF_MODE branch: a diff alone cannot reveal whether an unchanged README exists at a given ancestor level, so Step 2c will silently underreport in DIFF_MODE. Step 2c self-check on this PR produces no findings (REFERENCE.md updated, INDEX.md updated, no orphan ancestor directories).

---

### Blocking Findings

**[correctness]** `claude/skills/review/SKILL.md` — Step 2c DIFF_MODE detection of existing READMEs
Step 2c instructs: "In DIFF_MODE: infer from the diff whether a README is present at that level." A diff only contains changed files. An unchanged `README.md` at an ancestor directory does not appear in a pasted diff, so this rule cannot detect the most common case Step 2c is designed to catch (existing README that should have been updated). The result is silent false negatives in DIFF_MODE — the reviewer will be told "No README coverage gaps found" when in fact one or more exist.
**Fix:** Either (a) state explicitly that Step 2c's blocking-finding branch is skipped in DIFF_MODE and only the "suggest README creation" non-blocking branch applies, or (b) instruct the reviewer to ask the user to list ancestor READMEs when DIFF_MODE is active. Option (a) is simpler and matches the constraint honestly.

---

### Non-Blocking Findings

**[maintainability]** `claude/skills/review/SKILL.md` — Example finding references a specific historical incident
The first example finding cites "the career-playbook#168 incident that prompted issue #219." Example findings in a skill template are read on every `/review` invocation; embedding historical case references here risks the example aging poorly once the incident is far in the past, and it leaks context that belongs in ADR 020. The example reads more crisply as a generic "Adding a new file here without updating the index leaves sessions that orient from `context/README.md` unaware of the new file."
**Suggestion:** Drop the "the career-playbook#168 incident that prompted issue #219" clause from the example. Keep the issue reference in ADR 020 where it has durable context.

**[performance]** `claude/skills/review/SKILL.md` — `git show` invocation count is unbounded by file count
Step 2c invokes `git show origin/<headRefName>:<dir>/README.md` for each ancestor (up to depth 3) of each changed file. For a 50-file PR this is up to 150 git show calls. Not a correctness issue, but for large diffs the cost is non-trivial.
**Suggestion:** Add a one-line note that the reviewer should deduplicate ancestor paths across changed files before issuing `git show` — `set(ancestors)` once, query once per unique directory.

---

### Questions for Author

- Step 2c's "significantly rewritten" criterion is judgment-only with no operational handle. Is that intentional (relying entirely on the reviewing model's discretion), or should the SKILL provide a coarse heuristic (e.g., "≥50% of lines changed or the file's stated purpose in its docstring/header changed")? Either answer is defensible; just want to confirm the omission is deliberate.

---

PR #235, opened 2026-05-16
[chore] Fix post-tool-use.py KeyError for dev-env; add generic required_fields hook config

### LINKED ISSUE #233: [chore] Enforce "issue must be added to project board with Impact/Why before work" workflow
## Problem

The global `claude/CLAUDE.md` rule for the **Dev Env** GitHub Project ([§ GitHub Project](https://github.com/brownm09/dev-env/blob/main/claude/CLAUDE.md#github-project)) requires that every new dev-env issue be added to project #3 with an Impact rating and a Why description **before work begins**. The rule is documented but has no forcing function — after `gh issue create` returns, work routinely proceeds straight to file edits and the three-step project-add sequence is silently skipped.

**Triggering incident:** dev-env #231 was created in the session that shipped #265 (test coverage enforcement) and was never added to the project board. The omission was discovered only at PR merge, by which time the issue had auto-closed via `Closes #N`. A late-stage `gh project item-add` then succeeded but the follow-up `item-edit` calls for Impact and Why raced against the auto-close and failed with `The item does not exist in the project` GraphQL errors — leaving the item on the board with status Done but no Impact / no Why.

This is structurally the same failure mode as #265: a documented rule with no enforcement.

## Affected workflows

Both projects share this shape and are equally exposed:

- **Dev Env** (project #3) — Impact + Why required.
- **Lifting Logbook** (project #2) — Epic + Milestone required (lifting-logbook `CLAUDE.md`).

A generic fix that works for any repo whose `CLAUDE.md` declares a project-board workflow is preferable to two separate per-repo hooks.

## Proposed solutions

1. **`new-issue` wrapper script** (preferred). A `claude/scripts/new-issue.sh` that:
   - Takes the issue title and body.
   - Runs `gh issue create`.
   - Reads the project workflow from the repo's `CLAUDE.md` (or a `.claude/project-workflow.json` if simpler to parse than markdown).
   - Adds the issue to the named project.
   - Prompts for Impact/Why (dev-env) or Epic/Milestone (lifting-logbook).
   - Sets all required fields in one atomic flow.
   - Fails loudly if any step fails — so the gap surfaces during issue creation, not at merge time.

2. **`PostToolUse` Bash hook** (lighter, but softer). Detect `gh issue create` against any repo whose `CLAUDE.md` declares a project workflow, and emit a blocking reminder to run the project-add sequence. This is closer in spirit to `pr-merge-reminder.py`.

Recommend (1) because it eliminates the failure mode entirely rather than warning about it.

## Acceptance criteria

- [ ] A new dev-env issue cannot be created via the documented workflow without ending up on project #3 with Impact and Why populated.
- [ ] The same mechanism works (or has a clear extension point) for lifting-logbook project #2.
- [ ] If a hook approach is chosen, it must fail-closed: if the project-add fails, the failure is visible to the session, not silently swallowed.
- [ ] `claude/CLAUDE.md` and `docs/REFERENCE.md` updated to reference the new tool.
- [ ] ADR added if the chosen design changes the issue-creation rule shape (it likely does — moving from manual three-step sequence to a single command is a workflow rule change per ADR-022's warrant criteria).

## References

- dev-env #231 (the triggering incident — created without project assignment; field writes failed at merge time).
- lifting-logbook #265 / [PR #266](https://github.com/brownm09/lifting-logbook/pull/266) and dev-env [PR #232](https://github.com/brownm09/dev-env/pull/232) — same failure-mode template (documented rule, no enforcement) that this issue addresses for the project-board step.
- `claude/CLAUDE.md` § GitHub Project — the rule being enforced.

### LINKED ISSUE #234: [chore] Fix post-tool-use.py KeyError for dev-env and add generic required_fields config
## Problem

Closes #233

`post-tool-use.py`'s `format_reminder` hardcodes `config['epic_field_id']`, which raises a `KeyError` for dev-env sessions (dev-env's `hook-config.json` has no `epic_field_id`). The hook crashes after the project-add step, so the Impact/Why reminder is never shown. Sessions proceed silently without setting required project-board fields.

## Solution

- Refactor `format_reminder` to iterate over a generic `required_fields` array in `hook-config.json`, supporting `single_select`, `text`, and `milestone` field types.
- Keep `epic_field_id`/`milestones` as backward-compatible fallback for lifting-logbook.
- Add missing top-level `try/except` in `__main__` block (hook safe-exit invariant).
- Update dev-env's `hook-config.json` with Impact and Why field definitions.
- Update `claude/CLAUDE.md`, `docs/REFERENCE.md`, and add ADR 023.

brownm09:
## Summary

- **Root cause fixed:** `format_reminder` in `post-tool-use.py` hardcoded `config['epic_field_id']`, raising a `KeyError` for dev-env sessions (which have no such key). The hook crashed silently after the project-add step, so Impact/Why instructions were never shown and sessions proceeded without setting required fields.
- **Generic `required_fields` config:** refactored `format_reminder` to iterate a `required_fields` array in `hook-config.json`, supporting `single_select`, `text`, and `milestone` field types. Backward-compat fallback preserves existing lifting-logbook behavior.
- **Safe-exit guard added:** top-level `try/except` in `__main__` block, satisfying the ADR-007 hook invariant that was missing.
- **CLAUDE.md updated:** hook-driven workflow is now documented as the primary path; manual 3-step sequence demoted to fallback.
- **REFERENCE.md updated:** `required_fields` schema documented; deprecated `epic_field_id`/`milestones` keys noted.
- **ADR-023 added.**

## Test plan

- [x] `python3 -m py_compile claude/scripts/post-tool-use.py` passes
- [x] `hook-config.json` not in PR diff (machine-local, as required)
- [ ] Manual smoke test: create a dev-env issue via `gh issue create` and confirm hook fires with Impact/Why commands in the session output
- [ ] Backward-compat: lifting-logbook `hook-config.json` uses `epic_field_id` — verify fallback path still generates the Epic reminder (no script changes required there)

## Coverage gate

This change fixes a hook crash (no new testable behavior in the automated-test sense). The `post-tool-use.py` hook is exercised by running `gh issue create` and observing the PostToolUse output — tracked by the manual smoke test above.

Closes #233
Closes #234

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ All hooks are advisory — none block tool execution.
 | PreToolUse (Bash) | `pre-commit-branch-check.py` | Emits current branch as a checkpoint before `git commit` |
 | PreToolUse (Bash) | `pre-pr-create-check.py` | Emits test-verification checklist and documentation-gap warning before `gh pr create` |
 | PostToolUse (Bash) | `pr-merge-reminder.py` | Reminds to write a journal stub after `gh pr create`, `gh pr merge`, or `git push` (when the pushed branch has an open PR) |
-| PostToolUse (Bash) | `post-tool-use.py` | Auto-adds issues/PRs to configured GitHub Project |
+| PostToolUse (Bash) | `post-tool-use.py` | Auto-adds issues/PRs to configured GitHub Project; exits 2 with `required_fields` reminders |
 | PostToolUse (Bash) | `post-pr-merge-pull.py` | Fast-forwards local `main` after `gh pr merge` |
 | PostToolUse (Bash) | `post-pr-merge-project.py` | Auto-moves linked issue to Done on configured GitHub Project after `gh pr merge` |
 | PostToolUse (Bash) | `stub-push-archive-reminder.py` | Writes a sentinel flag after a stub is pushed to engineering-journal; Stop hook consumes it to remind Claude to archive |
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -171,7 +171,16 @@ All new dev-env issues must be added to the **Dev Env** project and given an Imp
 | Medium | Recurs periodically or silently degrades correctness over time |
 | Low | Nice-to-have; low frequency or easily worked around |
 
-**Workflow — run after `gh issue create`:**
+**Workflow — automated via PostToolUse hook:**
+
+After `gh issue create` succeeds, the `post-tool-use.py` hook fires automatically and:
+
+1. Adds the issue to project #3 (`gh project item-add`).
+2. Exits with code 2, printing the exact `gh project item-edit` commands to set Impact and Why.
+
+**Run those commands immediately — before any file edits.** Do not proceed to implementation until both Impact and Why are set. This is the same forcing function as the test-before-PR rule: the hook output is visible in the session and must be acted on.
+
+**Fallback (if the hook did not fire or the item-add failed):** run the three steps manually:
 
 ```bash
 # Requires project scope — add once if needed: gh auth refresh -s project
--- a/claude/scripts/post-tool-use.py
+++ b/claude/scripts/post-tool-use.py
@@ -94,30 +94,70 @@ def add_to_project(url: str, config: dict) -> str | None:
 
 
 def format_reminder(item_type: str, url: str, item_id: str, config: dict) -> str:
-    epic_options = config.get("epic_options", {})
-    epic_list = "\n".join(
-        f"      {name}: {opt_id}"
-        for name, opt_id in epic_options.items()
-    )
-    milestones = config.get("milestones", [])
-    milestone_list = ", ".join(f'"{m}"' for m in milestones) if milestones else "<milestone>"
+    lines = [
+        f"[project-hook] {item_type} added to project.",
+        f"  URL:     {url}",
+        f"  Item ID: {item_id}",
+    ]
 
-    return (
-        f"[project-hook] {item_type} added to project.\n"
-        f"  URL:     {url}\n"
-        f"  Item ID: {item_id}\n"
-        f"\n"
-        f"  Set Epic field:\n"
-        f"    gh project item-edit \\\n"
-        f"      --project-id {config['project_node_id']} \\\n"
-        f"      --id {item_id} \\\n"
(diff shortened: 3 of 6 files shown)


brownm09:
## Review: [chore] Fix post-tool-use.py KeyError for dev-env; add generic required_fields hook config

### Summary

Ready to merge with one minor latent issue noted as non-blocking. The root cause fix is correct — the generic `required_fields` loop replaces the hardcoded dict key access that crashed `format_reminder` for dev-env sessions, and the safe-exit guard satisfies ADR-007. Documentation reconciliation is complete: `docs/REFERENCE.md` and `docs/adr/` are both updated, ADR-023 is written. Coverage gate: valid deferral (no automated test infrastructure for hooks; py_compile passes; manual smoke test tracked in the PR body).

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/post-tool-use.py:~140`

`format_reminder` uses `config['project_node_id']` with bare dict key access inside the `single_select` and `text` branches. The surrounding `try/except Exception: sys.exit(0)` in `__main__` will catch a `KeyError` here and exit 0 silently — producing no reminder output at all. This is the same symptom as the original bug (silent no-op after a successful project-add), just caused by a different missing key and now guarded rather than crashing. A config where `project_number`/`project_owner`/`required_fields` are present but `project_node_id` is absent would trigger it.

**Suggestion:** Use `config.get('project_node_id', '<project-node-id>')` in both the `single_select` and `text` branches so a misconfigured hook emits a visibly broken command (immediately obvious) rather than silently producing no output.

---

**[documentation]** `README.md` — Hooks table entry for `post-tool-use.py`

The entry reads "Auto-adds issues/PRs to configured GitHub Project." The hook now also emits the required-field commands via exit 2, which is the primary enforcement mechanism for Impact/Why. The README description undersells it. The Documentation Maintenance rule triggers only on add/rename/remove (not behavior change), so this isn't a blocking gap, but the table now misrepresents the hook's role.

**Suggestion:** Update the README Hooks table entry to something like: "Auto-adds issues/PRs to configured GitHub Project; exits 2 with `required_fields` reminders."

---

### Style Notes

No style concerns. The loop structure in `format_reminder` is clear and the `lines +=` / `lines.append` split is consistent.

---

PR #239, opened 2026-05-23
feat(hooks): block Write/Edit/NotebookEdit to canonical repo root from worktrees

### LINKED ISSUE #238: PreToolUse hook: block Write/Edit to canonical repo root when session is in a worktree
## Problem

When a Claude Code session is running inside a worktree (`<repo>/.claude/worktrees/<name>/`), `Write` / `Edit` / `NotebookEdit` calls that use an absolute path beginning at the canonical repo root silently land on the main working tree instead of the worktree. The session continues without error — artifacts scatter between main and worktree until a later grep/awk fails to find a file at the expected path. Recovery is mechanical (`cp` into the worktree, `rm` orphans on main) but costs tokens and risks orphans being missed entirely.

Documented three times in career-playbook (most recent: PR #275 stub `2026-05-22_140307.stub.md`). Tracked downstream at brownm09/career-playbook#276.

The failure is harness-level, not project-level — any repo using worktrees is vulnerable. A per-project CLAUDE.md heuristic only protects that one repo and still relies on the model reading and obeying it mid-session, which is exactly what has failed three times.

## Proposed fix

New `PreToolUse` hook in `~/.claude/hooks/` (dev-env-owned via the symlink) matching `Write | Edit | NotebookEdit`.

Logic:

1. If `cwd` (or `$CLAUDE_PROJECT_DIR`) matches `.../​.claude/worktrees/<name>/`, extract the worktree root.
2. Read the tool input's `file_path`.
3. If `file_path` is absolute and starts with the canonical repo root but does NOT start with the worktree root → exit non-zero with a message naming both paths and the corrected worktree-relative path.
4. Otherwise pass.

Wired in `claude/settings.json` under `hooks.PreToolUse`.

## Judgment calls baked in

- **Block, not rewrite.** Silently rewriting the path is the same class of silent failure in the opposite direction. Block forces the model to re-issue with the right path, which is the behavior we want to reinforce.
- **Scope: file tools only, not Bash.** `Write` / `Edit` / `NotebookEdit` first. Bash is harder to parse reliably (redirects, `cp`, `mv`, here-docs); extend only if a recurrence happens through that surface.

## Acceptance

- A test session started in a worktree that attempts `Write` to the canonical repo-root path errors out before the file is written.
- The error message names (a) the attempted path, (b) the current worktree root, (c) the corrected path under the worktree.
- The hook is a no-op when `cwd` is not inside a worktree.
- `python3 -m py_compile claude/scripts/*.py` (or wherever the hook lives) passes.

## References

- brownm09/career-playbook#276 — downstream symptom-tracker
- career-playbook PR #275 — third documented occurrence
- engineering-journal `sessions/career-playbook/2026-05-22_140307.stub.md`

brownm09:
## Summary

- Adds `pre-tool-use-worktree-path-check.py` — a blocking `PreToolUse` hook that fires when a session running inside a Claude-managed worktree issues a `Write`, `Edit`, or `NotebookEdit` call whose absolute path targets the canonical repo root instead of the active worktree root.
- Wired in `settings.json` with three separate matchers (`Write`, `Edit`, `NotebookEdit`).
- Adds ADR-024 recording the block-not-rewrite and Bash-deferred judgment calls.
- Updates `README.md` and `docs/REFERENCE.md` hooks tables.

## What the hook does

1. If `cwd` does not match `.../.claude/worktrees/<name>`, exits 0 (no-op).
2. Extracts `canonical_root` and `worktree_root` from `cwd`.
3. Reads `file_path` (Write/Edit) or `notebook_path` (NotebookEdit).
4. If the path is absolute and starts with `canonical_root` but **not** `worktree_root` → exits 2, blocking the tool call with a message naming the attempted path, the worktree root, and the corrected path.
5. Otherwise exits 0.

## Testing

- `python3 -m py_compile claude/scripts/*.py` → **All OK**
- Manual acceptance: a session in a worktree that attempts `Write` to the canonical repo-root path will now receive a blocking error before the file is written, naming (a) the attempted path, (b) the current worktree root, (c) the corrected path.

Coverage gate: the hook has no automated test suite. This is consistent with all other hook scripts in `claude/scripts/` — they rely on the syntax check and manual verification. Tracked at #238.

Closes #238

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Custom slash commands loaded from `claude/skills/`. Invoke with `/skill-name [ar
 ## Hooks
 
 Hook scripts run automatically via Claude Code's `hooks` configuration in `claude/settings.json`.
-All hooks are advisory — none block tool execution.
+Most hooks are advisory — they emit reminders but do not block tool execution. The exception is `pre-tool-use-worktree-path-check.py`, which blocks `Write`, `Edit`, and `NotebookEdit` calls that target the canonical repo root instead of the active worktree.
 
 | Event | Script | Purpose |
 |---|---|---|
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ All hooks are advisory — none block tool execution.
 | UserPromptSubmit | `reconcile-open-prs.py` | Removes stale entries from `open-prs.jsonl` whose PRs are now merged/closed; emits surviving open PRs as session context |
 | PreToolUse (Bash) | `pre-commit-branch-check.py` | Emits current branch as a checkpoint before `git commit` |
 | PreToolUse (Bash) | `pre-pr-create-check.py` | Emits test-verification checklist and documentation-gap warning before `gh pr create` |
+| PreToolUse (Write/Edit/NotebookEdit) | `pre-tool-use-worktree-path-check.py` | Blocks file writes whose absolute path targets the canonical repo root instead of the active worktree |
 | PostToolUse (Bash) | `pr-merge-reminder.py` | Reminds to write a journal stub after `gh pr create`, `gh pr merge`, or `git push` (when the pushed branch has an open PR) |
 | PostToolUse (Bash) | `post-tool-use.py` | Auto-adds issues/PRs to configured GitHub Project; exits 2 with `required_fields` reminders |
 | PostToolUse (Bash) | `post-pr-merge-pull.py` | Fast-forwards local `main` after `gh pr merge` |
--- a/claude/scripts/pre-tool-use-worktree-path-check.py
+++ b/claude/scripts/pre-tool-use-worktree-path-check.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+"""Claude Code PreToolUse hook — blocks Write/Edit/NotebookEdit calls whose
+file_path targets the canonical repo root when the session is running inside a
+Claude-managed worktree.
+
+Problem: Absolute paths like `C:/Users/brown/Git/dev-env/foo.py` resolve to
+the main working tree, not the active worktree. Files land in the wrong place
+silently. This hook intercepts those calls before the write happens.
+
+Logic:
+  1. If cwd does not contain `/.claude/worktrees/<name>/`, pass immediately.
+  2. Extract canonical_root (repo root) and worktree_root from cwd.
+  3. Read file_path (Write/Edit) or notebook_path (NotebookEdit) from tool input.
+  4. If the path is absolute and starts with canonical_root but NOT with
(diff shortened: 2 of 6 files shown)


brownm09:
## Review: feat(hooks): block Write/Edit/NotebookEdit to canonical repo root from worktrees

### Summary

Needs minor revision before merging. The hook logic, ADR, and settings wiring are all correct; the blocking mechanism is sound. Two `README.md` and `REFERENCE.md` introductory statements flatly contradict the new hook's behavior and must be updated in this PR. No correctness, security, or reliability concerns.

---

### Blocking Findings

**[documentation]** `README.md:43`

The sentence "All hooks are advisory — none block tool execution." was not updated. The new hook added in this same PR exits 2 with a reason — which is an explicit block. The table row two lines below even says "Blocks file writes," making line 43 self-contradictory within the same section.

**Fix:** Change the sentence to something like: "Most hooks are advisory — they emit reminders but do not block tool execution. The exception is `pre-tool-use-worktree-path-check.py`, which blocks `Write`, `Edit`, and `NotebookEdit` calls that target the wrong file tree." (Exact wording is your call; the point is to remove the blanket "none block" claim.)

---

**[documentation]** `docs/REFERENCE.md:115`

The existing text reads: "All hooks are **advisory** — they emit `systemMessage` reminders via exit 2 but do not block tool execution. No hook exits with a non-zero code that prevents the matched tool from running." Both sentences are now false: the new hook exits 2 specifically to prevent the matched tool from running, and it does not emit a `systemMessage` — it prints a blocking `{"reason": "..."}` JSON payload.

**Fix:** Update this sentence to distinguish advisory hooks (exit 0 or exit 2 for `systemMessage`) from blocking hooks (exit 2 with `{"reason": ...}` that prevents tool execution). This also aligns with the authoring-rules language at line 230 that already describes blocking PreToolUse hooks as a known pattern.

---

### Questions for Author

- **`PR_BODY` — "Tracked at #238":** The coverage gate deferral says the lack of automated tests is "tracked at #238," but #238 is the issue being closed by this PR. Is the deferral permanently accepted (consistent with all other hook scripts, no new issue needed), or should a follow-up issue be opened? If permanent, the sentence should say something like "consistent with all other hook scripts — no automation planned" rather than implying a tracking issue exists.

---

### Style Notes

The module-level docstring is unusually long for a hook script (this repo's hooks are generally undocumented or minimally commented). Not a problem — the extra context is genuinely useful for this specific script given the non-obvious `cwd`-regex-then-path-comparison structure. Fine as-is.

---

PR #245, opened 2026-05-23
feat: usage-snapshot post-merge hook for daily limit awareness

### LINKED ISSUE #244: Add usage-snapshot hook for post-merge journal integration
## Problem

Claude Code displays session and weekly usage as a percentage of limits, but there's no way to track daily consumption against a personal soft budget or surface optimization hints in the engineering journal.

## Goal

After each PR merge, automatically capture:
- Weekly utilization vs. a day-of-week threshold (20% weekends / 12% weekdays)
- 5-hour window utilization
- Top 5 exchanges by token cost from the current session JSONL

Include this data in the post-merge journal stub so `/journal-compose` can surface it in the Token Usage section.

## Approach

- New `claude/usage-config.json` — user-editable threshold config
- New `claude/scripts/usage-snapshot.py` — queries `https://api.anthropic.com/api/oauth/usage`, parses session JSONL, emits a markdown block
- `claude/hooks/post-tool-use.py` — add detection of `gh pr merge` → invoke the script, output to session via exit code 2
- `claude/CLAUDE.md` — add stub convention: include usage snapshot block post-merge

brownm09:
## Summary

- Adds `claude/scripts/usage-snapshot.py` — a global `PostToolUse` hook that fires on every `gh pr merge`. Queries `https://api.anthropic.com/api/oauth/usage` (via the OAuth Bearer token in `~/.claude/.credentials.json`) and parses the session JSONL for top-5 exchanges by token count. Emits a `### Usage Snapshot (post-merge)` markdown block with weekly/5-hour utilisation vs. a day-of-week soft target.
- Adds `claude/usage-config.json` — user-editable thresholds (`weekday_pct: 12`, `weekend_pct: 20`, `alert_approaching_margin: 5`).
- Updates `claude/settings.json` to add the new hook to the PostToolUse Bash chain.
- Updates `claude/CLAUDE.md` with the stub-inclusion convention (include the emitted block verbatim in the post-merge stub body).
- Updates `README.md` and `docs/REFERENCE.md`.

## Test plan

- [x] `python3 -m py_compile claude/scripts/*.py` — all clean
- [x] Ran `usage-snapshot.py` directly with a simulated `gh pr merge` hook payload; confirmed weekly utilisation (13%), 5-hour (77%), day-1-of-7 target (≤20% ✅), extra usage, and top-5 exchange table all render correctly
- [x] Worktree-path fallback tested — script strips `/.claude/worktrees/<name>` suffix to find the canonical project JSONL directory
- [x] Verified non-merge commands produce exit 0 (silent)

Closes #244


brownm09:
## Review: feat: usage-snapshot post-merge hook for daily limit awareness

### Summary

Merge-ready with minor fixes. The hook is well-structured, error-safe by design (every non-happy path exits 0 silently), and the live API call plus worktree-path fallback were verified end-to-end. No blocking correctness or security issues. Seven non-blocking findings cover a docstring inaccuracy, a redundant import, a subtle already-expired-token UX bug, an mtime fallback that could silently attribute exchanges to the wrong session, and a few maintainability nits. The most important fix before the next merge is the already-expired-token message (finding 4) since it would actively mislead. The rest are improvements worth applying opportunistically.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/usage-snapshot.py:23`
The module docstring exit-code table says `Exit 0 — ... token expired, or API error; silent` but the token-expiry path in `main()` prints a warning to stderr and calls `sys.exit(2)`.
This is the first thing a future maintainer reads to understand the hook's contract; a wrong exit-code table will mislead anyone tracing why a merge surfaced a warning.
**Suggestion:** Update the table to: `Exit 0 — not a merge command, token missing/unparseable, or API error (silent)` / `Exit 2 — snapshot emitted, OR token expires within 1 hour (warning emitted)`.

---

**[style]** `claude/scripts/usage-snapshot.py:29`
`import os` is present but `os` is never referenced in the script.
**Suggestion:** Remove the line. If `os.environ` is added later for a `CLAUDE_HOOK_DEBUG` escape (see finding 7), restore it then.

---

**[maintainability]** `claude/scripts/usage-snapshot.py:187`
The `except` tuple in `fetch_usage` is `(urllib.error.URLError, urllib.error.HTTPError, json.JSONDecodeError)`. `HTTPError` is a subclass of `URLError`, so it is caught by `URLError` first and the `HTTPError` entry is unreachable.
Future readers may assume the author intended differentiated handling and add `HTTPError`-specific logic that silently never fires.
**Suggestion:** Drop `urllib.error.HTTPError` from the tuple. If differentiated handling is ever needed, add a separate `except urllib.error.HTTPError` block *before* the `URLError` one.

---

**[reliability]** `claude/scripts/usage-snapshot.py:454`
When `expires_at_ms` is already in the past, `expires_at_ms - time.time() * 1000` is a large negative number, which is still `< 3_600_000`, so the "expires within 1 hour" warning fires even for a token that expired hours ago.
A user who merges after their token has expired sees a misleading message ("within 1 hour") rather than the correct diagnosis ("already expired — re-authenticate").
**Suggestion:** Split the condition: check `expires_at_ms <= time.time() * 1000` first and exit 0 silently (token is already dead; the API call would fail anyway), then check the 1-hour window and emit the advisory.

---

**[reliability]** `claude/scripts/usage-snapshot.py:253–266`
In `_find_in_dir`, when `{session_id}.jsonl` does not exist in the project dir, the function falls back to the most-recently-modified JSONL in that directory. If two Claude sessions write to the same project dir concurrently, or if the session JSONL hasn't been flushed yet at merge time, this picks a sibling session's file and presents its exchanges as "Top exchanges this session."
The exchanges table would silently attribute data from a different session, undermining the feature's stated purpose with no indication to the user.
**Suggestion:** Drop the mtime fallback from Tiers 1 and 2 entirely — let Tier 3 (exact `session_id` search across all project dirs) handle the miss. If Tier 3 also misses, the `*(No session JSONL found — exchange breakdown unavailable)*` fallback is the correct output.

---

**[correctness]** `claude/scripts/usage-snapshot.py:241`
`datetime.fromisoformat()` on Python 3.10 (the current runtime, confirmed) does not accept ISO strings ending in `Z` (e.g., `"2026-05-30T04:00:00Z"`). The Anthropic API currently returns `+00:00`-suffixed timestamps (the live test passed), but this is undocumented behaviour of an unofficial endpoint and could change without notice. If it does, the `try/except` in `compute_cumulative_target` swallows the `ValueError` and returns `(0, 0, 7)`, causing the snapshot to display `day 0/7 — target ≤0%` with no error.
**Suggestion:** Normalize before parsing: `resets_at_str.replace("Z", "+00:00")`. Apply in both `compute_cumulative_target` and the `resets_display` formatting block in `format_snapshot` (line ~389).

---

**[reliability]** `claude/scripts/usage-snapshot.py:462`
`fetch_usage` is a synchronous HTTPS call with a 10-second timeout. It runs on the critical path after every `gh pr merge`. If `api.anthropic.com` is slow or the user is briefly offline, the merge feedback stalls for up to 10 seconds with no indication that the wait is from a hook.
**Suggestion:** Lower the timeout to 4–5 seconds (the data is advisory; a slightly stale read is acceptable), or print a one-line `[usage-snapshot] fetching…` to stderr before the call so the delay is attributable. Lowering the timeout is the lower-effort fix.

---

### Style Notes

The `sorted(..., reverse=True)[0]` idiom in `_find_in_dir` builds and sorts a full list just to pick the newest file. Replace with `max(project_dir.glob("*.jsonl"), key=lambda f: f.stat().st_mtime, default=None)` for O(n) behaviour. Minor, but the glob runs under a lock-free OS call on every merge and projects dirs can accumulate many JSONL files over time.

The pervasive `except Exception: pass` / `except Exception: sys.exit(0)` pattern is deliberately correct for a hook that must never block, but it makes silent failures completely invisible during development. A follow-up worth considering: check `os.environ.get("CLAUDE_HOOK_DEBUG")` in the top-level guard and re-raise (or print traceback to stderr + exit 0) when set, so diagnostics are available without modifying the script.

---

*Reviewed by claude-opus-4-5 via Claude Code*

PR #246, opened 2026-05-23
Repair incomplete worktree node_modules

### LINKED ISSUE #242: Worktree npm install: missing ESM build + turbo stub when postinstall skipped
## Symptoms

When adding a new package to a workspace in a git worktree and running `npm install` in the worktree for the first time, two failures can occur:

### 1. `@clerk/shared` postinstall fails — `std-env/dist/index.mjs` not found

```
npm ERR! path ...\.claude\worktrees\admiring-payne-d7bab6\node_modules\@clerk\shared
npm ERR! command failed
npm ERR! command ... node ./scripts/postinstall.mjs
npm ERR! Error [ERR_MODULE_NOT_FOUND]: Cannot find module '...\node_modules\std-env\dist\index.mjs'
npm ERR! Did you mean to import std-env/dist/index.cjs?
```

The worktree's root `node_modules/std-env` (v3.10.0) only contains `dist/index.cjs` and `dist/index.d.cts` — the ESM build (`dist/index.mjs`) is absent. The main repo's `node_modules/std-env` has all files. Both show version 3.10.0 in `package.json`, but the worktree copy is an incomplete install.

**Workaround used:** copied the missing files from the main repo's complete installation:
```bash
cp C:/Users/brown/Git/<repo>/node_modules/std-env/dist/index.mjs    C:/Users/brown/Git/<repo>/.claude/worktrees/<name>/node_modules/std-env/dist/
cp C:/Users/brown/Git/<repo>/node_modules/std-env/dist/index.d.mts    C:/Users/brown/Git/<repo>/.claude/worktrees/<name>/node_modules/std-env/dist/
cp C:/Users/brown/Git/<repo>/node_modules/std-env/dist/index.d.ts    C:/Users/brown/Git/<repo>/.claude/worktrees/<name>/node_modules/std-env/dist/
```
After this, `npm install` (without `--ignore-scripts`) succeeds.

### 2. `@turbo/windows-64/bin/turbo.exe` is a 9 KB stub instead of the real ~38 MB binary

The worktree's `node_modules/@turbo/windows-64/bin/turbo.exe` is a tiny placeholder (downloaded by turbo's own postinstall). When postinstall is skipped (via `--ignore-scripts`) or fails, the real binary is never fetched. The pre-commit hook runs turbo for lint and fails with:

```
Error: spawnSync ...turbo.exe UNKNOWN
errno: -4094, code: 'UNKNOWN', syscall: 'spawnSync ...'
```

**Note:** lifting-logbook PR #323 fixed `.husky/pre-commit` to resolve `turbo.exe` from the **main repo root** when running in a `.claude/worktrees/` path — so commits from worktrees now use the main repo's working binary. That fix addresses the commit-time symptom but not the underlying install-time problem.

**Workaround used (for the install-time case):** copied the real binary from the main repo:
```bash
cp C:/Users/brown/Git/<repo>/node_modules/@turbo/windows-64/bin/turbo.exe    C:/Users/brown/Git/<repo>/.claude/worktrees/<name>/node_modules/@turbo/windows-64/bin/turbo.exe
```

## Root cause

Git worktrees created by Claude Code have their own isolated `node_modules` tree. When a prior `npm install` ran with `--ignore-scripts` (to work around a failing postinstall), it left key files unbuilt/undownloaded. Subsequent plain `npm install` runs don't re-trigger postinstalls for already-installed packages, so the worktree's `node_modules` stays incomplete.

## Proposed fix

One or more of:
- **New script** `claude/scripts/fix-worktree-node-modules.sh` that detects and repairs the two known missing artifacts (std-env ESM build, turbo binary) by copying from the main repo root.
- **CLAUDE.md guidance** — add a platform constraint entry describing the symptom, the `--ignore-scripts` trigger, and the copy-from-main-repo recovery pattern.
- **Investigate root cause** of why the worktree's `node_modules/std-env` is missing the ESM build when the main repo's copy of the same version is complete. May be a timing/ordering issue in how npm workspaces hoist deps into a new worktree.

## Context

Discovered during lifting-logbook PR #321 (Clerk web auth) — adding `@clerk/nextjs` to `apps/web/package.json` triggered `@clerk/shared`'s postinstall, which exposed the incomplete `std-env`.

Chris79OG:
## Summary
- Repair known incomplete Claude worktree `node_modules` artifacts from the main checkout when `node_modules` already exists.
- Restore missing `std-env` ESM/type files and replace a tiny Windows Turbo stub with the real main-checkout binary when available.
- Keep the hook advisory and non-blocking; no credential or external state access.

## Verification
- `python -m py_compile claude\scripts\worktree-npm-install.py`
- Simulated `.claude/worktrees/<name>` repair case for missing `std-env` files and a 9 KB `@turbo/windows-64` stub.

Fixes #242

### DIFF
--- a/claude/scripts/worktree-npm-install.py
+++ b/claude/scripts/worktree-npm-install.py
@@ -21,9 +21,70 @@ Stdin JSON shape (UserPromptSubmit):
 Exit 0 always — advisory only, never blocks.
 """
 import json
+import shutil
 import subprocess
 import sys
 from pathlib import Path
+
+
+def emit(message: str) -> None:
+    print(json.dumps({"systemMessage": message}))
+    sys.stdout.flush()
+
+
+def main_checkout_for_worktree(cwd_path: Path) -> Path | None:
+    parts = cwd_path.parts
+    try:
+        claude_idx = parts.index(".claude")
+    except ValueError:
+        return None
+    if claude_idx + 1 >= len(parts) or parts[claude_idx + 1] != "worktrees":
+        return None
+    return Path(*parts[:claude_idx])
+
+
+def copy_if_missing(main_root: Path, cwd_path: Path, relative: str) -> bool:
+    source = main_root / "node_modules" / relative
+    target = cwd_path / "node_modules" / relative
+    if target.exists() or not source.exists():
+        return False
+    target.parent.mkdir(parents=True, exist_ok=True)
+    shutil.copy2(source, target)
+    return True
+
+
+def repair_known_incomplete_worktree_modules(cwd_path: Path) -> list[str]:
+    """Repair known postinstall-skip artifacts from the main checkout when possible."""
+    if not (cwd_path / "node_modules").exists():
+        return []
+
+    main_root = main_checkout_for_worktree(cwd_path)
+    if not main_root or main_root == cwd_path or not (main_root / "node_modules").exists():
+        return []
+
+    repaired: list[str] = []
+    for relative in (
+        "std-env/dist/index.mjs",
+        "std-env/dist/index.d.mts",
+        "std-env/dist/index.d.ts",
+    ):
+        if copy_if_missing(main_root, cwd_path, relative):
+            repaired.append(relative)
+
+    turbo_relative = "@turbo/windows-64/bin/turbo.exe"
+    source_turbo = main_root / "node_modules" / turbo_relative
+    target_turbo = cwd_path / "node_modules" / turbo_relative
+    if source_turbo.exists() and target_turbo.exists():
+        try:
+            source_size = source_turbo.stat().st_size
+            target_size = target_turbo.stat().st_size
+        except OSError:
+            source_size = target_size = 0
+        if source_size > 1_000_000 and 0 < target_size < 1_000_000:
+            shutil.copy2(source_turbo, target_turbo)
+            repaired.append(turbo_relative)
+
+    return repaired
 
 
 def main() -> None:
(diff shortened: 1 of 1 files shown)


brownm09:
Heads-up: [PR #1071](https://github.com/brownm09/dev-env/pull/1071) (dev-env#970, ADR-142) modifies the same file — `claude/scripts/worktree-npm-install.py` — adding a `node_modules`-truncation gate. **This PR was deliberately left open rather than superseded or absorbed**, because it fixes a different layer:

- **#1071** detects **package-level** absence (the package directory exists but has no `package.json` of its own) and repairs with `npm ci` through ADR-045's free-space gate.
- **This PR** repairs **intra-package file** absence (`std-env/dist/index.mjs` missing while the package's own `package.json` is present) by copying from the main checkout.

ADR-142's detector classifies the `std-env` case as `ok`, so it does not cover what this does.

Two things to know before rebasing:

1. **This branch will conflict.** Its base predates ADR-045, which rewrote the file's `main()` (adding `_gate_install()` and the reclamation ladder), and #1071 has since extracted the install-and-report block into a shared `_run_install()` helper. The `_emit()` refactor in this diff is now already upstream in a different form.
2. **`copy_if_missing()` copies without comparing versions.** It guards on target-absent + source-present, but not on the two checkouts agreeing about which version of the package is installed. If the worktree's branch moved `package-lock.json`, the copied `dist/index.mjs` can come from a different version than the worktree's own `std-env`. The turbo branch does compare sizes, but that only catches the stub case, not a version skew. Worth addressing in the rebase.

No action needed on this PR from #1071's side — filing this so the relationship is on the record rather than rediscovered at rebase time.

PR #248, opened 2026-05-24
fix: prune routine syncs its own worktree instead of primary checkout

### LINKED ISSUE #247: prune-stale-worktrees routine aborts when primary dev-env checkout is on a dirty feature branch
## Problem

The `prune-stale-worktrees` routine passes `REPO=C:/Users/brown/Git/dev-env` (the primary checkout) to `sync-routine-worktree`. If the primary checkout is on a feature branch with uncommitted changes, `git rebase origin/main` fails on the dirty tree and the entire routine aborts without pruning any worktrees.

Incident: routine aborted because primary checkout was on `fix/revert-opusplan-model` with `claude/settings.json` modified and `claude/scripts/__pycache__/` untracked.

## Root cause

The hardcoded primary checkout path is the wrong target. The routine already runs inside a `claude/*` worktree. The sync skill handles `claude/*` branches with a safe `--hard reset origin/main`, not a rebase — so there is no dirty-tree problem.

## Fix

Change `REPO` from the hardcoded primary checkout path to the current worktree's git root: `$(git rev-parse --show-toplevel)`. Update the script invocation path to match.

brownm09:
## Summary

- `prune-stale-worktrees` was passing `REPO=C:/Users/brown/Git/dev-env` (the primary checkout) to `sync-routine-worktree`
- If the primary checkout was on a feature branch with uncommitted changes, `git rebase origin/main` failed and the routine aborted without scanning any worktrees
- Fix: use `$(git rev-parse --show-toplevel)` so the routine syncs its own `claude/*` worktree — the sync skill handles `claude/*` branches with `--hard reset`, which never fails on dirty state

Closes #247

## Test plan

- [ ] Docs-only change to SKILL.md — no Python scripts modified; `python3 -m py_compile claude/scripts/*.py` passes (confirmed)
- [ ] Next scheduled prune run will proceed even when primary dev-env checkout is on a dirty feature branch

### DIFF
--- a/claude/routines/prune-stale-worktrees/SKILL.md
+++ b/claude/routines/prune-stale-worktrees/SKILL.md
@@ -9,12 +9,15 @@ Prune stale Claude session worktrees across all git repos under `C:/Users/brown/
 **Objective:** For every git repo directly under `C:/Users/brown/Git`, remove all `claude/*` worktrees whose branches are fully merged into `origin/main` and have no uncommitted changes. Also remove any non-primary worktrees accidentally checked out on `main`. Report the pruned/skipped summary per repo and a combined total.
 
 **Steps:**
-0. Sync the dev-env working tree to `origin/main`:
-   - Invoke `sync-routine-worktree` with `REPO=C:/Users/brown/Git/dev-env`, `VERIFY_FILE=claude/scripts/prune-merged-worktrees.py`, `PREFIX=prune-stale-worktrees`.
+0. Determine the current worktree's git root, then sync it to `origin/main`:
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: fix: prune routine syncs its own worktree instead of primary checkout

### Summary

Ready to merge. The one-line root cause (hardcoded primary-checkout path passed to `sync-routine-worktree`) and both affected references in the SKILL.md are correctly updated. No blocking findings.

---

*No Blocking Findings*

*No Non-Blocking Findings*

*No Questions for Author*

---

PR #250, opened 2026-05-24
feat: session-start permission mode prompt

### LINKED ISSUE #249: feat: session-start mode prompt hook
At session start, the permission mode is not reliably 'plan' because sessions spawned from an active bypass-permissions session inherit that mode rather than reading defaultMode from settings.json.

Add a UserPromptSubmit hook that fires on the first prompt of each new session and blocks with a prompt asking the user to confirm or change the permission mode before proceeding. Uses a short-lived marker file to allow the re-submit to pass through without re-blocking.

brownm09:
## Summary

- Adds `session-mode-prompt.py` hook wired as the first `UserPromptSubmit` handler
- On the first prompt of each new session (no prior assistant turns), exits 2 to block and shows a mode-selection reminder (plan / bypass / auto) with Shift+Tab instructions
- Uses a 2-minute marker file in `scratch/` so the user's re-submit after adjusting the mode passes through without re-blocking
- Updates `README.md` hooks table and `docs/REFERENCE.md` script descriptions

Addresses the Claude Code platform behavior where sessions spawned from an active bypass-permissions window inherit that mode rather than reading `defaultMode: "plan"` from `settings.json` (documented in ADR-025).

## Test plan

- [ ] `python3 -m py_compile claude/scripts/session-mode-prompt.py` passes
- [ ] New session: first prompt is blocked with the mode-selection message
- [ ] Re-submit after seeing the message: passes through without blocking
- [ ] Third prompt onward (established session): no blocking

## Testing

```
python3 -m py_compile claude/scripts/*.py
```

All scripts compile without syntax errors.

Closes #249


brownm09:
## Review: feat: session-start permission mode prompt

### Summary
Ready to merge with minor fixes. The hook design is sound — marker-write-before-exit-2 ordering correctly ensures the re-submit passes through, the worktree suppression mirrors the established pattern in `new-day-journal-check.py` and `multi-worktree-alert.py`, and first-prompt detection is robust. Two non-blocking issues: the `REFERENCE.md` entry omits the worktree suppression behavior (inconsistent with sibling entries), and a silent write failure on the marker file leaves no diagnostic path if the scratch directory is unavailable.

---

### Non-Blocking Findings

**[documentation]** `docs/REFERENCE.md` — worktree suppression not mentioned
The `session-mode-prompt.py` entry describes the marker-file passthrough and platform-limitation context but does not mention that the hook is suppressed in Claude-managed worktree sessions. Other hooks with the same suppression (e.g., `new-day-journal-check.py`, `multi-worktree-alert.py`) include "Suppressed in Claude-managed worktree sessions (`.claude/worktrees/` in cwd)." in their entries. A contributor reading only the reference doc has no way to know routine sessions are exempt.
**Suggestion:** Append that phrase to the `session-mode-prompt.py` row in the `docs/REFERENCE.md` table.

**[reliability]** `claude/scripts/session-mode-prompt.py:47` — silent marker-write failure
The `except: pass` block swallows any `IOError` when writing the marker. If the scratch directory is absent or has a permissions issue, the marker is never created. On re-submit the hook finds no marker, sees no prior assistant turns, and blocks again — the user would be stuck re-submitting indefinitely (the session never advances to an assistant turn). The scratch directory is always present in normal operation, but the silent failure leaves no path for diagnosing the rare case where it is not.
**Suggestion:** Replace `pass` with `sys.stderr.write(f"session-mode-prompt: could not write marker: {e}\n")` so the hook still exits 2 (blocking) but the error surfaces in hook output rather than disappearing silently.

---

PR #252, opened 2026-05-24
[config] Add suppression policy and ADR-026: no silent workarounds

### LINKED ISSUE #251: [config] Add suppression policy: no silent workarounds for TypeScript/lint errors
## Problem

Recent fix PRs in lifting-logbook (notably #338) added TypeScript suppressions (`!`, `?? null`) to silence compiler errors rather than fixing the underlying code. The current CLAUDE.md and testing rules do not distinguish between tests passing because the code is correct versus tests passing because suppressions were added. Suppressions accumulate and normalize over time.

## Solution

Add a `## Code Quality` section to `claude/CLAUDE.md` with three enforced rules:

1. **No suppression without justification** — every suppression in a PR requires a PR-body note naming the lines and stating the invariant
2. **Pre-existing errors must be filed, not silenced** — check whether the error predates the branch; if so, file/batch a GitHub issue and leave the error unmodified
3. **Pre-PR suppression check** — run a `git diff | grep` check before every `gh pr create` to surface new suppressions

Also write ADR-026 documenting the decision.

## Linked issue

lifting-logbook#343 is the first batch of suppressions to be cleaned up under this policy.

brownm09:
## Summary

- Adds a `## Code Quality` section to `claude/CLAUDE.md` with three binding rules for TypeScript/lint suppressions
- Appends a suppression-check cross-reference to the existing "Test before PR" bullet in `## Git Workflow`
- Writes ADR-026 documenting the decision and rationale
- Updates `docs/adr/INDEX.md` with the new ADR row

## Rules introduced

**Rule 1 — No suppression without justification.** Every suppression in a PR must have a PR-body note naming the lines and the invariant relied on.

**Rule 2 — Pre-existing errors must be filed, not silenced.** If an error predates the current branch, file a GitHub issue (batch if multiple) and leave the error unmodified.

**Rule 3 — Pre-PR suppression grep.** Run before every \`gh pr create\`:
\`\`\`bash
git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|!\s*[;,)]|\?\? null)'
\`\`\`

## Suppression check results for this PR

The grep matched two lines in \`claude/CLAUDE.md\` itself:
1. The definition sentence (contains the literal strings as examples of what counts as a suppression)
2. The grep command example in Rule 3

Both are documentation — no actual suppressions are introduced. This is a known false positive from pattern-matching a doc that defines the patterns.

## Testing

- \`python3 -m py_compile claude/scripts/*.py\` — **Syntax OK**
- Suppression grep diff reviewed — two matches are documentation lines, not suppressions (see above)
- ADR-026 appears in \`docs/adr/INDEX.md\`

## Linked

- Closes #251
- lifting-logbook#343 is the first batch of suppressions to clean up under this policy

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -262,8 +262,10 @@ If the error exists on the base branch, do **not** suppress it. File a GitHub is
 **Rule 3 — Pre-PR suppression check (required before `gh pr create`).**
 Run this from the repo root and review every match before opening a PR:
 ```bash
-git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|!\s*[;,)]|\?\? null)'
+git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|![.[]|!\s*[;,)]|\?\? null)'
 ```
+The pattern covers end-of-expression assertions (`!;`, `!,`, `!)`) and chained access (`!.`, `![`). It does not catch every `!` — also scan added lines in the diff manually for any remaining `!` that could be a suppression.
+
 If the output is non-empty:
 - Each line must map to a Rule 1 justification note in the PR body, **or**
 - The suppression must be removed and replaced with a proper fix.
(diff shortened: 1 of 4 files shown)


brownm09:
## Review: [config] Add suppression policy and ADR-026: no silent workarounds

### Summary

Ready to merge with one minor fix. The policy, ADR, and INDEX update are well-structured, internally consistent, and the PR body correctly accounts for the false-positive from documentation lines matching the grep pattern. One non-blocking finding: the grep pattern `!\s*[;,)]` has false negatives for chained non-null assertions (`foo!.bar`, `arr![0]`, bare `return foo!`) — a developer could run the check, get no output, and incorrectly believe no suppressions are present.

---

### Non-Blocking Findings

**[correctness]** `claude/CLAUDE.md` (Rule 3 grep) / `docs/adr/026-suppression-policy.md` (Consequences)

The pattern `!\s*[;,)]` in Rule 3's grep only catches non-null assertions immediately followed by `;`, `,`, or `)`. It misses: chained property access (`arr[i]!.bar`), chained indexing (`arr[i]![0]`), and bare end-of-expression without terminal punctuation (e.g., `return arr[i]!` on its own line). All three are valid TypeScript and all silently pass the grep with zero matches.

This matters because the rule says "run this grep and review every match" — a developer seeing no output has reasonable grounds to think they're clear. The false negative is silent. The ADR's Consequences section does not document this limitation.

**Suggestion:** Extend the pattern to cover chained access:
```bash
git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|![.[]|!\s*[;,)]|\?\? null)'
```
(`![.[]` catches `!.` and `![`.) Alternatively, add a sentence to Rule 3 and the ADR Consequences noting that the pattern covers end-of-expression assertions but chained assertions (`!.` and `![`) require a manual scan of the diff.

---

PR #256, opened 2026-05-24
fix: warn when canonical dev-env worktree is on a feature branch

### LINKED ISSUE #255: dev-env-sync: warn when canonical worktree is on a feature branch
## Problem

When the canonical dev-env worktree (`~/Git/dev-env`) is on a feature branch,
`dev-env-sync.py` exits silently. Because `~/.claude/settings.json` and
`~/.claude/scripts/` are symlinked/junctioned to that worktree, any files
merged to `origin/main` after the branch was cut are invisible — hooks and
scripts stay stale until the worktree returns to main.

This was observed with PR #250 (`session-mode-prompt.py`): the script and
settings.json update never reached `~/.claude/` because the canonical worktree
was on `config/suppress-policy-adr-026`.

## Fix

Replace the silent `sys.exit(0)` on the feature-branch path with a one-line
advisory:

```
⚠️  dev-env canonical worktree is on '<branch>' — ~/.claude/ symlinks may be
stale. Switch to main or use a separate worktree for dev-env changes.
```

Exit 0 (advisory, never blocking). Fires on every prompt until the worktree
returns to main, which is the intended signal.

brownm09:
## Summary

- Replaces the silent `sys.exit(0)` on the feature-branch path in `dev-env-sync.py` with a one-line advisory printed to stderr
- Message fires on every prompt until the canonical worktree returns to `main`, which is the intended forcing function
- No change to exit code (always 0 — advisory only, never blocking)

## Why

When `~/Git/dev-env` is on a branch other than `main`, `~/.claude/settings.json` and `~/.claude/scripts/` (both junctioned/symlinked to the canonical worktree) serve that branch's file state. Newly merged hooks and scripts are invisible. This caused `session-mode-prompt.py` (PR #250) to not fire after merge because the canonical worktree was still on `config/suppress-policy-adr-026`.

## Test plan

- [x] `python3 -m py_compile claude/scripts/*.py` passes
- [ ] With canonical worktree on a feature branch: next prompt shows `[dev-env-sync] ⚠️  Canonical worktree is on '...'` warning
- [ ] With canonical worktree on `main`: no change in behavior

Closes #255


brownm09:
## Review: fix: warn when canonical dev-env worktree is on a feature branch

### Summary
Ready to merge with minor changes — two documentation inaccuracies must be fixed first, both in existing descriptions of `dev-env-sync.py`. The logic itself is correct, the advisory pattern (stderr, exit 0) matches the established hook convention, and the CLAUDE.md rule addition is clear and well-placed.

---

### Blocking Findings

**[documentation]** `docs/REFERENCE.md` — dev-env-sync.py description row
The description reads: *"Fast-forward pulls the dev-env repo to `origin/main` so symlinked tooling stays current. Skips if a feature branch is active in the main worktree."*
The second sentence is now false — the script no longer skips silently when on a feature branch; it emits an advisory warning to stderr. Any session that orients from REFERENCE.md to understand hook behavior will see a description that contradicts what the hook actually does.
**Fix:** Update the description to reflect the new behavior, e.g.: *"Fast-forward pulls the dev-env repo to `origin/main` so symlinked tooling stays current. Warns (stderr, exit 0) when the canonical worktree is on a feature branch — symlinked tooling will serve stale files until it returns to main."*

---

**[documentation]** `claude/scripts/dev-env-sync.py:9-11` — module docstring
The docstring says: *"Only acts when the repo has `main` checked out — exits silently when a feature branch is active so dev-env PR sessions are not spammed with false-positive divergence warnings."*
Both the behavior description and the rationale are now wrong. The script no longer exits silently on a feature branch; it emits a warning. And the reason has flipped — the point is now to surface the staleness, not to suppress false positives.
**Fix:** Replace lines 9–11 with something like: *"Warns on every prompt when the canonical worktree is on a feature branch, since `~/.claude/` symlinks will serve that branch's stale files. Skips the fast-forward pull in that case — only syncs when `main` is checked out."*

---

### Style Notes

The new branch warning uses the `⚠️` emoji prefix (`[dev-env-sync] ⚠️  Canonical worktree…`) while the two existing warnings in the same file use a plain-text `WARNING:` prefix (`[dev-env-sync] WARNING: local dev-env repo has diverged…` and `[dev-env-sync] WARNING: fast-forward pull failed`). Not a blocker, but worth standardizing in a follow-up if the visual distinction between advisory and error warnings is intentional — if not, dropping the emoji in favor of the `WARNING:` prefix would be consistent.

---

*Reviewed by `claude-opus-4-5` via Claude Code*

PR #262, opened 2026-05-27
fix: restore hook execution via py -3 (closes #261)

### LINKED ISSUE #261: session-mode-prompt.py is silently failing to surface its banner
## Problem

The `session-mode-prompt.py` UserPromptSubmit hook is intended to block the first prompt of every human session and print a mode-confirmation banner. After PRs #258 (Unicode fix) and #260 (worktree CWD check replaced with XML-tag detection), the hook should fire for all human prompts in worktree-based user sessions.

Empirical observation in a real session today:
- No banner appeared at session start.
- The marker file at \`C:/Users/brown/.claude/scratch/session_mode_ack.txt\` was absent — indicating the hook never reached the marker-write step.
- The user's first prompt did not begin with \`<\`, so the XML suppression regex should not have matched.

## Why this matters

Without the forcing function, sessions inherited from a parent in \`bypass-permissions\` mode (the original motivation behind ADR-025) can silently skip the plan-mode default. The user falls back to typing the mode manually each session — exactly the friction the hook was created to eliminate.

## Scope

Diagnose the root cause. Add structured debug logging to the hook so we have evidence to act on, then ship a fix in a follow-up PR.

Likely candidates:
- Stdout-vs-stderr issue on exit 2 under Git Bash / Windows.
- Claude Code suppressing exit-2 output when the resolved \`defaultMode\` already matches the inherited mode.
- A residual Windows encoding issue beyond #258.

Tracked in plan: \`C:/Users/brown/.claude/plans/linked-mapping-deer.md\` (Phase 1).

brownm09:
## Summary

Restore Claude Code hook execution on Windows. Every hook in `settings.json` had been silently failing because `python3` in the hook runner's PATH resolves only to the Microsoft Store App Execution Alias stub at `WindowsApps/python3.exe`, which exits 49 without running. Discovery: while debugging why `session-mode-prompt.py`'s banner never appeared, found the scratch directory held **no** artifacts from **any** hook (no `ctx-warn-*` from turn-count-hook, no `session_mode_ack.txt`, etc.).

ADR-007 (2026-04-27) already predicted this failure mode would surface someday. This PR is that day.

## Changes

- `claude/settings.json`: replace `python3 ...` with `py -3 ...` in all 18 hook commands. `py.exe` (Windows Python Launcher) is installed to `AppData/Local/Programs/Python/Launcher/` which is on the Windows system PATH (verified). `py -3` picks the latest installed Python 3 so it survives Python upgrades, unlike a hardcoded absolute path.
- `claude/scripts/session-mode-prompt.py`: append a JSON line per invocation to `scratch/session-mode-prompt.log` so future silent failures can be diagnosed without re-instrumenting. Wrap banner print in try/except.
- `docs/adr/007-hook-command-invocation.md`: amend with the 2026-05-26 resolution and the rationale for choosing `py -3` over the originally-prescribed absolute path.
- `CLAUDE.md`, `claude/CLAUDE.md`, `docs/adr/019-doc-reconciliation-enforcement.md`: switch the `## Testing` and verification commands from `python3 -m py_compile ...` to `py -3 -m py_compile ...` (the existing references were broken on this machine too).
- `claude/hooks/pre-push`, `setup.sh`, `claude/setup-prompt.md`, `claude/skills/journal-compose/SKILL.md`, `docs/REFERENCE.md`: same `python3` → `py -3` substitution where user-facing or runtime-invoked.
- `docs/adr/INDEX.md`: update ADR-007 title and date.

## Testing

```
py -3 -m py_compile claude/scripts/*.py
```

All scripts compile clean.

Manual smoke tests in this PR's worktree:
- Fresh marker absent + human prompt → banner printed, exit 2, log records `banner_printed`.
- Fresh marker absent + XML-tagged prompt (`<scheduled-task>...`) → no banner, exit 0, log records `automated_suppressed`.
- Marker present and < 120s old + any prompt → no banner, exit 0, log records `cooldown_passthrough`.

## Coverage gate

The settings.json switch + the script logging are the testable behavior changes. They are exercised by the manual smoke tests above. Automated hook-output testing would require a new harness (no existing test framework for hook stdout/exit-code assertions in dev-env) — out of scope for this PR. Filed as a follow-up consideration during the original plan write-up.

## Suppression check

`git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|![.[]|!\s*[;,)]|\?\? null)'` — empty.

Closes #261


brownm09:
## Review: fix: restore hook execution via py -3 (closes #261)

### Summary

Merge with one small fix. The core change — `python3` → `py -3` across all 18 hook commands in `settings.json` plus the test commands in CLAUDE.md, REFERENCE.md, and ADR-019 — is correct, well-scoped, and clearly justified. The amended ADR-007 captures the rationale (Python Launcher beats absolute path because it survives Python upgrades) and the discovery path. The `session-mode-prompt.py` logging instrumentation is well-structured (`_log` never raises, three exit branches each tagged, banner write wrapped). One residual `python3` reference is user-facing instructional output and will mislead users on this machine — that one is blocking. A few non-blocking cleanups for consistency.

---

### Blocking Findings

**[correctness]** `claude/scripts/new-day-journal-check.py:255`
The journal hook prints a remediation message instructing the user to run `python3 ~/.claude/scripts/reconcile-late-stubs.py draft/<branch>`. On this machine `python3` resolves to the Microsoft Store App Execution Alias stub and exits 49 — the exact failure mode this PR exists to eliminate. Any user who follows the instruction after a resurrected-draft event will hit a silent failure and lose the recovery path.
**Fix:** Change line 255 to `f"Run: py -3 ~/.claude/scripts/reconcile-late-stubs.py draft/{resurrected[0]}"`.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/reconcile-late-stubs.py:10-11`, `claude/scripts/backfill-tokens.py:9`, `claude/scripts/token-report.py:5-11`
Module-level usage docstrings still document invocation as `python3 <script>.py ...`. Anyone reading the script header or copy-pasting the usage line will hit the same Microsoft Store stub. These are not as user-facing as the `new-day-journal-check.py` message (they require opening the script), but they are now drift against the ADR-007 invariant.
**Suggestion:** Replace `python3` with `py -3` in each `Usage:` docstring. Shebangs (`#!/usr/bin/env python3`) can stay — they are interpreted by the OS only when scripts are invoked directly, which never happens in this codebase (hooks always go through `py -3 script.py`).

**[maintainability]** `claude/settings.json:22`
`"Bash(python3 -m py_compile *)"` allow rule is retained alongside the new `"Bash(py -3 -m py_compile *)"` (line 23). The `python3` form is now dead — no test command in this repo uses it after this PR.
**Suggestion:** Remove the stale `python3 -m py_compile` allow rule for hygiene. Harmless to keep, but the codebase is moving away from `python3` and the allowlist should follow.

**[documentation]** `docs/adr/024-worktree-path-guard-hook.md:62`
ADR-024 cites `python3` as a valid `Bash` invocation example for deliberately bypassing the worktree-path-guard hook. After this PR, that example is misleading on Windows.
**Suggestion:** Change `python3` to `py -3` in the line "use `Bash` with `node -e` or a targeted `sed`/`python3` invocation". Trivial doc consistency fix; safe to defer to a follow-up sweep.

**[documentation]** Hook-output JSONL may capture sensitive prompt content
`session-mode-prompt.py` now writes `prompt[:80]` to `scratch/session-mode-prompt.log` on every prompt. The scratch directory is local-only and never committed, so blast radius is contained, but a user pasting an API key or password as their first prompt would land it on disk in cleartext.
**Suggestion:** Acceptable for the Phase 1 diagnostic window. When the log is gated behind `CLAUDE_HOOK_DEBUG=1` per the original plan, consider either dropping `prompt_prefix` entirely (the other fields are usually sufficient to diagnose) or truncating to ~20 chars.

---

### Style Notes

The session-mode-prompt.py rewrite is clean — the `event` dict accretion pattern keeps the three exit paths visually parallel and the log entries shape-consistent across branches. The amended ADR-007 is structurally good (Original Context → 2026-04-27 fix → 2026-05-26 amendment → updated Decision/Consequences), which makes the evolution traceable.

---

PR #265, opened 2026-05-27
fix: scope session-mode-prompt marker per session-id

### LINKED ISSUE #263: session-mode-prompt: shared marker file suppresses banner for new sessions started within cooldown window
The cooldown marker (session_mode_ack.txt) is a single global file. When one session shows the banner and writes the marker, any new session that starts within 120 seconds gets a cooldown passthrough — even though that session has never seen the banner. The fix is to make the marker per-session-id (session_mode_ack_{session_id}.txt).

brownm09:
## Summary

- The cooldown marker was a single global file (`session_mode_ack.txt`), so any session opening within 120 seconds of another session's banner display got a free pass — even though it had never seen the banner itself.
- Fix: use a per-session marker file (`session_mode_ack_{session_id}.txt`) so the cooldown is scoped to the re-submit window within a single session only.
- Added pruning of expired per-session markers on each invocation to keep `scratch/` clean.

## Root cause (from log)

```
01:07:10  session 1f02952d  banner printed   (marker written)
01:08:19  session e379b24a  cooldown passthrough  marker_age=68s  <-- missed banner
```

## Test plan

- [ ] `py -3 -m py_compile claude/scripts/*.py` — passes
- [ ] Open a new session within 120 seconds of this one; confirm the banner appears in the new session

## Suppression check

No suppressions added.

Closes #263

### DIFF
--- a/claude/scripts/session-mode-prompt.py
+++ b/claude/scripts/session-mode-prompt.py
@@ -25,7 +25,8 @@ import sys
 import time
 import traceback
 
-MARKER = "C:/Users/brown/.claude/scratch/session_mode_ack.txt"
+MARKER_DIR = "C:/Users/brown/.claude/scratch"
+MARKER_PREFIX = "session_mode_ack_"
 LOG_PATH = "C:/Users/brown/.claude/scratch/session-mode-prompt.log"
 COOLDOWN_SECS = 120  # covers the re-submit window after user sees the prompt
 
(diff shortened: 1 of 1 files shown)


brownm09:
Superseded by #266, which:

- Includes the per-session marker fix from this PR (same approach, slightly different naming — `MARKER_DIR` constant + a `_marker_path()` helper).
- Also fixes a second latent defect this PR missed: the banner was being written to **stdout**, but per [Claude Code's `UserPromptSubmit` hook contract](https://docs.claude.com/en/docs/claude-code/hooks#hook-input-output), exit-2 surfaces stderr to the user; stdout is fed back to the model as added context. The banner has likely been invisible to humans since the hook was added — that's why we never saw it before #262 made hooks executable.
- Adds [ADR-027](https://github.com/brownm09/dev-env/blob/fix/session-mode-prompt-stderr-and-per-session-marker/docs/adr/027-userpromptsubmit-blocking-hook-conventions.md) codifying both invariants (stderr-for-blocking, per-session marker pattern) and updates `docs/REFERENCE.md` to match.
- Drops the 120s cooldown as redundant — per-session marker existence alone is the correct signal and covers the user's re-submit case without time-bounding.

The stale-marker pruning logic added here was a nice touch but not strictly needed: orphaned markers are ~18 bytes each and harmless. Per ADR-027, a periodic sweep can be added if disk usage ever becomes a concern.

Closing without merge.

PR #266, opened 2026-05-27
fix: session-mode-prompt banner to stderr + per-session marker (closes #264)

brownm09:
## Summary

Two latent defects in `session-mode-prompt.py`, masked for over a month by the `python3`-stub hook-execution failure documented in ADR-007. PR #262 restored hook execution; the diagnostic log added in the same PR exposed both defects within the first three fresh sessions.

### Bug 1 — Banner written to stdout, invisible to the user

Per [Claude Code's `UserPromptSubmit` hook contract](https://docs.claude.com/en/docs/claude-code/hooks#hook-input-output), exit-2 surfaces **stderr** to the user; **stdout** is fed back to the model as added context. The original `print(...)` (and the `sys.stdout.write(...)` version that replaced it in #262) wrote the banner to stdout, so the human never saw it — Claude received it as input. The banner has likely been invisible to humans since the hook was added.

**Fix:** `sys.stderr.write(banner)` instead of `sys.stdout.write(banner)`.

### Bug 2 — Cross-session marker contamination

A single global file at `scratch/session_mode_ack.txt` was used as the cooldown marker. Observed from `scratch/session-mode-prompt.log` after #262 landed:

| Time | Session | Prompt | Result |
|---|---|---|---|
| 01:07:10 | A | `/review 352` | banner_printed, marker written |
| 01:08:19 | B (fresh) | `Test` | cooldown_passthrough — read A's 68s-old marker |
| 01:08:41 | B | `Isn't there supposed to be a banner...` | cooldown_passthrough |
| 01:09:13 | C (fresh) | `Test 2` | banner_printed (marker finally aged past 120s) |

**Fix:** per-session marker at `scratch/session_mode_ack_<session_id>.txt`. `session_id` is sanitized via `[^A-Za-z0-9_-]` before use as a filename. The 120-second cooldown is no longer needed — marker existence alone (for the current session) is the right signal and covers the user's re-submit case. `COOLDOWN_SECS` dropped.

## Changes

- `claude/scripts/session-mode-prompt.py`: rewritten with both fixes. Log event now includes `marker_path` so future contamination would be obvious. New log stage `session_acked_passthrough` replaces `cooldown_passthrough` to reflect the new semantics.
- `docs/adr/027-userpromptsubmit-blocking-hook-conventions.md`: new ADR codifying both invariants — stderr-for-blocking and per-session marker pattern — with the Anthropic hook docs as the primary-source citation for the exit-2 stream contract.
[...]

PR #273, opened 2026-05-27
fix: session-mode-prompt to exit-0 + additionalContext (closes #268)

### LINKED ISSUE #268: session-mode-prompt: design mismatch — exit-2 erases the prompt; redesign or remove
## Background

PRs #260, #262, #266 progressively fixed bugs in `claude/scripts/session-mode-prompt.py`:
- #260: XML-tag detection for automated sessions
- #262: hook execution restored (python3 → py -3) — exposed downstream defects
- #266: per-session marker + banner to stderr — addresses cross-session contamination and stdout-discard

After #266 merged and the canonical worktree pulled, two fresh sessions were opened to verify. Result: hook fired (logged `banner_printed`), but the user never saw a banner; one session reported "the worktree gets created, then Claude stops".

## Root cause

The hook design is misaligned with the documented `UserPromptSubmit` exit-2 contract. Per [Claude Code hook docs](https://code.claude.com/docs/en/hooks):

> *"Exit 2 means a blocking error. Claude Code ignores stdout and any JSON in it. Instead, stderr text is fed back to Claude as an error message."*

For `UserPromptSubmit` specifically:

> *"Blocks prompt processing and erases the prompt."*

So on exit 2:
1. The user's prompt is **deleted** from the conversation
2. stderr is forwarded to **Claude (the model)** as an error message — NOT to the user UI
3. stdout is **ignored entirely**

What we want is "show a one-time mode-confirmation reminder to the user". `UserPromptSubmit` exit-2 is for "reject this prompt with a reason" (e.g., refusing prompts containing secrets). It is fundamentally the wrong mechanism for a reminder banner. Every implementation so far has been chasing the wrong target.

Observed failure mode in the 2026-05-27 verification:
- Session A typed "Test 3"
- Hook fired, banner to stderr, exit 2 → prompt erased
- Claude (model) received the banner as an error with no prompt to act on
- Claude created a worktree (dev-env default behavior on new session) and stopped — no useful response possible

## Options

**A. Drop the hook.** `defaultMode: "plan"` in `settings.json` already starts fresh sessions in plan mode (the real safety property). `Shift+Tab` cycles modes. The banner adds friction (prompt deleted, must re-submit) without adding value. Recommended unless someone can articulate a concrete value the banner adds beyond `defaultMode`.

**B. Switch to `SessionStart` hook + stdout context injection.** Fires once when a session starts, output goes into Claude's startup context. Banner becomes advisory text Claude sees; Claude surfaces a mode-confirmation message in its first response without any prompt erasure.

**C. Keep `UserPromptSubmit`, switch to exit 0 + `additionalContext` JSON.** Return `{\"hookSpecificOutput\": {\"hookEventName\": \"UserPromptSubmit\", \"additionalContext\": \"<reminder>\"}}` and exit 0. Claude sees the reminder alongside the prompt, prompt preserved, no blocking.

## Acceptance criteria

- Fresh session opened, user types any prompt: the prompt is NOT erased; the user gets a useful response that includes the mode-selection reminder (or no reminder at all if option A); user does not have to re-submit.
- Per-session marker (or equivalent) prevents the reminder from re-appearing on every prompt in the same session.
- ADR-027 amended (or, if option A, deprecated with a note) to reflect the corrected understanding.

## Related

- #266 (merged) — the immediate prior fix that exposed this design issue.
- ADR-027 — the architectural record that needs updating once a design is chosen.
- Plan file `C:/Users/brown/.claude/plans/linked-mapping-deer.md` Phase 2 (opusplan model nudge) is downstream of this — it depends on a working reminder mechanism, so resolve this first.

brownm09:
## Summary
- Switches `session-mode-prompt.py` from `exit 2 + stderr` (which erases the prompt per the documented `UserPromptSubmit` contract) to `exit 0 + stdout` JSON `hookSpecificOutput.additionalContext`. The mode-confirmation reminder now travels alongside the user's prompt; the prompt is preserved, the user does not re-submit, and Claude weaves a mode-confirmation line into its first response.
- Amends [ADR-027](docs/adr/027-userpromptsubmit-blocking-hook-conventions.md) with a new rule: **choose the hook contract by intent**. Hooks that genuinely block a prompt still use `stderr + exit 2`; hooks that inject advisory context use `stdout JSON + exit 0`. The original stderr-on-exit-2 rule stands — the amendment adds the second case, not a replacement.
- Per-session marker, XML-tag automated-session suppression, and diagnostic logging are unchanged.

Closes #268.

## Background
After #266 shipped (banner → stderr + per-session marker), the user opened two fresh sessions and the banner was still invisible while the prompt was being erased. Re-reading the [Claude Code Hooks documentation](https://code.claude.com/docs/en/hooks) clarified the real defect: `UserPromptSubmit` exit 2 deletes the prompt and forwards stderr to Claude as a blocking error message — fundamentally the wrong mechanism for a one-time advisory. The chosen redesign (Option C from #268) uses `hookSpecificOutput.additionalContext`, which is the documented exit-0 channel for delivering text to Claude alongside an un-erased prompt.

## Files changed
- `claude/scripts/session-mode-prompt.py` — rewrite output contract; rename log stage `banner_printed` → `additional_context_emitted`; update docstring.
- `docs/adr/027-userpromptsubmit-blocking-hook-conventions.md` — append Amendment 2026-05-27 (issue #268).
- `README.md` — updated row in Hooks table.
- `docs/REFERENCE.md` — updated row in UserPromptSubmit hooks table.

## Tested
- `py -3 -m py_compile claude/scripts/*.py` → OK (project's documented test command from `## Testing` in `CLAUDE.md`).
- Local hook simulation with three synthetic payloads:
  1. Fresh `session_id` + plain prompt → emits the `hookSpecificOutput.additionalContext` JSON to stdout, exit 0.
  2. Same `session_id` re-submitted → silent passthrough, exit 0 (per-session marker hit).
  3. Different `session_id` with `<scheduled-task>` XML prompt → silent passthrough, exit 0 (automated-session suppression).
- End-to-end verification will happen on the next fresh canonical session after this PR merges and `dev-env-sync` pulls the new script.

## Coverage gate
This is a hook-contract behavior change; the project has no automated tests for hook scripts (per ADR-007 the tests are syntax-compile + manual simulation). Local simulation covers the three relevant paths above. Deferring automated test creation — there is no existing harness for hook-payload simulation in this repo and adding one is out of scope for this fix.

## Suppression check
`git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|!\[.[]|!\s*[;,)]|\?\? null)'` → no matches.

## ADR-warrant check
This change amends an existing ADR (ADR-027) in the same PR; no new ADR is needed.

## Doc-reconciliation check
README.md and docs/REFERENCE.md updated in this PR (the two reference docs required by the Documentation Maintenance table in `CLAUDE.md` for `claude/scripts/` changes).

## Test plan
- [x] Syntax check passes
- [x] Hook simulation: fresh session emits JSON
- [x] Hook simulation: re-submit passes through silently
- [x] Hook simulation: XML automated prompt passes through silently
- [ ] Post-merge: open a fresh canonical session, type any prompt, confirm prompt is preserved and the response includes a mode reminder; check `scratch/session-mode-prompt.log` for `additional_context_emitted` then `session_acked_passthrough` on the next prompt.


brownm09:
## Review: fix: session-mode-prompt to exit-0 + additionalContext (closes #268)

### Summary
Merge-ready with minor considerations. The change correctly applies the documented `UserPromptSubmit` exit-0 + `hookSpecificOutput.additionalContext` contract per the [Claude Code Hooks documentation](https://code.claude.com/docs/en/hooks), preserves the user's prompt, and amends ADR-027 with a clear intent-based rule rather than retracting prior decisions. Reference docs (README.md, docs/REFERENCE.md) are updated in the same PR, and the manual-simulation deferral for automated tests is explicitly justified. Highest-severity concern is non-blocking: the `additionalContext` text is phrased as a directive to Claude, which will produce a visible mode-confirmation preamble on the very first response of every session — that is the intentional design, but worth flagging in case the friction outweighs the safety benefit in practice.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/session-mode-prompt.py:50-55`
`_REMINDER_TEXT` instructs Claude to "briefly confirm which permission mode is active (plan / bypass / auto) and remind them they can press Shift+Tab to cycle modes" in its first response.
Because this fires on the **first prompt of every session**, every session's first response will now carry a 1–3 sentence mode-confirmation preamble — including sessions where the user typed something trivial like "hi" or `/clear`. The prior banner was at least dismissable noise; the new design embeds the reminder into Claude's output, which is harder to suppress on a per-session basis. If the user finds the preamble friction-inducing, they will have to either edit this string or remove the hook entirely.
**Suggestion:** Consider softening the directive to "if the user appears to be starting a substantive task" or similar conditional, so trivial first prompts don't force a preamble. Defer if you want to evaluate the actual friction first — easy to soften later.

**[maintainability]** `claude/scripts/session-mode-prompt.py:154-160`
The `additional_context_emit_failed` branch writes the failure message to stderr after `sys.stdout.write(json.dumps(...))` has already raised. On `UserPromptSubmit` exit 0, stderr is surfaced to the user as a visible info/warning line.
Same is true for the existing `marker_write_error` path at line 134. Both paths are rare (disk write or stdout failure), but if they fire, the user sees `session-mode-prompt: ...` text appended to their UI on the first prompt of a session. This is acceptable diagnostic behavior, but worth knowing — the hook is no longer fully silent on its error paths.
**Suggestion:** Acceptable as-is. If you want zero user-visible noise on hook errors, route the stderr writes through `_log()` only and drop the user-facing prints — the log file already captures the same data.

**[correctness]** `claude/scripts/session-mode-prompt.py:140-146`
The marker is written **before** the JSON payload is emitted. If `sys.stdout.write` raises (e.g., broken pipe), the marker has already been written, so the reminder is silently lost for that session — the user never gets the reminder, and the next prompt in the same session passes through silently.
The probability is extremely low (stdout write from a hook process is essentially guaranteed under normal Claude Code execution), and the diagnostic log captures the failure, so a missed reminder is observable rather than invisible. The pre-write order is intentional per the comment ("so any retry on the same session passes through") — flipping it would let a transient retry double-emit, which is worse.
**Suggestion:** No change needed; the tradeoff is correct. Worth a one-line comment noting that emit-failure deliberately consumes the reminder slot rather than risk double-emission.

---

### Style Notes
The docstring rewrite is clear and accurately describes the new contract; the inline comment updates (e.g., "Write marker BEFORE emitting so any retry on the same session passes through") match the new semantics. ADR-027's amendment section is well-structured — preserving the original body and adding intent-based guidance keeps the historical context intact while documenting the new rule. No style nits.

---

PR #283, opened 2026-05-27
feat: baseline test failures + fix-on-touch policy

### LINKED ISSUE #282: Baseline test failures + fix-on-touch policy
## Problem

Pre-existing test failures silently ride along across many branches in lifting-logbook (and any project with non-trivial test debt). Each session sees them, correctly declines to fix unrelated breakage, and the backlog grows invisibly. Today's gates (test-before-PR, test-integrity policy) catch newly *introduced* failures only — they have no concept of inherited red state.

## Goal

Make pre-existing failures **visible** (baseline snapshot at branch-cut), **actionable** (fix inline when cheap, file an issue when not), and **measurable** (end-of-session report) — without turning every small PR into a yak-shave on unrelated test debt.

## Approach

Hybrid baseline + fix-on-touch (full spec in plan file at \`~/.claude/plans/the-suggested-fix-on-touch-proxy-recursive-tome.md\`):

1. **\`new-branch\`** auto-snapshots failing tests to \`scratch/baseline_<repo>_<branch>.json\` when \`baseline_test_failure_tracking: true\` in \`.claude/hook-config.json\`.
2. **\`baseline-tests diff\`** pre-PR groups failures into: \`new\` (block), \`preexisting-touched\` (fix-or-file), \`preexisting-untouched\` (note only).
3. **Fix-on-touch rule**: inline fix iff failure is in a file the branch already modifies AND fix is ≤ ~20 LOC / ~15 min. Otherwise file/append a rolling tracking issue.
4. **CLAUDE.md** gets a new "Pre-existing test failure policy" subsection paralleling suppression / test-integrity policies.
5. **ADR-030** records the decision and the rejected alternatives (baseline-only / fix-on-touch-only / block-all-on-red).

## Out of scope (deferred)

- Pytest / Go / Rust adapters (Jest first)
- Automatic rolling-issue management
- CI-side enforcement
- Per-project enablement of lifting-logbook (separate follow-up PR there)

## Files touched

- NEW: \`claude/scripts/baseline-tests.sh\`, \`docs/adr/030-baseline-test-failure-policy.md\`
- EXTEND: \`claude/scripts/new-branch.sh\`, \`claude/scripts/pre-pr-create-check.py\`, \`claude/CLAUDE.md\`, \`docs/adr/INDEX.md\`, \`docs/REFERENCE.md\`, \`README.md\`

brownm09:
## Summary

Adds a per-branch baseline of failing tests so pre-existing red state stops silently riding along across sessions. Completes the test-policy trio: ADR-022 (missing tests), ADR-029 (degraded tests), ADR-030 (inherited red state).

- New `claude/scripts/baseline-tests.sh` with `snapshot` and `diff` subcommands. Jest `--json` adapter; fingerprint = `sha1(file + "::" + test_name + "::" + first_line_of_error)`. Snapshot path: `C:/Users/brown/.claude/scratch/baseline_<repo>_<branch>.json`.
- `new-branch.sh` runs the snapshot automatically when `baseline_test_failure_tracking: true` is set in the repo's `.claude/hook-config.json` (opt-in per project).
- `pre-pr-create-check.py` adds an advisory line pointing at `baseline-tests diff` when a baseline file exists for the current branch, or warning when the opt-in flag is on but no baseline was captured.
- `claude/CLAUDE.md` gets a new **Pre-existing test failure policy** subsection under `## Code Quality`, structured as three rules parallel to the suppression and test-integrity policies. Cross-reference added to the Test-before-PR bullet in Git Workflow.
- ADR-030 records the decision, rejected alternatives (baseline-only / fix-on-touch-only / block-all-on-red / full-stack fingerprint), and the explicitly judgment-based fix-on-touch threshold (~20 LOC / ~15 min).
- README + REFERENCE updates: new script row, two new `hook-config.json` fields documented, hook description amended.

Jest-only in this first implementation — CLAUDE.md scope note flags this so a clean run in a non-Jest repo is not misread as policy compliance. Pytest/Go/Rust adapters and automatic tracking-issue management are deferred.

## Testing

dev-env's `## Testing` command is a Python syntax check over hook scripts (no automated test suite for shell/markdown).

\`\`\`
$ py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py && bash -n claude/scripts/baseline-tests.sh claude/scripts/new-branch.sh
PY SYNTAX OK
BASH SYNTAX OK
\`\`\`

Tests: 0 passed, 0 skipped, 0 failed (n/a — dev-env has no automated test suite beyond syntax-check).

Additional manual verification done in-session:
- Hook smoke test (no hook-config → emits original 3-line checklist, no baseline line; exit 0).
- Hook smoke test (opt-in, no baseline file → emits ⚠ BASELINE warning at position 4; exit 0).
- Hook smoke test (opt-in, baseline file present → emits "run baseline-tests diff" advisory at position 4; exit 0).
- `baseline-tests --help` and unknown-subcommand (exit 2) behave as documented.

End-to-end snapshot/diff against a real Jest project is deferred until the follow-up lifting-logbook PR turns the flag on there; the parsing logic is exercised inline in the node `-e` blocks and verified via the help/error paths.

### Suppression check
\`\`\`
$ git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|![.[]|!\s*[;,)]|\?\? null)'
(no matches)
\`\`\`

### Test-integrity check
\`\`\`
$ git diff origin/main -- . | grep -E '(it\.skip|xit\(|xdescribe\(|test\.skip|describe\.skip|\.todo\(|pending\(|passWithNoTests|--bail|testPathIgnorePatterns)'
(no matches — feature is shell + python + markdown only)
\`\`\`

### Test coverage gate

This PR adds testable behavior (the `baseline-tests` script and the new hook branch). Deferred: dev-env has no automated test harness for shell scripts or for hook stdin/stdout contracts. Manual smoke tests above stand in. The follow-up enablement PR in lifting-logbook will exercise the snapshot/diff end-to-end against a live Jest suite, which is the first practical opportunity to verify the parser against a real project.

Closes #282


brownm09:
## Review: feat: baseline test failures + fix-on-touch policy

### Summary

Merge with minor changes. The three-rule structure (CLAUDE.md + helper script + advisory hook) mirrors ADR-026/029 cleanly and ADR-030 records the decision and rejected alternatives well. No blocking issues. The helper script has a few small robustness/heuristic concerns worth tightening before this gets exercised against a live Jest suite in the follow-up lifting-logbook PR — they're all in `baseline-tests.sh` and concentrated in the diff-classification path. Doc-reconciliation and test-integrity gates pass.

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/baseline-tests.sh:253-254`
The touched-file matcher falls through to a basename comparison: `if (base && t.split('/').pop() === base && t.includes(base))`. The `t.includes(base)` clause is dead — if `t.split('/').pop() === base`, then `t` contains `base` by definition. More importantly, pure basename matching produces false positives: a touched `src/foo.test.ts` and an untouched failing `lib/foo.test.ts` both share the basename, so the latter is classified `preexisting-touched` (forcing fix-on-touch or filing). Since this is the heuristic that decides whether Rule 2 fires, false positives translate into unnecessary yak-shave work.
**Suggestion:** Drop the redundant `includes` check, and tighten the basename branch — only accept basename match when the failure's `file` is reported as an absolute path (no `/` in `fileRel` after normalisation, or path doesn't resolve as repo-relative). Otherwise require a full suffix match.

**[reliability]** `claude/scripts/baseline-tests.sh:197, 222`
Test runs use `eval "$tc" > "$tmp" 2>/dev/null || true`. When the test command fails to *start* (binary missing, wrong cwd, turbo wrapper swallows `--json`), stderr is discarded and the user sees only the generic `"Test command produced no stdout"` message. That's the exact failure mode the `test_command` override field is meant to address, so the diagnostic matters most when this branch fires.
**Suggestion:** Redirect stderr to a sibling tmp file (`2> "$tmp.err"`) and, on the empty-stdout branch, echo its tail (`tail -n 5 "$tmp.err"`) before exiting 2. Clean up the err file on success.

**[reliability]** `claude/scripts/baseline-tests.sh:138-178, 237-279`
The `node -e` blocks embed `$repo`, `$branch`, `$raw`, `$out`, `$baseline`, `$current`, `$touched_file` via shell interpolation into a single-quoted-looking-but-actually-double-quoted JS source. Branch names are sanitized only for `/` → `-`; values containing `'`, `\`, or `$` would break the JS source (or worse). In this workflow the inputs are operator-controlled so the practical risk is low, but the script will look unsafe to any reader and is fragile if branch names ever evolve (e.g., a UTF-8 branch with an apostrophe).
**Suggestion:** Pass the values as environment variables instead of interpolating: `repo="$repo" branch="$branch" raw="$raw" out="$out" node -e '... process.env.repo ... fs.readFileSync(process.env.raw,...) ...'`. Single-quote the JS body so shell doesn't expand anything inside it.

**[reliability]** `claude/scripts/baseline-tests.sh:182`
`touched_files()` greps `git diff --name-only origin/main`. If the worktree's `origin/main` ref hasn't been fetched recently (common in long-running worktrees), the touched set is stale and a file modified on the branch can be missed → its failure gets bucketed `preexisting-untouched` instead of `-touched`, silently skipping Rule 2.
**Suggestion:** `git fetch origin main` (or `git fetch --quiet origin`) at the top of `cmd_diff` before computing the touched set. The snapshot path doesn't need this — it captures whatever state exists at branch creation.

**[maintainability]** `claude/scripts/baseline-tests.sh:280-282`
`local rc=$?` immediately after the heredoc `node -e "..."` block captures the exit status of `node`, but the cleanup `rm -f` and final `exit $rc` are written as if outside a function. Bash accepts `local` only inside functions — this works because `cmd_diff` is a function, but it's easy to read as top-level code. Worth a one-line comment, or move `rm -f` before the `node -e` block ends so the exit code propagates without the temp var.

**[reliability]** `claude/scripts/new-branch.sh:319-338`
Snapshot runs *synchronously* immediately after `git checkout -b`. ADR-030 explicitly accepts this for lifting-logbook (sub-minute suite). For a repo where the opt-in is enabled but the suite later grows past a few minutes, every `new-branch` call blocks on the full suite with no opt-out short of editing the config. There's no documented escape valve (e.g., `NO_BASELINE=1 new-branch foo`).
**Suggestion:** Honor an env var like `BASELINE_TESTS_SKIP=1` to bypass the snapshot for a single invocation, and mention it in the ADR-030 Consequences section.

---

### Questions for Author

**Question:** Should `baseline-tests diff` fetch `origin/main` before comparing, or is the operator expected to do that manually before running it?
**Context:** `touched_files()` at [claude/scripts/baseline-tests.sh:182](claude/scripts/baseline-tests.sh:182) diffs against `origin/main` without a prior fetch. In a worktree opened hours/days after the last fetch, the touched set will be missing files added to `main` in the meantime — those bucket as `preexisting-untouched` and silently skip Rule 2.
**Tradeoffs:** (a) Fetch in `cmd_diff` — one extra network round-trip per diff, but makes the classification accurate; matches how `pre-pr-create-check.py` and `pre-push` already behave. (b) Document the assumption in CLAUDE.md Rule 3 ("run `git fetch origin` first") — cheaper but adds another ordering rule the operator has to remember. (c) Leave as-is and accept silent miscategorization in stale worktrees — saves the round-trip but undermines the whole point of the diff.

---

### Style Notes

The bash + embedded-node pattern is consistent with the rest of the repo (per CLAUDE.md's "jq is NOT available" rule). The JS-in-double-quoted-heredoc style does make these blocks one of the harder things in the repo to read; if more of them appear over time, consider extracting `claude/scripts/baseline-tests-parse.mjs` and `claude/scripts/baseline-tests-diff.mjs` as standalone node modules invoked with file paths as argv. Not worth doing for two call sites.

---

PR #287, opened 2026-05-28
journal-compose: add Start here dashboard block to top-level README

### LINKED ISSUE #286: journal-compose: add Start here dashboard block to top-level README
## Problem

The top-level `engineering-journal/README.md` is a catalog of 11 projects with per-project `Recent:` / `Open:` / `Next:` bullets. To find "what should I pick up today?" the user has to scan all 11 sections and reconstruct cross-project priorities manually.

## Goal

Add a single-glance landing block at the top of `engineering-journal/README.md` that `/journal-compose` overwrites on every run, containing:

1. `**Last composed:** YYYY-MM-DD` — freshness stamp
2. `## Start here` — top 3–5 priorities across all projects

## Approach

- Extend the manifest schema with an optional `priorities` array (each entry: `label` required, `ref`/`why` optional).
- Add **Step 8a** to `journal-compose/SKILL.md`: aggregate `priorities` from today's manifests + remainder from `open-prs.jsonl` across all projects, dedupe by `ref`, cap at 5.
- Insert/replace the block between literal markers `<!-- start-here:begin -->` / `<!-- start-here:end -->` so future composes are idempotent.
- Document the schema change in `claude/CLAUDE.md` → "Manifest format".
- Companion engineering-journal PR updates `scripts/validate-jsonl.js` to accept the new optional field, and adds a CLAUDE.md note that the block is skill-owned.

## ADR

Warrants an ADR — changes documented skill behavior and the manifest schema.

## Verification

See plan in `~/.claude/plans/is-the-journal-compose-skill-robust-pony.md` → Verification section.

brownm09:
## Summary

- Adds Step 8a to `/journal-compose` that rewrites a marker-delimited `## Start here` block at the top of `engineering-journal/README.md` on every compose
- Block contains a `**Last composed:**` freshness stamp + the top 3–5 cross-project priorities, aggregated from a new optional manifest field plus `open-prs.jsonl` fallback (deduped by `ref`, capped at 5)
- Extends manifest schema with optional `priorities` array (`label` required; `ref`/`why` optional)
- Records the design decision in [ADR-032](docs/adr/032-journal-start-here-dashboard.md); updates `INDEX.md`

Closes #286

## Companion PR

Engineering-journal PR (to land after this merges):
- Extend `scripts/validate-jsonl.js` to accept the optional `priorities` field on manifest lines
- Add a note to `engineering-journal/CLAUDE.md` that the marker-delimited block is skill-owned

## Testing

Ran the project test command from `dev-env/CLAUDE.md`:
```
```
Result: **all hook scripts parse cleanly.** No hook scripts were modified in this PR.

Tests: N/A — no automated test suite covers the skill markdown itself. End-to-end verification is in the plan file (`~/.claude/plans/is-the-journal-compose-skill-robust-pony.md`) and will run on the next real `/journal-compose` invocation.

**Pre-PR suppression check:**
```
git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|\?\? null)'
```
No matches — this PR adds no suppressions.

**Pre-PR test-integrity check:** No test files modified, no skip markers added, no coverage thresholds changed.

**Coverage gate:** The new behavior (Step 8a aggregation + Edit) has no automated test coverage; engineering-journal has no skill test harness. The validator change in the companion PR is the closest enforceable test and will cover the schema half.

## Test plan

- [ ] After merge, manually re-symlink check: `~/.claude/skills/journal-compose/SKILL.md` should reflect the new Step 8a
- [ ] Next real `/journal-compose` run — verify the block appears above `## Projects` in engineering-journal README
- [ ] Re-compose later that week — verify the block is replaced (not appended)
- [ ] Flag a priority in a stub manifest, compose, verify it surfaces

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -516,6 +516,12 @@ echo '{"stub":"YYYY-MM-DD_HHMMSS.stub.md","topic":"<H2 heading>","tokens":{"inpu
 
 - `prs_opened`: PR numbers opened during this session (e.g., `[54]`). Empty array if none.
 - `prs_closed`: PR numbers reviewed/merged during this session (e.g., `[54]`). Empty array if none.
+- `priorities` (optional): array of items to surface on the top-level README "Start here" dashboard.
+  Each entry: `label` (required string, short title); `ref` (optional string, `owner/repo#N` or
+  freeform key used for dedupe); `why` (optional string, one-sentence rationale). Example:
+  `"priorities":[{"label":"Staging gate fix","ref":"lifting-logbook#346","why":"blocks next deploy"}]`.
+  `/journal-compose` aggregates these across all projects (deduped by `ref`, capped at 5) — see
+  [ADR-032](https://github.com/brownm09/dev-env/blob/main/docs/adr/032-journal-start-here-dashboard.md).
 
 The manifest lets `/journal-compose` see the session count, topics, token data, and PR lifecycle
 without reading individual stubs. It is advisory: if the manifest is missing or has fewer entries
--- a/claude/skills/journal-compose/SKILL.md
+++ b/claude/skills/journal-compose/SKILL.md
@@ -779,6 +779,104 @@ Top-level entry format (labeled bullets, no prose block):
 If the project does not yet appear in the README, add a new `### <Project>` section
 under `## Projects` using this format.
 
+## Step 8a — Update top-level "Start here" dashboard block
+
+After Step 8, refresh the marker-delimited block at the top of `engineering-journal/README.md`. This block surfaces a freshness stamp and the top 3–5 cross-project priorities. It is rewritten on every compose; it is not hand-edited between composes. See [ADR-032](https://github.com/brownm09/dev-env/blob/main/docs/adr/032-journal-start-here-dashboard.md) for rationale.
+
+**Aggregate the priority list (max 5 entries, deduped by `ref`):**
+
+```bash
+TMPFILE="C:/Users/brown/.claude/scratch/tmp_start_here_$$.json"
+node -e "
+  const fs = require('fs'); const path = require('path');
+  const root = 'C:/Users/brown/Git/engineering-journal';
+  const date = 'YYYY-MM-DD';   // <-- substitute the compose date
+  const items = [];
+  const seen = new Set();
+  const push = (it) => {
+    const key = (it.ref || '').trim() || ('__no_ref_' + items.length);
+    if (seen.has(key)) return;
+    seen.add(key);
+    items.push(it);
+  };
+  const sessionsDir = path.join(root, 'sessions');
+  const projects = fs.readdirSync(sessionsDir).filter(d =>
+    fs.statSync(path.join(sessionsDir, d)).isDirectory());
+  // Source 1: today's manifests
+  for (const proj of projects) {
+    const f = path.join(sessionsDir, proj, date + '.manifest.jsonl');
+    if (!fs.existsSync(f)) continue;
+    for (const line of fs.readFileSync(f,'utf8').split('\n').filter(Boolean)) {
+      try {
+        const m = JSON.parse(line);
+        for (const p of (m.priorities || [])) {
+          push({ source:'manifest', project:proj, label:p.label, ref:p.ref||'', why:p.why||'' });
+        }
+      } catch (e) {}
+    }
+  }
+  // Source 2: open-prs.jsonl across projects (fills to 5)
+  if (items.length < 5) {
+    for (const proj of projects) {
+      if (items.length >= 5) break;
+      const f = path.join(sessionsDir, proj, 'open-prs.jsonl');
+      if (!fs.existsSync(f)) continue;
+      for (const line of fs.readFileSync(f,'utf8').split('\n').filter(Boolean)) {
+        if (items.length >= 5) break;
+        try {
+          const o = JSON.parse(line);
+          const ref = (o.url||'').replace(/^https:\/\/github.com\//,'').replace(/\/pull\//,'#');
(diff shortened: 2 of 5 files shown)


brownm09:
## Review: journal-compose: add Start here dashboard block to top-level README

### Summary
Scope is appropriate (202 lines across docs, skill, ADR). Approach matches ADR-032 cleanly. **One blocking correctness issue:** the cross-source `ref` dedup key won't actually match between manifest priorities and `open-prs.jsonl` entries as written, so the same PR flagged in both sources will appear twice in the top-5 list — defeating the dedup rule the ADR specifies. Two non-blocking gaps (underspecified `/pull` vs `/issues` rendering, and a procedural gap where the skill writes priorities to a tmpfile but never tells the LLM-executor to read it back to render bullets). REFERENCE.md still describes `/journal-compose` as producing only the 11-section document — non-blocking but worth a one-line update.

---

### Blocking Findings

**[correctness]** `claude/skills/journal-compose/SKILL.md:817-826`
The dedup logic keys on `ref` to merge "same PR flagged in manifest + sitting in open-prs.jsonl" into one item (ADR-032's stated rationale). Manifest entries use short refs like `lifting-logbook#346` (per the schema example in `claude/CLAUDE.md:520` and the ADR's example). Open-prs entries are derived from `o.url` via `replace(/^https:\/\/github.com\//,'').replace(/\/pull\//,'#')`, producing fully-qualified refs like `brownm09/lifting-logbook#346`. The two strings are not equal, so `seen.has(key)` returns false and the same underlying PR renders as two separate bullets.
**Fix:** Normalize both refs to a single canonical form before checking `seen`. Simplest: strip the owner prefix from open-prs refs (`brownm09/lifting-logbook#346` → `lifting-logbook#346`) since the manifest convention is owner-less. Or normalize the other direction by always prepending `brownm09/` to manifest refs that lack an owner. Document the chosen convention in ADR-032 under "Why dedupe by `ref`" so the companion validator enforces it.

---

### Non-Blocking Findings

**[maintainability]** `claude/skills/journal-compose/SKILL.md:837-841`
The aggregation `node -e` block writes per-item data (`label`, `ref`, `why`, `project`, optional `url`) to `$TMPFILE` and prints only `PRIORITY_COUNT=N`. The "Render the block" prose that follows describes the output shape with placeholders but never tells the executor to read `$TMPFILE` back. A future maintainer (or LLM session) reading Step 8a in isolation has to infer the data path.
**Suggestion:** Add one explicit instruction between the aggregation and rendering blocks: "Read `$TMPFILE` (an array of `{source, project, label, ref, why, url?}` objects) and render each entry as a numbered bullet per the rules below." Makes the data flow self-evident.

**[correctness]** `claude/skills/journal-compose/SKILL.md:858`
"If `ref` matches `owner/repo#N`, render as `[#N](https://github.com/owner/repo/<pull|issues>/N)`. Default to `/pull/` unless the manifest entry specifies `/issues/`." The manifest schema in `claude/CLAUDE.md:519-522` defines `ref` as a freeform string with no way to specify pull vs. issue, so the "unless specified" clause is unreachable. An issue ref (`engineering-journal#287`) will always render as a `/pull/` link.
**Suggestion:** Either (a) make `ref` parse a trailing `!issue` / `!pr` hint and document it in the schema, or (b) drop the unreachable clause and acknowledge in ADR-032 that all GitHub refs render as `/pull/` (acceptable since open-prs is the dominant source, and a 302 from `/pull/N` for an issue number is harmless).

**[documentation]** `docs/REFERENCE.md` (not in diff)
The `/journal-compose` entry in `docs/REFERENCE.md` still says it "produces the canonical 11-section document, commits to the `draft/YYYY-MM-DD` branch, and opens a PR." It does not mention the Start here block. This isn't strictly required by the Documentation Maintenance table (no skill add/remove/rename, no invocation syntax change), but the user-visible behavior of `/journal-compose` now includes rewriting a top-of-README block on every run.
**Suggestion:** Add one sentence to the `/journal-compose` entry in REFERENCE.md: "Also refreshes the marker-delimited `## Start here` block at the top of `engineering-journal/README.md` (see [ADR-032](../docs/adr/032-journal-start-here-dashboard.md))." Can land in this PR or the engineering-journal companion.

**[maintainability]** `claude/skills/journal-compose/SKILL.md:868-871`
Step 3 of the insertion logic ("insert the new block immediately above the first `## Projects` heading") has no fallback if `## Projects` is missing. The current `engineering-journal/README.md` does have it, so this is fine today, but a future refactor of that README's structure would silently produce no insertion and the block would never appear. The script also prints no error.
**Suggestion:** Add an explicit failure path: "If neither the markers nor a `## Projects` heading exist, abort Step 8a with `START_HERE_INSERT_FAILED: anchor not found` and continue the compose; do not silently skip." Lets future failures be detected via the symptom-(b) check the ADR already documents.

---

### Questions for Author

**Question:** Is the omission of an `## Projects`-missing fallback intentional (premised on the README structure being stable), or worth hardening now?
**Context:** Step 8a relies on either the marker pair existing or `## Projects` being present as an anchor (`SKILL.md:868-871`). The ADR's "Detection" section names a missing marker pair as symptom (b), but a missing `## Projects` heading would produce the same symptom for a different reason and is not currently distinguished.
**Tradeoffs:** (a) Leave as-is — README structure has been stable for months and adding an abort path is yak-shaving for a hypothetical future. (b) Add the abort + named error — cheap (one line of prose in the skill, one bullet in the ADR) and turns a silent failure mode into a diagnosable one. (c) Anchor the block to a different marker (e.g., always insert directly under the `# Engineering Journal` H1) — removes the dependency on `## Projects` ordering but couples the dashboard to the title structure instead.

---

### Style Notes

ADR-032 and the SKILL.md addition both read cleanly — prose-heavy where it should be, code-heavy where it should be. The Step 8a heading uses "Step 8a" (lowercase `a`) which matches existing convention. No style concerns.

---

PR #289, opened 2026-05-29
Prevent Windows system sleep while Claude is processing

### LINKED ISSUE #288: Prevent system sleep while Claude is processing
## Problem

While Claude Code is working through a long turn (large agent fan-out, multi-file edits, CI polling), the Windows machine can enter sleep, which suspends Claude's process tree and forces a session resume. Manually toggling sleep settings or PowerToys Awake before every long session is friction.

## Goal

Hook-driven sleep prevention:
- `UserPromptSubmit` → block system sleep
- `Stop` / `Notification` → release the block

The machine sleeps only while Claude is idle (waiting on my input or permission), not while it's processing.

## Constraints

- Hook scripts are short-lived child processes — `SetThreadExecutionState` clears on process exit. Need a detached background watcher held alive by a sentinel file.
- Sentinel must auto-expire on a heartbeat timeout (~30 min) so a crashed/orphaned session does not leave the machine permanently awake.
- Idempotent: re-entering UserPromptSubmit while a watcher is already running should be a no-op.
- Windows-only — uses `kernel32!SetThreadExecutionState` via ctypes.

## Out of scope

- Display wake (only block system sleep, not the monitor).
- External CI waiting after `Stop` fires — hooks have no signal for that.

brownm09:
## Summary

Hook-driven sleep prevention for Windows. UserPromptSubmit blocks system sleep; Stop / Notification releases it. Machine sleeps only while Claude is idle (waiting on user input or permission).

- New script `claude/scripts/awake-blocker.py` with `start` / `stop` / `--watcher` modes.
- Registered in `settings.json` on UserPromptSubmit, Stop, and (new block) Notification.
- Detached watcher holds `SetThreadExecutionState(ES_CONTINUOUS \| ES_SYSTEM_REQUIRED)`; sentinel file + 30-min heartbeat timeout for orphan safety.
- ADR-033 records design and alternatives.

Closes #288

## Testing

`py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py` — OK.
`py -3 -c "import json; json.load(open('claude/settings.json'))"` — OK (validated settings.json shape).

No automated tests for hook scripts in this repo. Manual verification deferred to post-merge (the script only takes effect once `~/.claude/settings.json` symlink picks up the merged file, which requires being on `main` in the canonical worktree). Verification plan: open a session, run a long agent fan-out, confirm machine does not sleep; let Stop fire, confirm watcher exits (check `scratch/awake.log`).

Tests: 0 passed, 0 skipped, 0 failed (syntax-only).

### Coverage gate

No automated test infrastructure in dev-env covers hook-script behavior. The watcher's correctness depends on Windows API and OS scheduler behavior that is impractical to fake. Deferring tests is consistent with how other hook scripts in `claude/scripts/` are validated (syntax + manual).

### Suppression / test-integrity / baseline checks

- Suppression policy patterns target TS/JS — not applicable to a pure-Python change. No `# type: ignore`, `# noqa`, or equivalent silencers added.
- No test-integrity violations (no test files modified).
- Baseline-tests not configured for dev-env.

## ADR-warrant + doc-reconciliation

- ADR-033 added; INDEX updated.
- README hooks table + REFERENCE UserPromptSubmit and Stop sections updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Most hooks are advisory — they emit reminders but do not block tool execution.
 | UserPromptSubmit | `turn-count-hook.py` | Warns when session context token count exceeds threshold |
 | UserPromptSubmit | `multi-worktree-alert.py` | Lists active worktrees in `repo:branch` format when ≥2 are open |
 | UserPromptSubmit | `reconcile-open-prs.py` | Removes stale entries from `open-prs.jsonl` whose PRs are now merged/closed; emits surviving open PRs as session context |
+| UserPromptSubmit / Stop / Notification | `awake-blocker.py` | Spawns a detached watcher that holds a Windows system-sleep lock while Claude is processing; releases on Stop or Notification |
 | PreToolUse (Bash) | `pre-commit-branch-check.py` | Emits current branch as a checkpoint before `git commit` |
 | PreToolUse (Bash) | `pre-pr-create-check.py` | Emits test-verification checklist, documentation-gap warning, and pre-existing-failure baseline advisory before `gh pr create` |
 | PreToolUse (Write/Edit/NotebookEdit) | `pre-tool-use-worktree-path-check.py` | Blocks file writes whose absolute path targets the canonical repo root instead of the active worktree |
--- a/claude/scripts/awake-blocker.py
+++ b/claude/scripts/awake-blocker.py
@@ -1,0 +1,186 @@
+"""Block Windows system sleep while Claude is processing.
+
+Invoked from settings.json hooks:
+  - UserPromptSubmit: start (idempotent — spawns watcher if not running, refreshes heartbeat)
+  - Stop / Notification: stop (removes sentinel, watcher exits within 1s)
+
+Mechanism: a detached background watcher process calls
+SetThreadExecutionState(ES_CONTINUOUS | ES_SYSTEM_REQUIRED) in a loop and exits
+when the sentinel file is missing or its heartbeat is older than HEARTBEAT_MAX_AGE.
+Exit clears the execution-state flag automatically.
+
+Safe-exit guard: any exception in hook mode exits 0 so a bug never blocks a prompt.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import subprocess
+import sys
+import time
+from pathlib import Path
+
+SCRATCH = Path("C:/Users/brown/.claude/scratch")
+SENTINEL = SCRATCH / "awake.lock"
+PID_FILE = SCRATCH / "awake.pid"
+LOG_FILE = SCRATCH / "awake.log"
+
+# Heartbeat must be refreshed at least this often. UserPromptSubmit refreshes it on every prompt;
+# if the session crashes or is killed, the watcher self-terminates after this many seconds.
+HEARTBEAT_MAX_AGE = 30 * 60  # 30 minutes
+WATCHER_POLL_INTERVAL = 1.0   # seconds between sentinel checks
+EXEC_STATE_REFRESH = 30.0     # seconds between SetThreadExecutionState calls
+
+# SetThreadExecutionState flags
+ES_CONTINUOUS = 0x80000000
+ES_SYSTEM_REQUIRED = 0x00000001
+
+
+def _log(msg: str) -> None:
+    try:
+        SCRATCH.mkdir(parents=True, exist_ok=True)
+        with LOG_FILE.open("a", encoding="utf-8") as f:
+            f.write(f"{time.strftime('%Y-%m-%dT%H:%M:%S')} pid={os.getpid()} {msg}\n")
+    except Exception:
+        pass
+
+
+def _pid_alive(pid: int) -> bool:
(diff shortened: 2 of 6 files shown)


brownm09:
## Review: Prevent Windows system sleep while Claude is processing

### Summary
Merge with minor changes. The hook + detached-watcher design is sound, ADR-033 is thorough, and doc reconciliation is complete (README, REFERENCE, INDEX). One reliability finding worth addressing pre-merge: `_pid_alive` matches by PID alone, which on Windows is vulnerable to PID reuse and would silently leave a session unprotected. Everything else is non-blocking.

---

### Blocking Findings

**[reliability]** `claude/scripts/awake-blocker.py:55-63`
`_pid_alive(pid)` runs `tasklist /FI "PID eq N"` and returns True if `str(pid)` appears in stdout. Windows recycles PIDs aggressively — a long-lived but later-died watcher writes its PID to `awake.pid`; the OS reuses that PID for an unrelated process (any `py.exe`, `cmd.exe`, IDE child, etc.); the next `start()` reads the stale PID, `tasklist` confirms "some process" exists, and `start()` returns without spawning a fresh watcher. Result: the user gets no sleep-block for the rest of that session and the failure is silent (the log shows no spawn event).
**Fix:** Filter tasklist on image name as well, e.g. `tasklist /FI "PID eq N" /FI "IMAGENAME eq py.exe" /NH`, and require both the PID and `py.exe` to appear in stdout before treating the watcher as alive. (Optionally also record the watcher's start time alongside the PID for stronger identity.)

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/awake-blocker.py:24-26` — Sentinel/PID files are shared across concurrent Claude sessions
`SENTINEL` and `PID_FILE` are fixed paths under `scratch/`. With two parallel sessions, session A's `Stop` deletes the sentinel while session B is still mid-turn; the watcher exits within 1s; session B's next prompt re-spawns it, but the gap window is real. Given the ADR explicitly defers "external CI waiting" as out-of-scope and parallel Claude sessions are an adjacent concern, this is worth documenting rather than fixing.
**Suggestion:** Either note the multi-session caveat in ADR-033 → Consequences, or keep a small set-of-active-session-ids in the sentinel file and only stop when the set becomes empty.

**[maintainability]** `claude/scripts/awake-blocker.py:44-50` — `awake.log` grows unbounded
`_log` appends forever; nothing rotates or truncates. Likely fine in practice (a few hundred bytes per session boundary), but the file is in `scratch/`, which is otherwise ephemeral.
**Suggestion:** Either truncate the log at startup of each new watcher, or cap by size (rotate to `awake.log.1` past N KB). Trivial change; can be deferred.

**[correctness]** `claude/scripts/awake-blocker.py:163-167` — `Notification` semantics may be broader than "idle"
The ADR justifies treating `Notification` as "Claude is idle (waiting on user)". Per the [hooks reference](https://docs.claude.com/en/docs/claude-code/hooks), `Notification` fires for permission prompts *and* other notification types. If a future Notification subtype fires mid-processing (e.g., a non-blocking status surface), `stop()` would release the wake-lock prematurely; the next UserPromptSubmit then re-arms it, but the in-between window the machine could sleep.
**Suggestion:** When acting on a Notification event, inspect the stdin JSON for a subtype/`notification_type` field if present, and only `stop()` on the idle/permission-wait subtypes. If the field is unavailable, the current behavior is acceptable.

**[maintainability]** Diff contains a Python file but Step 2e's automated integrity scanners target JS/TS frameworks
The diff is pure-Python; the global Test Integrity grep patterns (`it.skip`, `xdescribe`, etc.) cannot inspect for `@pytest.mark.skip` / `pytest.skip(` / `unittest.skip`. None of those appear here on manual inspection (the file has no tests at all, by design — there are no test scripts in `claude/scripts/`), but the PR body's claim that "test-integrity violations: none" should be read as manually verified rather than tool-confirmed.
**Suggestion:** No action — manual verification is in place. Noting for the record per ADR-029 language-scope rule.

---

### Questions for Author

**Question:** Should the script short-circuit on non-Windows platforms instead of relying on the safe-exit guard?
**Context:** [`awake-blocker.py:88-91`](claude/scripts/awake-blocker.py:88) passes `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` to `Popen` unconditionally. On non-Windows, `creationflags` is accepted but those constants have no meaning and `["py", "-3", ...]` won't resolve. The `try/except` in `main()` catches it and logs `hook error:`, so behavior is correct, but every UserPromptSubmit on Linux/macOS writes an error line to `awake.log`.
**Tradeoffs:** (a) Add `if sys.platform != "win32": return 0` at the top of `main()` — silent no-op on other platforms, cleaner log, costs two lines. (b) Leave as-is — relies on documented safe-exit behavior, but produces log noise on any non-Windows machine that ever pulls dev-env. (c) Check platform inside `start()` only — slightly narrower than (a) and keeps the dispatch path uniform.

---

PR #295, opened 2026-06-01
Use pyw -3 for hook commands to suppress console window flash

### LINKED ISSUE #294: Hook py invocations briefly flash a console window
## Problem

Each `py -3 ...` hook command in `claude/settings.json` launches `python.exe` (console subsystem). When Claude Code spawns these hooks (PreToolUse, PostToolUse, UserPromptSubmit, Stop, etc.), Windows briefly allocates and tears down a console window for each one. Visually this is a small flash on every prompt and tool call.

## Fix

Swap `py -3` to `pyw -3` in hook commands only. `pyw.exe` is the Python Launcher's windowless variant — it invokes `pythonw.exe` (Windows subsystem), so no console allocation. Stdin/stdout/stderr still work normally over the pipes Claude Code wires up to the subprocess.

Scope: ~25 hook command entries in `claude/settings.json`. Leave `Bash(py -3 *)` permission entries alone — those govern Claude running py via the Bash tool, which executes inside the existing shell and does not pop a window. Skill docs and the `pre-push` hook also stay on `py -3` (they run from a shell).

## Why it matters

Cosmetic but high-frequency annoyance — every hook fire produces a flash. The swap is a one-character change per line, fully reversible.

brownm09:
## Summary

Swap `py -3` → `pyw -3` in `claude/settings.json` hook command entries (~25 hooks). `py.exe` invokes `python.exe` (console subsystem), so each hook spawn briefly allocates a console window — visible as a flash on every prompt and tool call. `pyw -3` invokes `pythonw.exe` (Windows subsystem), no console allocation. Stdin/stdout/stderr still work normally over the pipes Claude Code wires up to the subprocess.

Scope: `settings.json` hook commands only. Left unchanged:
- `Bash(py -3 *)` permission entries — Claude invoking via the Bash tool runs inside Git Bash, no flash.
- The `## Testing` command and skill `py -3` examples — shell contexts.
- The `pre-push` hook — runs from a `git push` context.

ADR-007 amended with a 2026-06-01 decision section recording the swap and scope. `docs/REFERENCE.md` updated to match.

Closes #294

## Testing

```
node -e "JSON.parse(require('fs').readFileSync('claude/settings.json','utf8'))"
```

Both pass. Verified via grep that 25 hook commands now use `pyw -3` and zero `"command": "py -3` remain.

`Tests: 0 passed, 0 skipped, 0 failed (n/a)` — no automated test suite; verification is the syntax + JSON parse pair above per the `## Testing` section in the project CLAUDE.md.

No new testable behavior introduced (config change to invocation only). Suppression check: clean. Test-integrity check: n/a (no test files touched).

## Test plan

- [x] Hook scripts parse cleanly
- [x] settings.json parses as valid JSON
- [x] Hook command grep shows 25× `pyw -3`, 0× `py -3`
- [ ] After merge, observe no console window flash on next session's hook fires

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -20,6 +20,12 @@ Run from the repo root to verify all hook scripts are syntax-clean:
 py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py
 ```
 
+Additionally, run the `pyw -3` stdio verification (Windows-only — confirms `pythonw.exe` honors parent-supplied pipes, the invariant ADR-007's 2026-06-01 decision relies on):
+
+```bash
+py -3 claude/scripts/tests/test_pyw_stdio.py
+```
+
 (On Windows, `python3` resolves to the Microsoft Store stub; use `py -3` — see [ADR-007](docs/adr/007-hook-command-invocation.md).)
 
 For docs-only changes to `claude/CLAUDE.md`: run `grep -n 'date -u' claude/CLAUDE.md` and
--- a/claude/scripts/tests/test_pyw_stdio.py
+++ b/claude/scripts/tests/test_pyw_stdio.py
@@ -1,0 +1,245 @@
+#!/usr/bin/env python3
+"""Verify `pyw -3` stdio behavior for Claude Code hook compatibility.
+
+`pythonw.exe` (which `pyw -3` resolves to) is built against the Windows
+subsystem rather than the console subsystem. The known wrinkle: when the
+parent process does NOT supply stdio handles, `pythonw` leaves
+`sys.stdin`, `sys.stdout`, and `sys.stderr` set to `None` instead of
+wiring them to a console. Any hook script that touches stdio under those
+conditions would crash with `AttributeError: 'NoneType' object has no
+attribute 'read'` (or `write`, or `flush`).
+
+Claude Code spawns hooks with stdio pipes (it has to — that's how it
+delivers the hook event payload as JSON on stdin and reads the
+`systemMessage` JSON back on stdout). So the question is empirical: does
+`pyw -3` honor parent-supplied pipes the same way `py -3` does?
+
+This test answers that question by running `pyw -3` under
+`subprocess.Popen` with `stdin=PIPE, stdout=PIPE, stderr=PIPE` — exactly
+the shape Claude Code uses — and asserting:
+
+  1. `sys.stdin`, `sys.stdout`, `sys.stderr` are all non-None inside the
+     child.
+  2. Bytes written to the child's stdin are readable via `sys.stdin.read()`.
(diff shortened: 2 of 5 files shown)


brownm09:
## Review: Use pyw -3 for hook commands to suppress console window flash

### Summary
Merge with minor changes. The swap is well-scoped (hook command entries only, shell contexts unchanged), the rationale is recorded in ADR-007 with a clean History entry, and `docs/REFERENCE.md` is updated to match. The single concern worth surfacing before merge is a `pythonw.exe` stdio risk — `pythonw` historically sets `sys.stdin/stdout/stderr` to `None` when the parent does not provide pipes. Empirically the hooks were exchanging JSON over stdio under `py -3`, so Claude Code is presumably wiring pipes, but it is worth confirming on the first post-merge session that no hook silently breaks (e.g., a script that previously relied on `print()` to a real console handle for a fallback path).

---

### Non-Blocking Findings

**[maintainability]** `docs/REFERENCE.md:119`
The amended sentence lists shell-invoked Python as "the `## Testing` command, the `pre-push` hook" but the ADR's matching scope list also includes "skill docs". Two adjacent documents giving slightly different scope lists for the same rule is a small drift hazard — future readers may interpret REFERENCE.md as authoritative and forget that skill examples are also `py -3`.
**Suggestion:** Add "skill `py -3` examples" to the REFERENCE.md sentence so it matches the ADR-007 Scope bullet list.

**[maintainability]** `docs/adr/007-hook-command-invocation.md:74`
The previous Consequences bullet noted that `python3 -m py_compile ...` would silently no-op on this machine — useful trivia that explained why the `## Testing` command uses `py -3 -c "import ast"` rather than `python3`. The amended Consequences removed that note. The information now lives only in the 2026-05-26 History entry.
**Suggestion:** Either keep a one-line breadcrumb in Consequences ("`python3` still resolves to the App Execution Alias on this machine — shell commands must use `py -3`, not `python3`") or accept the loss as redundant with the History entry. Low priority.

---

### Questions for Author

**Question:** Have you confirmed (or will you confirm on first post-merge session) that every hook script tolerates `pythonw.exe`'s stdio behavior?
**Context:** `claude/settings.json` lines 36–178 now invoke 25 hooks via `pyw -3`. `pythonw.exe` differs from `python.exe` in that, without a parent-supplied stdio pipe, `sys.stdin`, `sys.stdout`, and `sys.stderr` are `None` rather than wired to a console handle. The PR body asserts Claude Code wires pipes, which matches observed behavior, but no script-by-script check is documented.
**Tradeoffs:** (a) Trust the empirical "hooks already exchanged JSON over stdio" evidence and verify on first session — fastest path, with the risk that an edge-path (e.g., a `try/except` that falls back to `print` after JSON output) silently swallows output. (b) Grep `claude/scripts/*.py` for `sys.stdin`/`sys.stdout`/`print(` and confirm each access is either piped JSON exchange or guarded — cheap (one grep) and removes the residual uncertainty before merge.

---

### Style Notes

None.

---

PR #299, opened 2026-06-01
Suppress console-window flashes from hook subprocesses

### LINKED ISSUE #297: Hook subprocess spawns flash console windows under pyw -3
After dev-env#295 (PR #295) swapped `py -3` → `pyw -3` for `settings.json` hook commands, the Python launcher itself no longer flashes a console. But the cosmetic flash persists because the hook scripts then spawn `git`, `gh`, and `bash` as subprocesses. Under `pythonw.exe` (no console), every child console application gets a freshly-allocated console window — visible as a flash for `git.exe`, the Git Bash `mingw` console, etc.

## Evidence

- 19 hook scripts in `claude/scripts/` use `subprocess.run` / `subprocess.Popen`.
- Only `awake-blocker.py` already sets `CREATE_NO_WINDOW` (because its detached `pythonw.exe` watcher was the original flash source ADR-033 hit).
- User reports `py.exe` and `mingw` console flashes still occur after the #295 merge.

## Root cause

[Python docs](https://docs.python.org/3/library/subprocess.html#subprocess.CREATE_NO_WINDOW): without `creationflags=CREATE_NO_WINDOW`, Windows allocates a new console for any child console application when the parent has no console.

## Fix

Add `claude/scripts/_winsubp.py` — on import, monkey-patches `subprocess.Popen.__init__` to OR `CREATE_NO_WINDOW` into `creationflags`. Idempotent (sentinel attribute), safe-fallback (try/except), no-op on non-Windows.

Each subprocess-using hook script adds one line near the top:
\`\`\`python
import _winsubp  # noqa: F401  -- suppress console windows on Windows
\`\`\`

## Verification

Extend `test_pyw_stdio.py` with a fourth test: spawn `pyw -3` running a script that imports `_winsubp` then runs `subprocess.run(["git","--version"])` — assert it succeeds and that `creationflags` was set to include `CREATE_NO_WINDOW`.

Followup to #294 / #295.

brownm09:
## Summary

After #295 swapped the launcher to `pyw -3`, the launcher itself stopped flashing — but the user still saw `py.exe` and `mingw` console flashes. Root cause: every `subprocess.run`/`Popen` call inside a hook still allocated a fresh console window for its child (`git`, `gh`, `bash`, `py`). Under `pythonw.exe` (no console), Windows allocates a console per child unless `creationflags=CREATE_NO_WINDOW` is set.

Primary source: [Python docs — `subprocess.CREATE_NO_WINDOW`](https://docs.python.org/3/library/subprocess.html#subprocess.CREATE_NO_WINDOW); [Microsoft — Process Creation Flags](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags).

## Approach

- Add `claude/scripts/_winsubp.py` — on import, monkey-patches `subprocess.Popen.__init__` to OR `CREATE_NO_WINDOW` into `creationflags`. Idempotent (sentinel `subprocess._winsubp_patched`), no-op on non-Windows, wrapped in `try/except` so a future Python signature change cannot crash a hook.
- `import _winsubp  # noqa: F401  -- suppress console windows on Windows` added near the imports of all 18 subprocess-using hook scripts.
- Extended `test_pyw_stdio.py` with two new checks:
  4. Under `pyw -3`, `import _winsubp` patches subprocess and leaves `subprocess.run` functional (echo round-trip).
  5. Static scan — every hook script that uses `subprocess` imports `_winsubp`. **Prevents new hooks from silently re-introducing the flash.**
- ADR-007 amended with a 2026-06-01 follow-up section.
- `docs/REFERENCE.md` Authoring rules: rule 3 reworded (`pyw -3`, not `py -3`); new rule 4 (`import _winsubp` whenever a hook spawns subprocesses).
- `README.md` Hooks intro notes the helper.

Closes #297

## Testing

```
```

Both pass. AST: clean across all hook scripts. Stdio + patch tests:

```
Tests: 5 passed, 0 skipped, 0 failed (n/a)
```

(No time tracked — Python's `time` module is unused in this test runner; the duration is not measured.)

**Coverage gate.** New testable behavior is the `_winsubp` patch itself. Two new tests (4 and 5 above) cover it — runtime patch behavior + static "every subprocess-using hook imports the helper" guard. Future hooks that forget the import will fail check 5.

**Suppression check.** Clean (`git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|...)'` returns nothing).

**Test integrity check.** No tests deleted or skipped; no coverage thresholds changed. Two tests added.

## Test plan

- [x] AST parse clean
- [x] test_pyw_stdio.py: 5 passed
- [x] Static scan confirms all 18 subprocess-using hooks import `_winsubp`
- [ ] After merge, restart Claude Code session and confirm no `py.exe` / `mingw` console flashes

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ Custom slash commands loaded from `claude/skills/`. Invoke with `/skill-name [ar
 Hook scripts run automatically via Claude Code's `hooks` configuration in `claude/settings.json`.
 Most hooks are advisory — they emit reminders but do not block tool execution. The exception is `pre-tool-use-worktree-path-check.py`, which blocks `Write`, `Edit`, and `NotebookEdit` calls that target the canonical repo root instead of the active worktree.
 
+Hooks that spawn subprocesses (`git`, `gh`, `bash`, …) must `import _winsubp` — a helper module at `claude/scripts/_winsubp.py` that patches `subprocess.Popen.__init__` to set `CREATE_NO_WINDOW` so children don't flash a console window under `pythonw.exe`. See [ADR-007](docs/adr/007-hook-command-invocation.md).
+
 | Event | Script | Purpose |
 |---|---|---|
 | UserPromptSubmit | `session-mode-prompt.py` | Injects a one-time mode-confirmation reminder into Claude's context on the first prompt of each new session |
--- a/claude/scripts/_winsubp.py
+++ b/claude/scripts/_winsubp.py
@@ -20,12 +20,26 @@ Properties:
   - Idempotent — guarded by `subprocess._winsubp_patched`. Importing twice
     is harmless.
   - No-op on non-Windows (`os.name != 'nt'`).
-  - Safe-fallback — any unexpected failure during patch installation is
-    swallowed so a hook never crashes because of this module.
+  - Safe-fallback — any unexpected failure during patch installation OR at
+    Popen call time is swallowed so a hook never crashes because of this
+    module. If the patch fails for any reason, the worst outcome is the
+    original cosmetic flash returns.
   - Caller-respecting — if the caller already passed `creationflags`, the
     flag is OR-merged, not overwritten. Callers that explicitly want a
     console (none currently exist among hooks) can still set
     `CREATE_NEW_CONSOLE` and both flags coexist.
+
+Assumptions:
+  - Callers pass `creationflags` as a keyword argument, never positionally.
+    `subprocess.Popen.__init__` accepts `creationflags` as its 13th positional
+    parameter; a caller that passes 13+ positional args would collide with
+    `kwargs["creationflags"]` and raise TypeError. Every hook in this repo
+    uses kwargs, so the simple kwargs-only path is correct in practice; the
+    runtime `try/except` below covers the hypothetical regression.
+  - The generic `(self, *args, **kwargs)` signature on the patched init
+    intentionally shadows `Popen.__init__`'s detailed signature — IDE and
+    type-checker parameter help should be consulted on `subprocess.Popen`
+    directly.
 
 Usage:
     import _winsubp  # noqa: F401  -- suppress console windows on Windows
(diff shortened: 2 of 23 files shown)


brownm09:
## Review: Suppress console-window flashes from hook subprocesses

### Summary
Merge with minor changes. The monkey-patch approach is well-scoped, idempotent, safe-fallback, and the static test in `test_pyw_stdio.py` is exactly the right forcing function to prevent regression. Highest-severity concern: the static-scan regression guard has two narrow false-negative paths — a comment containing `_winsubp` (substring match) and a `from subprocess import run` idiom would both bypass the check. Worth tightening before merge so the guard cannot be silently bypassed by reasonable future code.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/_winsubp.py:74-77`
The patched `__init__` reads `creationflags` from `kwargs` only. `subprocess.Popen.__init__` accepts `creationflags` as the 13th positional argument; a caller passing all positional args would land `creationflags` outside `kwargs`, and the patch would then inject `kwargs["creationflags"]` alongside the positional value, raising `TypeError: got multiple values for argument 'creationflags'` at every Popen call (the outer module-level `try/except` does not cover runtime calls). No hook currently passes 13+ positionals, so this is latent.
**Suggestion:** Either document the assumption in the module docstring ("assumes callers never pass `creationflags` positionally — true of every hook in this repo"), or wrap the body of `_patched_popen_init` in `try/except` so a signature collision degrades to "no patch applied" rather than "every Popen crashes."

**[correctness]** `claude/scripts/tests/test_pyw_stdio.py:347-353` (static-scan check 5)
The regression guard uses substring match `"_winsubp" not in text` to declare a hook compliant. A hook that merely mentions `_winsubp` in a comment or docstring — without importing it — would pass. Also, the subprocess-detector keys on `subprocess.{run,Popen,…}` only; a hook using `from subprocess import run; run(...)` would be skipped entirely. Both are narrow but each lets a future regression slip past the guard the ADR depends on.
**Suggestion:** Tighten the import check to a regex like `^\s*import\s+_winsubp\b` (multiline) and broaden the subprocess detector to also match `from subprocess import`. Add a quick negative-control comment in the test asserting that an obviously-non-importing reference is rejected.

**[maintainability]** `claude/scripts/_winsubp.py:74`
`def _patched_popen_init(self, *args, **kwargs)` shadows the real `Popen.__init__` signature, so IDEs and type checkers lose the parameter help across the entire codebase. Acceptable for an unconditional pass-through, but worth a one-line note in the docstring that this is intentional and that callers should consult `subprocess.Popen` for parameter help.
**Suggestion:** Add a one-liner to the docstring: "Signature intentionally generic; consult `subprocess.Popen` for parameter docs."

---

### Style Notes
The 18 hook edits all place `import _winsubp` immediately above the first standard-library import, with a consistent `# noqa: F401  -- suppress console windows on Windows` comment. That's the right shape — it sits before `import subprocess` so the patch always lands first, and the comment explains why a seemingly-unused import is intentional. Good consistency.

---

PR #301, opened 2026-06-02
Spawn awake-blocker watcher via sys.executable (windowless chain)

### LINKED ISSUE #300: awake-blocker watcher still flashes py.exe console window on every prompt
After #299 added `_winsubp` to suppress console allocation for hook subprocesses, every prompt still flashes a `py.exe` console window. Root cause: `awake-blocker.py` line 145 spawns its detached watcher via `["py", "-3", ...]` — the console launcher. `_winsubp` correctly OR's `CREATE_NO_WINDOW` into `creationflags`, so the `py.exe` process itself does not get a window. But `py.exe` is a console-subsystem program that then calls `CreateProcess` on `python.exe` *without* passing `CREATE_NO_WINDOW` through. Because `py.exe`'s own process has no console (we suppressed it), Windows allocates a fresh console for the child `python.exe` — visible as the residual flash.

**Fix:** Spawn the watcher via `sys.executable` instead. The hook runs under `pyw -3`, so `sys.executable` is `pythonw.exe` (Windows subsystem, no console). No launcher chain, no console allocation at any level. Update `WATCHER_IMAGE` to derive from `Path(sys.executable).name` so the `_pid_is_watcher` tasklist check still matches.

Impact: High — every prompt flashes a console window, defeating the goal of #294/#295/#297.

brownm09:
## Summary

The residual `py.exe` console flash on every prompt — still present after #295 (`py -3` → `pyw -3` in settings.json) and #299 (`_winsubp` adds `CREATE_NO_WINDOW` to subprocess spawns) — came from `awake-blocker.py` spawning its detached watcher via `[\"py\", \"-3\", ...]`. `_winsubp` correctly suppressed `py.exe`'s own console, but `py.exe` is a console-subsystem launcher: it spawns `python.exe` without passing the flag through, and because `py.exe` had no console (we suppressed it), Windows allocated a fresh console for the grandchild `python.exe`. Flash.

## Fix

- Spawn the watcher via `sys.executable` rather than the `py -3` launcher. Under the `pyw -3` hook invocation, `sys.executable` is `pythonw.exe` (Windows-subsystem interpreter, no console). No launcher chain → no grandchild console → no flash.
- Derive `WATCHER_IMAGE` from `Path(sys.executable).name` so the `tasklist`-based PID-reuse defense in `_pid_is_watcher` tracks whichever interpreter actually spawned the watcher.
- ADR-007 amended with a 2026-06-01 follow-up 2 section explaining the launcher-chain bug.

Primary sources: [Microsoft — Process Creation Flags](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags), [Python — \`sys.executable\`](https://docs.python.org/3/library/sys.html#sys.executable).

Closes #300

## Testing

\`\`\`
py -3 -c \"import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]\" claude/scripts/*.py
\`\`\`

Both pass. AST clean. Stdio + patch tests:

\`\`\`
Tests: 5 passed, 0 skipped, 0 failed
\`\`\`

**Coverage gate.** The new behavior (watcher spawn path) is exercised indirectly by the existing static check (test 5 — all subprocess-using hooks import \`_winsubp\`). No programmatic way to assert \"no console window was allocated.\" Final verification is manual: restart Claude Code, confirm no flash.

**Suppression / integrity checks.** Both clean.

## Test plan

- [x] AST clean across all hook scripts
- [x] test_pyw_stdio.py: 5/5 pass
- [ ] After merge, restart Claude Code and confirm no \`py.exe\` flash on prompts (user-side)

### DIFF
--- a/claude/scripts/awake-blocker.py
+++ b/claude/scripts/awake-blocker.py
@@ -40,12 +40,15 @@ EXEC_STATE_REFRESH = 30.0     # seconds between SetThreadExecutionState calls
 # Log rotation: when LOG_FILE exceeds this many bytes, rotate to LOG_FILE_ROTATED.
 LOG_MAX_BYTES = 256 * 1024  # 256 KiB
 
-# Watcher process image name — used together with PID to defeat PID-reuse false positives.
-# Derived from `sys.executable` so it tracks whatever interpreter actually spawned the
-# watcher (`pythonw.exe` under the `pyw -3` hook invocation; `python.exe` if a developer
-# runs the hook manually via `py -3` for debugging). This must stay in sync with the
-# executable used to spawn the watcher in start() below.
-WATCHER_IMAGE = Path(sys.executable).name if sys.executable else "pythonw.exe"
+# Watcher executable + image name — single source of truth so the spawn argv (start())
+# and the `tasklist` PID-reuse defense (_pid_is_watcher) cannot drift. Derived from
+# `sys.executable` so it tracks whatever interpreter actually spawned the watcher
+# (`pythonw.exe` under the `pyw -3` hook invocation; `python.exe` if a developer
+# runs the hook manually via `py -3` for debugging). Falls back to `pythonw.exe`
+# in the (extremely rare on Windows) case where `sys.executable` is empty — both
+# the spawn and the image-match check then resolve consistently against PATH.
+_WATCHER_EXEC = sys.executable or "pythonw.exe"
+WATCHER_IMAGE = Path(_WATCHER_EXEC).name
 
 # SetThreadExecutionState flags
 ES_CONTINUOUS = 0x80000000
@@ -157,7 +160,7 @@ def start() -> None:
     creationflags = DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP
 
     proc = subprocess.Popen(
-        [sys.executable, str(Path(__file__).resolve()), "--watcher"],
+        [_WATCHER_EXEC, str(Path(__file__).resolve()), "--watcher"],
         stdin=subprocess.DEVNULL,
         stdout=subprocess.DEVNULL,
         stderr=subprocess.DEVNULL,
--- a/claude/scripts/awake-blocker.py
+++ b/claude/scripts/awake-blocker.py
@@ -41,7 +41,11 @@ EXEC_STATE_REFRESH = 30.0     # seconds between SetThreadExecutionState calls
 LOG_MAX_BYTES = 256 * 1024  # 256 KiB
 
 # Watcher process image name — used together with PID to defeat PID-reuse false positives.
-WATCHER_IMAGE = "py.exe"
+# Derived from `sys.executable` so it tracks whatever interpreter actually spawned the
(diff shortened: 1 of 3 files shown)


brownm09:
## Review: Spawn awake-blocker watcher via sys.executable (windowless chain)

### Summary
Merge with minor changes. The diagnosis is correct and well-documented: `_winsubp` cannot propagate `CREATE_NO_WINDOW` through the `py.exe` launcher intermediary, so using `sys.executable` directly bypasses the launcher chain and eliminates the grandchild console allocation. The ADR follow-up captures the rationale clearly. Two non-blocking gaps worth addressing: a fallback-vs-spawn inconsistency, and a missing regression test that would catch a future `py`/`py.exe` re-introduction without requiring a human to watch for flashes.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/awake-blocker.py:48`
`WATCHER_IMAGE` falls back to `"pythonw.exe"` when `sys.executable` is empty, but the `Popen` call at line 160 uses `sys.executable` directly with no equivalent fallback. If `sys.executable` is ever `""` (extremely rare on Windows, but defensively considered elsewhere in this file), the watcher spawn will fail with FileNotFoundError while WATCHER_IMAGE will still match `pythonw.exe`. The two values should either share a fallback or both fail loud.
**Suggestion:** Compute `_EXEC = sys.executable or "pythonw.exe"` once at module load, then use `_EXEC` in both the `Popen` argv and `Path(_EXEC).name` for `WATCHER_IMAGE`. Keeps the comment's "must stay in sync" invariant enforced by code rather than convention.

**[maintainability]** `claude/scripts/tests/test_pyw_stdio.py` (not in this diff)
The static check in test 5 verifies every subprocess-using hook imports `_winsubp`, but nothing asserts that `awake-blocker.py` spawns the watcher via `sys.executable` rather than `"py"`. The exact regression this PR fixes — naming `py` in a `subprocess.Popen` argv inside a hook — is not detectable by any existing test. A future refactor could silently re-introduce it, and the only verification path is a human watching the screen for flashes.
**Suggestion:** Add a static-scan test that fails if any file in `claude/scripts/` contains `["py"` or `"py", "-3"` as the first element of a list literal (or similarly: any `subprocess.Popen`/`subprocess.run` whose first argv element is the string literal `"py"`). Pair with an allow-list comment to permit legitimate non-spawn uses.

**[documentation]** `docs/adr/007-hook-command-invocation.md` (Consequences section, unchanged)
The Consequences list still names only the `pyw -3` settings.json requirement and the `_winsubp` import requirement. The new rule established by follow-up 2 — "any hook that spawns Python must use `sys.executable`, never `py` or `py -3`" — is stated in the narrative but not codified in Consequences where future readers scan for rules.
**Suggestion:** Add a Consequences bullet: "Any hook in `claude/scripts/` that spawns Python as a subprocess must use `sys.executable` (or `pyw -3` if the command must be a literal string), never `py` or `py -3`. Naming `py` re-introduces the launcher-chain console flash even under `_winsubp`."

---

PR #307, opened 2026-06-04
[config] Global lockfile-drift prevention: pre-push hook + dependency-edit rule

### LINKED ISSUE #305: [config] Global lockfile-drift prevention: pre-push hook + dependency-edit CLAUDE.md rule
## Problem

Across npm repos, editing a dependency in `package.json` without regenerating `package-lock.json` produces a lockfile that drifts from the declared ranges. CI then fails at `npm ci` (which refuses an out-of-sync lockfile) with a cryptic `EUSAGE` error, and `main` can go red. This recently happened on `brownm09/lifting-logbook` (clerk-backend, clerk-shared, react drift), fixed reactively but with no global prevention. Per-repo CI guards (e.g. lifting-logbook#435) catch it after push; nothing stops the drift locally or codifies the author obligation.

## Goal

Add the two global (cross-repo) layers of a defense-in-depth strategy:

1. **Global pre-push hook** (`claude/hooks/pre-push`, run via `core.hooksPath`): when the push range touches any `package.json`, run a lockfile-sync check for the affected repo and **block the push on drift**, with a clear remediation message. Must chain to any existing repo-level `.git/hooks/pre-push`, skip cleanly when `npm` is absent, work under Git Bash on Windows, and follow the three hook invariants (atomic, safe-exit guard, no `bash -c`).
2. **Global CLAUDE.md rule** (`claude/CLAUDE.md`): editing any dependency in a `package.json` requires running `npm install` and committing the regenerated lockfile in the same change, plus a pre-PR lockfile-sync check.

This is ADR-warranted (changes a hook and a documented global rule); capture #1 and #2 in a single ADR.

## Acceptance Criteria

- [ ] `claude/hooks/pre-push` blocks a push that touches `package.json` when the lockfile is out of sync, with an actionable message
- [ ] Hook chains to a pre-existing repo-level `.git/hooks/pre-push`
- [ ] Hook skips cleanly (exit 0) when `npm` is not on PATH
- [ ] Works under Git Bash on Windows; respects the three hook invariants
- [ ] `claude/CLAUDE.md` documents the dependency-edit + pre-PR lockfile-sync rule
- [ ] New ADR covers both layers; `docs/adr/INDEX.md` and `README.md` hooks table updated

brownm09:
## Summary

Layers 3 and 4 of a 4-layer lockfile-drift defense (layers 1–2 are per-repo: branch protection + a CI guard, implemented for lifting-logbook in [lifting-logbook#435](https://github.com/brownm09/lifting-logbook/issues/435)). Background: a `package.json`/`package-lock.json` drift turned lifting-logbook `main` red at `npm ci` (fixed reactively in #432/#433). These two **global** layers apply to every npm repo with no per-repo wiring.

### Layer 3 — global `pre-push` hook (`claude/hooks/pre-push`)
Extends the existing hook (one `pre-push` per repo under `core.hooksPath`, so extension — not a second file — is the only correct shape). When the push range touches any `package.json` (root or workspace), it regenerates the lockfile metadata and **blocks the push on drift** with an actionable message. Properties:
- **Non-destructive:** backs up and restores the working-tree lockfile on every path (drift / in-sync / npm-failure).
- **Skips cleanly** when `npm` is absent or the repo has no root lockfile.
- **Does not block on tooling failure** (npm offline → warn + proceed; CI remains the backstop).
- **Chains** to any repo-level `.git/hooks/pre-push`.

### Layer 4 — `claude/CLAUDE.md` 'Dependency and lockfile policy'
Rule 1: editing a dependency requires regenerating + committing `package-lock.json` in the same change. Rule 2: a pre-PR lockfile-sync check (`npm install --package-lock-only --ignore-scripts && git diff --exit-code package-lock.json`).

### Docs
New [ADR-036](docs/adr/036-lockfile-drift-prevention.md) (covers both layers, with primary-source references); `docs/adr/INDEX.md` row; `README.md` hooks-table row updated (doc-reconciliation).

## Testing

dev-env has no Jest suite; verification per the project `## Testing` section plus targeted functional tests:

- `py -3 -c "import ast ..." claude/scripts/*.py` → **ALL PYTHON SCRIPTS PARSE OK**
- `bash -n claude/hooks/pre-push` → **HOOK SYNTAX OK**
- **Functional (guard core, against a real npm repo):**
  - Drifted `@types/jest` range → **BLOCK(drift)**, lockfile restored **PRISTINE**, backup cleaned up.
  - In-sync lockfile → **PASS(in-sync)**.
  - `npm` absent → **SKIP cleanly** (exit 0, no block).
  - Working tree clean after all paths.
- Suppression / test-integrity greps on the diff: **(none)**.

No automated test failures. Diff is one bash hook + three markdown/docs files; no Python scripts changed.

## Acceptance Criteria

- [x] Hook blocks a drifting push with an actionable message
- [x] Chains to a pre-existing repo-level `.git/hooks/pre-push`
- [x] Skips cleanly (exit 0) when `npm` is not on PATH
- [x] Works under Git Bash on Windows; respects the three hook invariants
- [x] `claude/CLAUDE.md` documents the dependency-edit + pre-PR lockfile-sync rule
- [x] ADR covers both layers; INDEX + README updated

Closes #305

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Hooks that spawn subprocesses (`git`, `gh`, `bash`, …) must `import _winsubp` 
 | Stop | `token-tracker.py` | Aggregates session token usage to `scratch/token-sessions.jsonl` |
 | Stop | `journal-stop-check.py` | Checks sentinel flag and stale open journal stubs at session end; emits closing reminder if stub was pushed this session |
 | PostCompact | `post-compact.py` | Emits compaction status line (trigger type + remaining tokens) |
-| Git pre-push | `hooks/pre-push` | Warns when branch merge base diverges from `origin/main` in squash-merge repos |
+| Git pre-push | `hooks/pre-push` | Warns when branch merge base diverges from `origin/main` in squash-merge repos; blocks engineering-journal pushes to already-merged draft branches; blocks pushes that drift `package-lock.json` from `package.json` (see [ADR-036](docs/adr/036-lockfile-drift-prevention.md)) |
 
 ## Routines
 
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -359,6 +359,22 @@ A PR that adds new failures or lists no PR-body justification for outstanding `p
 
 **Scope note.** The first implementation supports Jest only. Pytest, Go `testing`, and Rust have different JSON output formats and each needs its own parser — a clean run in a non-Jest repo is not evidence the policy applies. The opt-in flag should stay off in repos whose test runners are not yet supported.
 
+### Dependency and lockfile policy
+
+In any npm repo, `package-lock.json` must stay in sync with the dependency ranges declared in `package.json`. When they drift, `npm ci` refuses to install (it requires an in-sync lockfile) and CI fails at the install step with a cryptic `EUSAGE` error — and `main` can go red. The motivating incident is lifting-logbook #432/#433, where clerk-backend, clerk-shared, and react drift reddened `main` and required reactive recovery PRs. See [ADR-036](../docs/adr/036-lockfile-drift-prevention.md) for the full defense-in-depth rationale.
+
+**Rule 1 — Edit a dependency, regenerate the lockfile, commit both together.**
+Any change that adds, removes, or changes the version range of a dependency in a `package.json` (root or workspace) **must** run `npm install` and commit the regenerated `package-lock.json` in the **same** change. Never commit a `package.json` dependency edit without its lockfile update.
+
+**Rule 2 — Pre-PR lockfile-sync check (required before `gh pr create` when any `package.json` changed).**
+Run this from the repo root and confirm it exits clean before opening the PR:
+```bash
+npm install --package-lock-only --ignore-scripts && git diff --exit-code package-lock.json
+```
+A non-empty diff means the lockfile is out of sync — run `npm install`, commit the result, and re-run the check. The same check runs automatically in the global `pre-push` hook (it blocks the push on drift) and should run in each repo's CI as an early step; the pre-PR run catches it before either fires.
+
+**Scope note.** Applies to npm/`package-lock.json` repos. Yarn (`yarn.lock`) and pnpm (`pnpm-lock.yaml`) have equivalent drift failure modes but different regeneration commands (`yarn install`, `pnpm install --lockfile-only`) — extend the check per repo before relying on it there.
+
 ---
 
 ## Hook Safety
--- a/claude/hooks/pre-push
+++ b/claude/hooks/pre-push
@@ -66,28 +66,32 @@ done
 # Lockfile-drift guard: when a pushed package.json changed, verify package-lock.json
 # is in sync with it, and block the push on drift so CI never hits the cryptic
 # `npm ci` EUSAGE error. Skips cleanly when npm is absent or the repo has no root
-# lockfile. The check is non-destructive — the working-tree lockfile is always
-# restored, whatever the outcome.
+# lockfile. The check is non-destructive — the working-tree lockfile is restored on
+# every exit path (including SIGINT/SIGTERM) by a trap, and the backup is held
+# outside the repo (mktemp) so an interrupted run never leaves a stray file behind.
+# Note: the check evaluates the current working tree, not the exact pushed tree; for
+# the common clean-tree push of HEAD these coincide, and CI is authoritative on the
+# committed state regardless.
 if [ "$PKG_TOUCHED" = "1" ] && command -v npm >/dev/null 2>&1; then
   repo_root=$(git rev-parse --show-toplevel 2>/dev/null)
   if [ -n "$repo_root" ] && [ -f "$repo_root/package.json" ] && [ -f "$repo_root/package-lock.json" ]; then
     lock="$repo_root/package-lock.json"
-    backup="${lock}.prepush.bak"
+    backup=$(mktemp 2>/dev/null) || backup="${lock}.prepush.bak"
     cp "$lock" "$backup"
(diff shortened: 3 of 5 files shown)


brownm09:
## Review: [config] Global lockfile-drift prevention: pre-push hook + dependency-edit rule

### Summary
**Merge with minor changes considered.** The hook logic is correct on all the paths that matter, the non-destructive backup/restore is implemented on every branch, the npm-absent and no-lockfile skips are clean, and the docs (ADR-036 + INDEX + README + CLAUDE.md rule) are thorough, internally consistent, and link-correct. No blocking findings. The single finding worth acting on before merge is the **backup file living inside the repo working tree with no `trap` protection** — a hard interrupt mid-`npm install` can leave a stray `package-lock.json.prepush.bak` and a regenerated lockfile behind. Everything else is edge-case notes and one design question about working-tree vs. pushed-ref semantics.

---

### Non-Blocking Findings

**[reliability]** `claude/hooks/pre-push:64-66` (backup lifecycle)
The backup is written to `${lock}.prepush.bak` inside the repo root and restored only on the three normal code paths. There is no `trap`, so if the push is interrupted (Ctrl-C, SIGTERM) during `npm install --package-lock-only` — the longest operation in the block — the restore never runs: a regenerated `package-lock.json` and a stray `.prepush.bak` are left in the working tree. The `.bak` also sits in the repo, where a later `git add .` could stage it.
**Suggestion:** Put the backup outside the tree and guard it with a trap, e.g. `backup=$(mktemp)` then `trap 'cp "$backup" "$lock" 2>/dev/null; rm -f "$backup"' EXIT` set immediately after the `cp`. That makes restore signal-safe and keeps the artifact out of the repo entirely. The explicit `cp`/`rm` lines can then be dropped from the three branches (the trap covers them), or kept as belt-and-suspenders.

---

**[reliability]** `claude/hooks/pre-push:60-72` (validates the working tree, not the pushed refs)
`PKG_TOUCHED` is decided from the pushed commit range (correct), but the drift check itself runs `npm install --package-lock-only` against the **working tree** `package.json` and compares to the **working-tree** lockfile. For the common case (clean tree, pushing `HEAD`) these coincide. But if the tree has uncommitted `package.json` edits while pushing commits that also touch `package.json`, the hook judges the dirty tree, not the refs being pushed — it can block a push whose committed state is actually in sync.
**Suggestion:** Acceptable for a local backstop (CI Layer 2 is authoritative on committed state), but worth one line in the error message or a code comment noting the check reflects the current working tree. A fully ref-accurate check would require materializing the pushed tree (e.g., `git archive`/worktree), which is heavier than this layer warrants — documenting the limitation is the right call.

---

**[maintainability]** `claude/hooks/pre-push:51-54` (new-branch base resolution assumes `origin/main`)
For a new branch (`remote_sha` all-zeros) the range base is `origin/main`; if that ref doesn't resolve (repo whose default branch is `master`/`trunk`, or `origin` not yet fetched), `base` is empty, `pkg_range` is empty, and the lockfile check is silently skipped for that ref. Existing-branch pushes (`remote_sha..local_sha`) are unaffected.
This is consistent with the hook's existing `origin/main` assumption (the merge-base warning above uses it too), so it's not a regression — just a coverage gap for new branches in non-`main`-default repos.
**Suggestion:** Leave as-is unless a non-`main`-default npm repo enters the rotation; if one does, fall back to `git rev-parse "@{u}"` or the symbolic `origin/HEAD` before giving up.

---

**[maintainability]** `claude/hooks/pre-push` (no regression test for the hook)
The hook now contains real branching logic (range detection, three-way npm outcome, restore), verified manually this session, but dev-env has no automated test for `pre-push` the way `test_pyw_stdio.py` guards the Python hooks. A future edit could silently break the restore path.
**Suggestion:** Optional and out of scope for this PR — consider a lightweight `bats` (or shell) self-test that drives the guard block with a stubbed `npm` against a fixture repo, asserting BLOCK/PASS/SKIP and a clean tree afterward. Track in a follow-up issue rather than expanding this PR.

---

### Questions for Author

**Question:** Should the backup be relocated outside the repo (mktemp + trap) in this PR, or deferred?
**Context:** `claude/hooks/pre-push:64` writes `${lock}.prepush.bak` into the repo root and restores it only on normal exits; an interrupt during `npm install` leaves both a regenerated lockfile and the `.bak` behind (first finding above).
**Tradeoffs:** (a) Fix now with `mktemp`+`trap` — a ~3-line change, makes the hook signal-safe and repo-clean, and is the kind of thing that's easy to land while the context is fresh. (b) Defer to a follow-up — keeps this PR's diff minimal, but the leftover-artifact failure mode ships in the meantime. Given the fix is small and the hook is global (runs in every repo), (a) is the low-risk choice.

---

### Style Notes
Comments are excellent — the "Layer 3 of the lockfile defense" cross-references and the inline rationale on the `exit 1` (intentionally skips the repo-hook chain) make the control flow legible. ADR-036 is a model entry: the four-layer table, the explicit "why working-tree not HEAD" alternative, and primary-source References (npm-ci, githooks, core.hooksPath) all match the repo's documentation bar.

---

brownm09:
**Review findings addressed** (commit `48f3c86`):

- **[reliability] backup lifecycle** — backup now goes to a `mktemp` file *outside* the repo and is restored via `trap ... EXIT INT TERM`, so an interrupted push leaves neither a stray `.prepush.bak` nor a regenerated lockfile. Re-tested: BLOCK / PASS / SKIP all pass, tree clean, no stray file. ADR-036 wording updated to match.
- **[reliability] working-tree vs pushed-refs** — documented in a hook comment (the check evaluates the working tree; CI is authoritative on committed state). Accepted as the right scope for a local backstop.
- **[maintainability] new-branch base resolution** — accepted as-is; consistent with the hook's existing `origin/main` assumption (the merge-base warning uses it too). Will revisit if a non-`main`-default npm repo enters rotation.
- **[maintainability] no regression test** — filed as follow-up #308 (Low) rather than expanding this PR.

PR #309, opened 2026-06-04
Automate disk reclamation for idle worktree node_modules (#306)

### LINKED ISSUE #306: [chore] Automate disk reclamation for stale worktree node_modules (C: hit 100% full)
## Problem

The `C:` drive filled to **100% (476G used, 0 bytes free)** during a session, hard-blocking all work: `npm install`, `npm run build`, git commits, Prisma migrations, and even writing a plan file all failed with `ENOSPC: no space left on device`.

### Root cause (measured 2026-06-03)

- **33 git worktrees** under `C:/Users/brown/Git/lifting-logbook/.claude/worktrees/`, each carrying its own full monorepo `node_modules`. The project CLAUDE.md (`## Worktree Setup`) mandates `npm install` in every worktree because the Husky pre-commit hook needs a local `turbo` binary. At ~1–2 GB per worktree that is an estimated 30–60 GB of duplicated, mostly-stale dependencies.
- Docker is only ~4.7 GB total (400 MB reclaimable) — **not** a meaningful contributor.
- `.claude/scratch` 239 MB, `AppData/Local/Temp` 872 MB — minor.

The existing `prune-merged-worktrees.py` (weekly) and `prune-stale-worktrees` routine are evidently not keeping pace — 33 worktrees accumulated, and pruning a merged worktree still leaves its `node_modules` behind until the directory is removed.

## Proposed solution

Automate reclamation so the disk never reaches saturation. Candidate behaviors (to be refined):

1. **Strip `node_modules` from idle worktrees.** For any worktree whose branch has no uncommitted changes and no commits ahead of `origin/main` (or whose PR is merged), delete `node_modules` (and `.turbo`). It is regenerable via `npm install`. This reclaims the bulk without removing the worktree itself.
2. **Prune merged/gone-branch worktrees aggressively.** Remove worktree directories whose upstream branch is merged or deleted on the remote — extend/replace `prune-merged-worktrees.py`.
3. **Threshold trigger, not just weekly cron.** Add a free-space check (e.g. warn at <20 GB, act at <10 GB) so saturation is prevented between weekly runs. Could hook into `UserPromptSubmit` or a more frequent scheduled task.
4. **Optional:** fold in `docker system prune` of dangling images and `npm cache` trimming as low-risk extras.

## Acceptance criteria

- [ ] A scheduled task (or threshold-triggered hook) reclaims regenerable artifacts (`node_modules`, `.turbo`) from idle/merged worktrees automatically.
- [ ] Worktrees with uncommitted or unpushed work are never touched.
- [ ] Free-space headroom is maintained below a saturation threshold between weekly prune runs.
- [ ] README / routine docs updated for the new behavior.

## Context

Discovered while planning lifting-logbook #427; the full disk blocked even writing the plan file. Filed at the user's request to make this a tooling fix rather than a manual cleanup.

brownm09:
## Summary

Automates reclamation of regenerable disk artifacts so `C:` never reaches saturation. Root cause of dev-env#306: dozens of Claude-managed worktrees each carry a full monorepo `node_modules` (~1–2 GB), and the daily merged-worktree prune leaves `node_modules` behind on idle-but-unmerged worktrees. `node_modules`/`.turbo` are fully regenerable — `worktree-npm-install.py` (ADR-016) reinstalls on next use — so stripping them from idle worktrees is safe and self-healing.

Three coordinated artifacts (user-confirmed design: **both** a scheduled routine and a free-space threshold hook; eligibility = clean **AND** merged-or-not-ahead):

- **`claude/scripts/reclaim-worktree-disk.py`** — core script. Strips `node_modules` (top-level + nested monorepo packages) and `.turbo` from worktrees under `.claude/worktrees/` that are not primary/protected, clean, and (merged into `origin/main` **or** zero commits ahead). Flags `--dry-run`/`--repo-path`/`--scan-dir`/`--min-free-gb`/`--protect-cwd`. Reuses the helper structure of `prune-merged-worktrees.py`. Deletes only regenerable dirs — never the worktree, branch, or git state.
- **`claude/scripts/disk-space-check.py`** — `UserPromptSubmit` hook. Warns once/session below 20 GB; below 10 GB spawns the core script **detached** (via `sys.executable`, never the `py` launcher — dev-env#300) so the delete never blocks the prompt. Per-band `session_id`-keyed marker (ADR-027). Wired into `settings.json`.
- **`claude/routines/reclaim-worktree-disk/`** — cron `0 */6 * * *` routine; `sync-routine-worktree` Step 0 (ADR-013); `--scan-dir` mode; push-notifies when ≥1 GB reclaimed.

**Design note:** the eligibility gate is path-based (`.claude/worktrees/`), not branch-prefix-based — reclamation only deletes regenerable artifacts, so what matters is whether the ADR-016 auto-reinstall safety net covers the worktree. This includes Claude worktrees with `feat/*`/`config/*` branches and excludes manual sibling clones.

## Acceptance criteria (dev-env#306)
- [x] Scheduled task + threshold hook reclaim regenerable artifacts from idle worktrees automatically
- [x] Worktrees with uncommitted or unpushed-ahead work are never touched
- [x] Free-space headroom maintained below saturation between weekly prune runs (6-hourly routine + <10 GB hook)
- [x] README / routine docs updated

## Testing
Per `## Testing` in project CLAUDE.md (syntax check) plus new unit tests:

- `py -3 -c "import ast … " claude/scripts/*.py` → **clean** (all scripts syntactically valid)
- `py -3 claude/scripts/tests/test_pyw_stdio.py` → `Tests: 6 passed, 0 skipped, 0 failed` (static guards confirm the new hook imports `_winsubp`, resolves in `settings.json`, and does not spawn via the `py` launcher)
- `py -3 claude/scripts/tests/test_reclaim_worktree_disk.py` → **`Tests: 6 passed, 0 skipped, 0 failed`** (eligibility decision table, `.claude/worktrees/` path gate, reclaim-dir walk incl. nested + `.git` skip, dry-run reports-without-deleting, real reclaim deletes only reclaim dirs)
- Manual dry-run: `reclaim-worktree-disk.py --repo-path … --dry-run` correctly protected the current worktree, skipped dirty trees and manual sibling clones, and selected only idle Claude worktrees.
- Hook smoke test: synthetic `UserPromptSubmit` payload → exit 0 with the expected `systemMessage` (validated the ACT path end-to-end against the live <10 GB condition).

**Coverage gate:** new testable behavior (eligibility logic, reclaim-dir discovery, byte tally, dry-run vs. real deletion) is covered by `test_reclaim_worktree_disk.py`. The git-dependent guards (dirty/primary/merge checks) are exercised by the dry-run.

**Suppression check:** `git diff origin/main | grep -E '(ts-ignore|…)'` → no matches.
**Test-integrity check:** integrity grep → no matches; no skips, deletions, or threshold changes.

## ADR / docs
- **ADR-036** `docs/adr/036-worktree-disk-reclamation.md` + INDEX row (adds a hook + routine + script and a saturation-prevention policy).
- Doc-reconciliation: README (Hooks + Routines tables), `docs/REFERENCE.md` (UserPromptSubmit hook entry, Routines section, Utilities row) updated in this PR.

Closes #306


brownm09:
## Review: Automate disk reclamation for idle worktree node_modules (#306)

### Summary
**Merge with minor changes.** This is a well-scoped, defensively-built feature: the eligibility gate is conservative (clean **AND** merged-or-zero-ahead), reclamation is self-healing via ADR-016, and only regenerable dirs are ever deleted. No blocking correctness, security, or reliability findings. The highest-value improvement is a cheap reordering in `reclaim_one` that avoids a per-worktree network `gh` call in the common case. All findings below are non-blocking.

---

### Non-Blocking Findings

**[performance]** `claude/scripts/reclaim-worktree-disk.py:537-540`
`reclaim_one` computes `merged = is_merged(...)` *before* `ahead = commits_ahead(...)`. `is_merged` falls back to a network `gh pr list` call whenever the branch is not a local ancestor of `origin/main` (the squash-merge case). But `is_idle_eligible` returns `True` as soon as `ahead == 0`, which is the common case for idle worktrees — so the expensive `gh` call is paid even when the cheap local `commits_ahead` check would have short-circuited eligibility. In `--scan-dir` mode across many repos × dozens of worktrees this is dozens of avoidable network round-trips per run.
**Suggestion:** Compute `ahead = commits_ahead(...)` first; if `ahead == 0`, treat as eligible without calling `is_merged`. Only call `is_merged` when `ahead` is non-zero or `None`. Same decision table, fewer network calls.

---

**[reliability]** `claude/scripts/reclaim-worktree-disk.py:384-392, 371-381`
`commits_ahead(branch, repo)` and `is_merged(branch, ...)` run with `cwd=repo` (the primary repo) and interpolate `branch` into the revision range `origin/main..{branch}`. `parse_worktrees` initializes `branch=""` and only overwrites it on a `branch ` or `detached` porcelain line. A `<detached>` worktree is handled safely (bad-revision → `None`/not-merged → skipped). But an empty-string branch (rare: a worktree porcelain entry with neither line) produces `git rev-list --count origin/main..`, where the empty right operand resolves to the **primary repo's HEAD**, not the worktree's — a silent misclassification.
**Suggestion:** Guard explicitly — if `branch` is falsy or `"<detached>"`, skip the worktree before the merge/ahead checks (record a "no resolvable branch" skip reason). Cheap, removes the only path where the ahead-count is computed against the wrong ref.

---

**[reliability]** `claude/scripts/reclaim-worktree-disk.py:473-486` / `claude/scripts/disk-space-check.py:138-163`
The 6-hourly routine and the `< 10 GB` threshold hook can both spawn reclamation concurrently, calling `shutil.rmtree` on the same `node_modules`. The loser gets `OSError` (already-deleted), which is caught — functionally safe. But `dir_size_bytes(d)` is summed *before* the delete, so two overlapping runs can each count the same bytes and over-report the reclaimed total (and the routine's "≥ 1 GB → push-notify" threshold could double-fire).
**Suggestion:** Acceptable to defer given the self-correcting OSError handling, but worth a one-line comment noting the totals are best-effort under concurrent runs. A lightweight lock file in `scratch/` (mirroring the per-session marker pattern) would make totals exact if desired later.

---

**[maintainability]** `claude/scripts/disk-space-check.py:116-130`
Per-session marker files `disk_space_check_<session_id>_<band>.flag` are written to `scratch/` and never cleaned up, so they accumulate one or two files per session indefinitely. This mirrors the existing `reconcile-open-prs.py` sentinel pattern, so it is consistent — but the directory grows unbounded over time.
**Suggestion:** Defer (consistent with existing convention), but consider a periodic sweep of stale `*.flag` files in an existing routine if `scratch/` clutter becomes a concern.

---

**[maintainability]** `claude/scripts/reclaim-worktree-disk.py` (orchestration coverage)
The unit tests cover the pure helpers well (eligibility table, path gate, dir discovery, byte tally, dry-run vs. real delete). The git-dependent orchestration — `reclaim_one`, `is_merged`, `commits_ahead`, `is_dirty` — is exercised only by the manual dry-run, as the PR body honestly notes. This is an accepted deferral, not a gap to block on; flagging only so the reordering fix in the first finding is manually re-verified with a dry-run before merge.
**Suggestion:** After applying the perf reordering, re-run `--dry-run --scan-dir` and confirm idle worktrees are still selected and ahead/dirty/primary are still skipped.

---

### Questions for Author

**Question:** Should the `disk-space-check.py` hook fire reclamation when the *active session's own worktree* is the dominant disk consumer, given `--protect-cwd` shields it?
**Context:** `disk-space-check.py:185` passes `cwd` as `--protect-cwd`, so the current worktree is never stripped. If the active worktree is itself the large consumer and all *other* worktrees are already idle/stripped, the `< 10 GB` spawn may reclaim little or nothing yet still mark the band fired for the session.
**Tradeoffs:** (a) Keep current behavior — never strip the active worktree (correct: it is in use, and stripping it would trigger an immediate ADR-016 reinstall); the warning still surfaces so the user can prune manually. (b) Additionally emit the reclaimed total back to the session so a near-zero reclaim makes the "manual prune" guidance actionable — small UX gain, but the detached spawn discards stdout by design. Current behavior is the right default; raising only to confirm the no-op-spawn case is intended.

---

### Style Notes
Clean and consistent with the existing `prune-merged-worktrees.py` conventions — type hints, docstrings explaining the *why*, named threshold constants with rationale comments, and `_winsubp` imported first per the hook contract. No style concerns.

---

*Reviewed by `claude-opus-4` via Claude Code*

Issue #314, opened 2026-06-04
[docs] Add an index/pointer for claude/hooks tests


brownm09:
Follow-up from the review of #310 (non-blocking, ADR-028 fix-or-file → filed as tracked).

`claude/hooks/tests/` is the first test dir under `claude/hooks/`, mirroring `claude/scripts/tests/`. There is no index tying them together, so hook tests are not discoverable.

**Deferred by design:** the finding's own recommendation is to act "when a second hook test appears." Tracking here so it isn't lost; when the next hook test lands, add a one-line pointer in `docs/REFERENCE.md` Hooks section (and/or a short `claude/hooks/tests/README.md`).

PR #310, opened 2026-06-04
[test] Add behavioral self-test for the pre-push lockfile-drift guard

brownm09:
## Summary

Implements the follow-up deferred from the review of #307: a behavioral regression test for the lockfile-drift guard in `claude/hooks/pre-push` (added in #307 / ADR-036), which previously had only manual verification — unlike the Python hooks, which `test_pyw_stdio.py` guards.

`claude/hooks/tests/test-pre-push-lockfile.sh` drives the **real** hook against throwaway fixture git repos with a **stubbed `npm`** (no network, no real install), asserting:

| Scenario | Expected |
|---|---|
| Drifted lockfile | BLOCK (exit 1), working tree restored, no stray backup |
| In-sync lockfile | PASS (exit 0), tree clean, no stray backup |
| `npm` exits non-zero | SKIP (exit 0, warn — not blocked on tooling failure) |
| `npm` absent from PATH | SKIP (exit 0, guard skipped cleanly) |
| Repo-level `.git/hooks/pre-push` present | chained hook fires on the pass path |

The fixtures and stub bin live outside any repo (mktemp), so a clean working-tree assertion genuinely reflects the hook's restore behavior. Portable to Git Bash (Windows) and Linux CI.

It also doubles as a regression test for the #307 review hardening (trap-based restore, backup outside the repo): scenario 2 fails if the lockfile is not restored after a block.

## Testing

- `bash claude/hooks/tests/test-pre-push-lockfile.sh` → **Results: 12 passed, 0 failed** (exit 0).
- `bash -n` on the test → clean; `py -3 ast.parse claude/scripts/*.py` → OK (no Python changed).
- Suppression / test-integrity greps on the diff → **(none)**.

(During development the test initially reported 3 failures — traced to the stub `npm` bin being created *inside* the fixture repo, which dirtied `git status`; moving it to a separate mktemp dir resolved it. Not a hook defect.)

## Docs

- Wires the test into the `## Testing` section of `claude/CLAUDE.md`.
- Updates `docs/REFERENCE.md` pre-push section to document the lockfile-drift guard and draft-branch-block behaviors (which #307 added to `README.md` but not `REFERENCE.md`) plus the new self-test.

No new ADR — covered by [ADR-036](https://github.com/brownm09/dev-env/blob/main/docs/adr/036-lockfile-drift-prevention.md), whose Alternatives section named this self-test as deferred follow-up.

Closes #308

PR #316, opened 2026-06-04
[test] Guarantee absent-npm coverage in pre-push lockfile self-test

### LINKED ISSUE #313: [test] Guarantee absent-npm coverage in pre-push lockfile self-test
Follow-up from the review of #310 (non-blocking, ADR-028 fix-or-file).

Scenario 4 (npm absent) in `claude/hooks/tests/test-pre-push-lockfile.sh` rebuilds PATH by removing only npm's directory, then **skips with a NOTE** if it cannot isolate npm (e.g. npm shares a dir with coreutils). On such a host the scenario silently degrades to a no-op.

**Fix:** point PATH at a purpose-built temp dir that symlinks (or copies) only the tools the hook needs — `git`, `cp`, `rm`, `diff`, `grep`, `mktemp`, `bash` — so the absent-npm path is always exercised. Keep the skip-NOTE only as a last-resort fallback if a needed tool can't be located.

brownm09:
## Summary

Retro-fix bringing #310 into ADR-028 compliance — addresses the two non-blocking findings the #310 review left 'as-is' (which it should not have, per the all-findings merge gate):

- **[reliability] absent-npm isolation** — *fixed here.* Scenario 4 of `claude/hooks/tests/test-pre-push-lockfile.sh` previously skipped with a NOTE when it couldn't isolate npm from PATH, silently becoming a no-op. It now constructs an npm-absent PATH (drops npm's dir; shims any tool that lived only there) and **fails loudly** if it can't, rather than skipping. Coverage is guaranteed on normal hosts (Windows dev + Linux CI). Removed the now-unused `note()` helper.
- **[maintainability] hook-test discoverability** — *filed as #314* (deferred by design: its own recommendation is to act 'when a second hook test appears').

## Testing

- `bash claude/hooks/tests/test-pre-push-lockfile.sh` → **Results: 12 passed, 0 failed**. Scenario 4 now reports `exit 0 (guard skipped cleanly, npm absent)` — the assertion runs instead of skipping.
- `bash -n` clean; suppression/integrity greps → (none).

This PR's own review findings will each be dispositioned (fixed-in-PR or filed) before merge — no 'as-is'.

Closes #313

### DIFF
--- a/claude/hooks/tests/test-pre-push-lockfile.sh
+++ b/claude/hooks/tests/test-pre-push-lockfile.sh
@@ -20,7 +20,6 @@ ZERO="0000000000000000000000000000000000000000"
 
 PASS=0
 FAIL=0
-note() { echo "  NOTE: $*"; }
 ok()   { echo "  ok: $*"; PASS=$((PASS + 1)); }
 bad()  { echo "  FAIL: $*"; FAIL=$((FAIL + 1)); }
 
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [test] Guarantee absent-npm coverage in pre-push lockfile self-test

### Summary
**Merge.** Scenario 4 now genuinely exercises the npm-absent path instead of skipping, and a construction failure is a loud `bad` rather than a silent no-op — exactly what the #310 finding asked for. The degradation path is safe: if the shim can't restore a needed tool, the hook run fails and the scenario reports a failure (not a false pass). Dead `note()` helper removed. No blocking or non-blocking findings.

### Review findings disposition
- None raised — nothing met the fix-or-file bar.

PR #326, opened 2026-06-06
Add 'no terminal-spawn in Windows scripts' rule to global CLAUDE.md (ADR-041)

brownm09:
## Summary

- Add a new `### No spawning new terminal windows (Windows scripts)` subsection to global `claude/CLAUDE.md` under `## Code Quality`, immediately after `### Fix errors on encounter`. Names the four pattern shapes (`Start-Process -Verb RunAs` self-relaunch, `-NoExit` sibling consoles, `cmd /c start ... powershell ...`, equivalent round-trips), points at `#Requires -RunAsAdministrator` and inline `IsInRole` as the right alternatives, captures the closed allowlist of three pre-existing files, and makes `/review` enforcement explicit and blocking.
- ADR-041 captures the full rationale, the incident sequence (user fixed this once yesterday and reminded me again today after I let `win11-init-tools#19` land `configure_dev_env.ps1` with the self-relaunch pattern unflagged), and the rejected alternatives.
- INDEX.md updated with the new row and full Tags column.

## Why this addition belongs in global CLAUDE.md and not the win11-init-tools project CLAUDE.md

Per [ADR-040](../docs/adr/040-global-claudemd-layering-and-slimming.md), project-specific content belongs in the project file; the global file retains only genuinely cross-project content. The rule here is **not** win11-init-tools-specific — it applies identically to any PowerShell or batch script in any project (scheduled-tasks routines, lifting-logbook helpers, ad-hoc fixes in scratch). Global placement means it loads in every Windows-context session, not only when the user is editing win11-init-tools.

The win11-init-tools project CLAUDE.md already had a sentence preferring `#Requires`, but it sits next to a sentence noting the self-relaunch pattern is "kept for end-user convenience" — and I read that as covering new additions too. The closed allowlist in the new global rule removes that ambiguity.

## File-size budget

Global CLAUDE.md: **38,997 → 40,624 chars (+1,627)**, still under the ~41,000-char budget set by ADR-040. The new subsection is ~1.4k chars by design; the verbose rationale lives in the ADR.

Closes #324

## Testing

Per `## Testing` in the dev-env project CLAUDE.md:

- **Hook-script syntax check** (`py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py`): **OK** — all hook scripts parse.
- **`pyw -3` stdio verification**: **N/A** — this PR does not touch `pythonw.exe` paths or stdio behavior.
[...]

PR #327, opened 2026-06-06
Add Pass 3 risk-dimension audit + per-project Observability section (ADR-042)

### LINKED ISSUE #325: Plans don't force consideration of observability, security, resilience, performance, or data-integrity
## Problem

Implementation plans reliably consider **testing** because it's gated three ways (the per-project `## Testing` section requirement, the ADR-022 coverage gate, and the suppression/integrity/baseline policies). But nothing forces a plan to consider the other dimensions where changes commonly go wrong:

- **Observability** — what gets logged/traced at boundaries and on failure
- **Security** — authz, input validation, secret handling, sensitive-data/PII exposure
- **Resilience / failure modes** — error handling, timeouts, fallbacks, rollback
- **Performance** — N+1, hot paths, payload/bundle size
- **Data integrity & migrations** — schema changes, multi-tenant isolation, reversibility

The gap is concrete. In lifting-logbook the LLM cycle-generation path has no timeout/circuit-breaker, `$queryRaw` calls aren't auto-traced, LLM adapters have no PII scrubbing, and Prisma migrations have no documented rollback — none caught by the testing gate. Conversely, career-playbook (a content/docs repo) would be forced into meaningless answers by a flat security/perf checklist; its real gates are reference integrity and briefing regeneration.

## Goal

1. Add a universal, tiered **plan risk-dimension audit** to the plan-then-optimize protocol in global CLAUDE.md (six mandatory dimensions with an explicit "N/A — <reason>" escape; accessibility conditional; project-specific gates deferred to each project's CLAUDE.md).
2. Add a mandatory per-project `## Observability` section requirement, mirroring the existing `## Testing` mechanism, with **advisory** enforcement (no new hook).
3. Record the decision in ADR-041.

Enforcement is advisory to stay consistent with how `## Testing` works today and avoid false PR blocks.

brownm09:
## Summary

Plans were gated on **testing** three ways (the `## Testing` section, the ADR-022 coverage gate, the suppression/integrity/baseline policies) but nothing forced a plan to consider **observability, security, resilience, performance, or data-integrity/migrations**. This adds a forcing function for those dimensions.

A two-project audit supplied the motivating evidence and the key constraint — the two have near-inverse risk profiles, so a flat checklist is wrong for both:
- **lifting-logbook** (full-stack app): LLM path has no timeout/circuit-breaker; `$queryRaw` not auto-traced; LLM adapters have no PII scrubbing; migrations lack a rollback strategy — none caught by the testing gate.
- **career-playbook** (content/docs repo): security/perf/observability largely N/A; real gates are `validate.sh` and briefing regeneration.

> **Note:** originally authored as ADR-041; renumbered to **ADR-042** after PR #326 (opened 42s earlier) claimed 041 concurrently.

## Changes

- **`claude/CLAUDE.md` — Pass 3 (Risk-dimension audit)** added to the plan-then-optimize protocol: six mandatory dimensions (Testing, Observability, Security, Resilience/failure-modes, Performance, Data-integrity/migrations), each "address or `N/A — <reason>`". Accessibility is conditional (UI-touching changes); project-specific gates are deferred to each project's CLAUDE.md. Intro updated "two passes" → "three".
- **`claude/CLAUDE.md` — mandatory per-project `## Observability` section** added to `## Per-Project CLAUDE.md Requirements`, mirroring `## Testing`. **Advisory** enforcement (CLAUDE.md rule + existing pre-PR reminder; no new hook).
- **`docs/adr/042-...md` + INDEX row** recording the decision, the inverse-risk-profile evidence, and the rejected alternatives (flat checklist, blocking hook, mandatory a11y, folding into Pass 2).

## Testing

This is a docs/config-only change (no scripts, no hooks touched), so the relevant dev-env `## Testing` items:

- **Docs-only guard** (`grep -n 'date -u' claude/CLAUDE.md`): no matches. ✅
- **Hook-script syntax check**: N/A — no `claude/scripts/*.py` changed.
- **Render check**: `docs/adr/INDEX.md` new row well-formed; ADR-042 file links resolve. ✅
- **Suppression grep**: clean. ✅
- **Test-integrity grep**: clean. ✅
- **Lockfile sync**: N/A — no `package.json` changed.

Tests: N/A (docs/config-only) — 0 passed, 0 skipped, 0 failed.

## Coverage gate

N/A — no testable behavior added; process-documentation change only.

## Doc reconciliation

No README.md / docs/REFERENCE.md update required — the Documentation Maintenance table covers skill/hook/routine/config-schema changes, not global CLAUDE.md rule additions.

## Merge note

Shares `docs/adr/INDEX.md` with concurrent PR #326. Whichever merges second resolves the INDEX conflict by keeping both the 041 and 042 rows.

## Follow-ups (filed)

Per-project `## Observability` sections, tracked in their own repos:
- brownm09/lifting-logbook#456
- brownm09/career-playbook#336

Closes #325

`/review` (claude-opus-4-8) posted [here](https://github.com/brownm09/dev-env/pull/327#issuecomment-4639277940) — 1 blocking, 1 non-blocking, 1 style.

- **[blocking · documentation] dev-env self-non-compliance** — **fixed in `7634938`.** Added a `## Observability` section to dev-env's project `CLAUDE.md` (it had `## Testing` but no `## Observability`, so it would have failed the rule this PR introduces). Documents the no-runtime model and defers Pass 3's Observability check to the existing hook/script verification suite (ADR-007/ADR-027).
- **[non-blocking · maintainability] INDEX 040→042 numbering gap** — **no change; verified at merge.** Correct given #326 claims 041 concurrently. #326 confirmed still OPEN at merge time, so #327 merges cleanly into main; #326 resolves the INDEX conflict (keep both 041 + 042 rows) when it merges second. Only actionable if #326 is abandoned, in which case renumber or add a withdrawn-placeholder row.
- **[style] `N/A — <reason>` placeholder inconsistency** — **fixed in `7634938`.** Unified `claude/CLAUDE.md` (was escaped `\<reason\>`) and ADR-042 §1 (was unescaped, vanishes as HTML tag on GitHub) to backtick form `` `N/A — <reason>` `` so it renders literally in both raw-text and HTML contexts.


brownm09:
## Review: Add Pass 3 risk-dimension audit + per-project Observability section (ADR-042)

### Summary
**Merge with one fix.** The change is clean, well-scoped docs/config work — the wording is precise, the ADR thoroughly records the inverse-risk-profile reasoning and rejected alternatives, and all pre-PR gates are correctly N/A or clean. The single blocking finding is self-referential: the PR makes a `## Observability` section **mandatory for every project CLAUDE.md**, but dev-env's own project `CLAUDE.md` doesn't get one in this PR, so the repo is non-compliant with its own new rule the moment it merges.

---

### Blocking Findings

**[documentation]** `CLAUDE.md` (dev-env project file, repo root)
The new rule in `claude/CLAUDE.md` states *"Every project CLAUDE.md **must** also include a `## Observability` section."* dev-env is itself a project with a project-level `CLAUDE.md`, and it has a `## Testing` section but no `## Observability` section. Follow-ups were filed for lifting-logbook (#456) and career-playbook (#336), but dev-env — the repo introducing the rule — was overlooked. On merge, dev-env instantly fails the rule it ships, which undercuts the rule's authority and is the kind of inconsistency the next reader will trip on.
**Fix:** Add a `## Observability` section to dev-env's project `CLAUDE.md` in this same PR (trivially in-scope — same repo). dev-env has no long-running runtime to instrument; the section should say so and name the equivalent verification (hook-script syntax check via `ast.parse`, the `pyw -3` stdio test, the pre-push lockfile self-test, and hook stderr/exit-code conventions per ADR-027).

---

### Non-Blocking Findings

**[maintainability]** `docs/adr/INDEX.md:48`
The INDEX jumps `040 -> 042`. This is correct **given** concurrent PR #326 claims 041 — but the integrity of the sequence is now contingent on an external PR merging. If #326 is closed without merging, ADR-041 never exists and the INDEX permanently skips a number with no explanation in-repo (the rationale lives only in this PR's body, which isn't visible from the INDEX).
**Suggestion:** No change needed if #326 merges (the expected path). If #326 is abandoned, either renumber this ADR back to 041 at that time, or add a one-line "041 — withdrawn (see #326)" placeholder row so the gap is self-documenting. Worth a glance at merge time to confirm #326's status.

---

### Questions for Author

**Question:** Should the dev-env `## Observability` follow-up be handled inline here, or do you want it tracked separately like the other two projects?
**Context:** #456 and #336 defer the lifting-logbook/career-playbook sections to their own repos because those are cross-repo changes. The dev-env section is *same-repo*, so it can be fixed in this PR with no extra coordination — unlike the other two.
**Tradeoffs:** (a) Fix inline now — closes the self-non-compliance window entirely, keeps the rule and its first compliant example in one atomic change; (b) File a dev-env follow-up issue — consistent with how #456/#336 were handled, but leaves dev-env knowingly non-compliant between merge and the follow-up, for no real isolation benefit since it's the same repo.

---

### Style Notes
Minor escaping inconsistency: `claude/CLAUDE.md` writes the placeholder as `"N/A — \<reason\>"` (backslash-escaped angle brackets) while ADR-042's Decision section 1 writes `"N/A — <reason>"` (unescaped). CLAUDE.md is consumed as raw text by the agent, so the literal backslashes will appear in-context; the ADR renders as HTML on GitHub where `<reason>` would vanish as an unknown tag. Each file's choice is locally defensible, but they disagree — pick one convention (e.g. backticks `N/A — <reason>`) that reads cleanly in both contexts.

---

### Findings Disposition
Recorded in the PR body before merge (per ADR-028/ADR-039): the blocking finding fixed in this PR; the non-blocking INDEX item verified at merge; the style nit fixed in this PR.

---

PR #329, opened 2026-06-07
Block writes from orphaned Claude-managed worktrees (ADR-024 addendum) (#328)

### LINKED ISSUE #328: Harden against orphaned Claude-managed worktrees silently resolving to the canonical repo
## Symptom

An **orphaned Claude-managed worktree** — a directory under `<repo>/.claude/worktrees/<name>/` that still exists on disk but is **not registered** in `git worktree list` and is **missing its `.git` link file** — silently causes every git command run from that cwd to resolve *up the tree* to the **canonical repo's** `.git`. The harness still treats the directory as the session's worktree (it sets cwd there and force-resets cwd to it after every command), so the failure is invisible until something goes wrong.

Observed in a career-playbook session (2026-06-06): the designated worktree `.claude/worktrees/youthful-williams-ff1a17` was orphaned. As a result, before it was noticed:
- `git status` / `git branch` / `git stash` / `git checkout -b` all operated on the **canonical** checkout, not the worktree.
- `git stash -u` stashed the canonical checkout's in-progress WIP (a different branch's uncommitted work); `git checkout -b <newbranch> origin/main` switched the canonical checkout onto the new branch.
- `Write`/`Edit` (which take absolute paths) landed new files in the disconnected directory, invisible to git.

Recovery took several careful steps (restore canonical to its prior branch, `git stash pop` the WIP, `git worktree add --force` to re-create the designated path as a real worktree, move the stranded files back). Full trace: engineering-journal `sessions/career-playbook/2026-06-06_105718.stub.md` (Decisions/notes), recorded as a meta-journal candidate.

## Root cause

1. Nothing verifies a worktree's `.git` link resolves before work begins in it.
2. `claude/scripts/pre-tool-use-worktree-path-check.py` keys on the **cwd's** `.claude/worktrees/<name>` prefix and trusts it to be a *real* worktree. For an orphaned dir, the hook computes `worktree_root` from the path string, sees the (mis-resolved) file path is "inside" it, and **passes** — so it does not catch the exact case it would most want to (writes silently hitting the canonical tree).

## Proposed hardening (implementer's choice)

- **Option A — extend `pre-tool-use-worktree-path-check.py`:** when cwd is under `.claude/worktrees/<name>`, additionally assert the worktree is live before allowing the write — e.g. `git -C <cwd> rev-parse --show-toplevel` equals the worktree root **and** `<cwd>/.git` exists. If not, **block** with an "orphaned/disconnected worktree — git is resolving to the canonical repo" message plus the recovery recipe (`git worktree add --force <path> <branch>`), instead of silently allowing a canonical-tree write.
- **Option B — session-start guard:** a `SessionStart` hook that, when cwd is a `.claude/worktrees/<name>` path, verifies `git rev-parse --git-dir` resolves to `<canonical>/.git/worktrees/<name>` (not the bare canonical `.git`) and that the dir is in `git worktree list`; emit a loud warning/halt otherwise.
- **Option C — tie into existing worktree tooling** (`claude/scripts/multi-worktree-alert.py`, `prune-merged-worktrees.py`) which already reason about worktree registration, so orphans are surfaced/cleaned proactively.

Option A is the highest-leverage (it converts the silent-write-to-wrong-repo failure into a hard block at the moment of risk). A is also testable in the same hermetic style as the existing hook tests.

## Acceptance

- An Edit/Write issued from an orphaned `.claude/worktrees/<name>` cwd is blocked (or session-start halts) with an actionable message, rather than silently writing to the canonical tree or a disconnected directory.
- Covered by a hook self-test (mirroring the existing `claude/scripts/tests/` pattern).

brownm09:
## Summary

Hardens the worktree-path guard against **orphaned Claude-managed worktrees** — a `<repo>/.claude/worktrees/<name>/` directory that still exists on disk but is no longer a registered git worktree (its `.git` link is gone). Git then resolves every command *up the tree* to the canonical repo, so writes silently land on the wrong tree or in a directory invisible to git, while the harness keeps treating the dead dir as the active worktree.

The existing `pre-tool-use-worktree-path-check.py` keyed only on the cwd path *string*, computed the worktree root from it, saw the target was "inside" it, and **passed** — missing the exact case it most needed to catch.

Closes #328.

## Changes

- **`claude/scripts/pre-tool-use-worktree-path-check.py`** — adds a liveness guard that runs *before* the existing path-scoping (so it covers relative and in-worktree absolute paths, not just canonical-root paths):
  - `_worktree_is_live()` — (1) `<worktree_root>/.git` must exist (orphan signature, caught without spawning git); (2) `git -C <cwd> rev-parse --show-toplevel` must equal the worktree root (catches mis-resolution to canonical); (3) if git can't run but `.git` is present → treated as live (no false-block on transient git failure).
  - Not-live → exit 2 with `{"reason"}` naming the worktree + cwd and the recovery recipe `git worktree add --force <worktree_root> <branch>`.
  - Adds `import _winsubp` (now spawns `git` under `pythonw` — ADR-007 console-flash suppression).
- **`claude/scripts/tests/test_worktree_path_check.py`** — new hermetic self-test (unit decision table + subprocess integration).
- **Docs:** README hooks table/intro, `docs/REFERENCE.md` hook entry, **ADR-024 addendum (2026-06-06)**, `docs/adr/INDEX.md` row.

## Testing

Per dev-env `## Testing`:

- **Hook-script syntax check** (`ast.parse` over `claude/scripts/*.py`) — OK.
- **New self-test** `py -3 claude/scripts/tests/test_worktree_path_check.py` — `Tests: 4 passed, 0 skipped, 0 failed`. Covers the liveness decision table (6 combinations), `.git`-missing short-circuit, end-to-end block of an Edit from an orphaned worktree cwd (exit 2 + recovery text — the acceptance criterion), and non-worktree no-op (exit 0).
- **`pyw -3` stdio test** `py -3 claude/scripts/tests/test_pyw_stdio.py` — `Tests: 6 passed, 0 skipped, 0 failed` (confirms the new `_winsubp` import keeps every hook on the `pyw -3` contract; "every subprocess-using hook imports `_winsubp`" now counts this script).

The live-worktree path is also exercised implicitly: this PR was authored from a live worktree and all Edit/Write calls succeeded.

**Outcome:** all green.

## Coverage / integrity gates

- **Coverage gate:** new behavior covered by the new self-test (above).
- **Suppression grep:** clean (no suppressions added).
- **Test-integrity grep:** the only matches are the `xit(` false-positive class matching `e**xit(**` in the hook's normal `sys.exit(...)` calls — no skip markers, no deleted tests, no lowered thresholds.
- **Lockfile / baseline:** N/A — no `package.json` change; dev-env is not a Jest repo.

## Risk audit

- **Resilience:** fail-closed (block, recoverable with recipe) on a genuine orphan; fail-open only on transient git failure when the `.git` link is present.
- **Performance:** one `git rev-parse` per file write *in a worktree only*, short-circuited (no subprocess) when `.git` is already missing; no-op outside worktrees unchanged.
- **Security:** `git` invoked with list args (no shell); cwd from the harness; no new input surface.

`/review` run in-session (0 blocking, 1 non-blocking, 1 question) — [comment](https://github.com/brownm09/dev-env/pull/329#issuecomment-4641279583), `reviewed-by-claude` applied. All findings dispositioned:

- **[non-blocking, maintainability] Python language-scope scanner gap** — verified, no change. The new `.py` test was manually confirmed to contain no skip idioms (`@pytest.mark.skip`, `pytest.skip(`, etc.); ADR-029's JS/TS-only scanners simply can't inspect it.
- **[question] per-write `git rev-parse` cost** — fixed in `9925b41`. Added an ADR-024 judgment-call bullet documenting that signal 2 is not redundant with the `.git`-existence check: it catches a distinct orphan mode (the `.git` file exists but its `gitdir:` target was pruned), so the per-edit windowless git spawn is a deliberate, justified cost; per-session memoization was considered and rejected to keep the hook stateless.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Custom slash commands loaded from `claude/skills/`. Invoke with `/skill-name [ar
 ## Hooks
 
 Hook scripts run automatically via Claude Code's `hooks` configuration in `claude/settings.json`.
-Most hooks are advisory — they emit reminders but do not block tool execution. The exception is `pre-tool-use-worktree-path-check.py`, which blocks `Write`, `Edit`, and `NotebookEdit` calls that target the canonical repo root instead of the active worktree.
+Most hooks are advisory — they emit reminders but do not block tool execution. The exception is `pre-tool-use-worktree-path-check.py`, which blocks `Write`, `Edit`, and `NotebookEdit` calls that target the canonical repo root instead of the active worktree, **and** blocks any such call issued from an orphaned/disconnected worktree (one whose `.git` link is gone, so git silently resolves to the canonical repo).
 
 Hooks that spawn subprocesses (`git`, `gh`, `bash`, …) must `import _winsubp` — a helper module at `claude/scripts/_winsubp.py` that patches `subprocess.Popen.__init__` to set `CREATE_NO_WINDOW` so children don't flash a console window under `pythonw.exe`. See [ADR-007](docs/adr/007-hook-command-invocation.md).
 
@@ -58,7 +58,7 @@ Hooks that spawn subprocesses (`git`, `gh`, `bash`, …) must `import _winsubp` 
 | PreToolUse (Bash) | `pre-commit-branch-check.py` | Emits current branch as a checkpoint before `git commit` |
 | PreToolUse (Bash) | `pre-pr-create-check.py` | Emits test-verification checklist, documentation-gap warning, and pre-existing-failure baseline advisory before `gh pr create` |
 | PreToolUse (Bash) | `pre-merge-findings-gate.py` | Blocks `gh pr merge` when a `/review` comment reports open findings and the PR body records no disposition — mechanical enforcement of the all-findings merge gate (see [ADR-039](docs/adr/039-merge-gate-findings-enforcement.md)) |
-| PreToolUse (Write/Edit/NotebookEdit) | `pre-tool-use-worktree-path-check.py` | Blocks file writes whose absolute path targets the canonical repo root instead of the active worktree |
+| PreToolUse (Write/Edit/NotebookEdit) | `pre-tool-use-worktree-path-check.py` | Blocks file writes whose absolute path targets the canonical repo root instead of the active worktree, and blocks all writes from an orphaned worktree whose `.git` link no longer resolves (see [ADR-024](docs/adr/024-worktree-path-guard-hook.md)) |
 | PostToolUse (Bash) | `pr-merge-reminder.py` | Reminds to write a journal stub after `gh pr create`, `gh pr merge`, or `git push` (when the pushed branch has an open PR) |
 | PostToolUse (Bash) | `post-tool-use.py` | Auto-adds issues/PRs to configured GitHub Project; exits 2 with `required_fields` reminders |
 | PostToolUse (Bash) | `post-pr-merge-pull.py` | Fast-forwards local `main` after `gh pr merge` |
--- a/claude/scripts/pre-tool-use-worktree-path-check.py
+++ b/claude/scripts/pre-tool-use-worktree-path-check.py
@@ -10,11 +10,18 @@ silently. This hook intercepts those calls before the write happens.
 Logic:
   1. If cwd does not contain `/.claude/worktrees/<name>/`, pass immediately.
   2. Extract canonical_root (repo root) and worktree_root from cwd.
-  3. Read file_path (Write/Edit) or notebook_path (NotebookEdit) from tool input.
-  4. If the path is absolute and starts with canonical_root but NOT with
+  3. Liveness guard (ADR-024 addendum, dev-env#328): assert the worktree is a
+     *live* registered worktree, not an orphan whose `.git` link is gone. An
+     orphaned worktree dir silently resolves every git command up the tree to
+     the canonical repo's `.git`, so writes land on the wrong tree or in a
+     disconnected directory invisible to git. If `<worktree_root>/.git` is
+     missing, or `git -C <cwd> rev-parse --show-toplevel` resolves to anything
+     other than worktree_root → exit 2 with the recovery recipe.
+  4. Read file_path (Write/Edit) or notebook_path (NotebookEdit) from tool input.
+  5. If the path is absolute and starts with canonical_root but NOT with
      worktree_root → exit 2 with a blocking message naming both paths and the
      corrected worktree-relative path.
-  5. Otherwise pass (exit 0).
+  6. Otherwise pass (exit 0).
 
 Blocking: exits 2 with JSON {"reason": "..."} — the harness refuses the tool
 call and shows the reason to Claude so it can re-issue with the correct path.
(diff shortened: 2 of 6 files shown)


brownm09:
## Review: Block writes from orphaned Claude-managed worktrees (ADR-024 addendum) (#328)

### Summary
**Merge with minor consideration.** Clean, well-scoped change (351 lines, 6 files) that closes a real silent-failure class with a testable, dependency-injected liveness guard. No blocking findings. The implementation correctly handles the Windows path-format mismatch (git emits forward slashes; `_normalize` reconciles via `normpath`+`normcase`), uses `os.path.exists` so the `.git` *file* link is detected (not just a dir), checks `<worktree_root>/.git` rather than `<cwd>/.git` so nested-cwd sessions still resolve, and short-circuits before spawning git when `.git` is already missing. Doc-reconciliation and test-coverage gates are satisfied. The one thing worth a second look is the per-edit git-spawn cost in the common (live) case — raised as a question below.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/tests/test_worktree_path_check.py`
The diff adds a Python test file, but Step 2e's automated integrity scanners target JS/TS frameworks (Jest/Vitest/Mocha) — they cannot inspect pytest/`unittest` skip idioms (`@pytest.mark.skip`, `pytest.skip(`, etc.). A clean JS-pattern grep is not evidence the integrity policy held for this Python code.
This matters here only weakly: the new test is a plain `main()`-dispatch script with no skip markers, and I manually confirmed it introduces none.
**Suggestion:** None required — recorded for traceability. If future test edits add conditional skips, justify them in the PR body per ADR-029.

---

### Questions for Author

**Question:** Is the per-file-write `git rev-parse --show-toplevel` spawn worth keeping for the *live* (common) case, or should the guard rely on the `.git`-existence check alone?
**Context:** In `_worktree_is_live` ([claude/scripts/pre-tool-use-worktree-path-check.py:108](claude/scripts/pre-tool-use-worktree-path-check.py)), a live worktree passes signal 1 (`.git` exists) and then *always* spawns git for signal 2. Since the guard now runs before path-scoping, every `Write`/`Edit`/`NotebookEdit` in a worktree pays one git subprocess — and edit-heavy sessions issue many. The documented incident (career-playbook 2026-06-06) was an orphan with a *missing* `.git`, which signal 1 already catches with zero subprocess cost.
**Tradeoffs:** (a) Keep both signals (current) — strongest detection, also catches the subtler "`.git` present but git resolves to canonical" case, at the cost of one fast (~10–30 ms, windowless) git spawn per write in every live worktree. (b) Drop signal 2, rely on `.git` existence only — zero subprocess cost and still catches the actual documented failure mode, but misses the theoretical mis-resolution-with-`.git`-present case. (c) Keep both but memoize liveness per session via a marker file keyed on cwd — preserves detection, pays the cost once, but adds state to a currently stateless hook. The current choice (a) is defensible for a safety guard; flagging so the cost is a conscious decision rather than incidental.

---

### Findings Disposition
- **[maintainability] Python language-scope scanner gap** — disposition recorded below.
- **[question] per-edit git-spawn cost** — disposition recorded below.

---

PR #335, opened 2026-06-07
Fix get-project-item.sh scratch-path resolution on Windows + Git Bash (#334)

### LINKED ISSUE #334: get-project-item.sh broken on Windows: $HOME scratch path resolves differently in Git Bash vs Node
@
## Problem

`bash claude/scripts/get-project-item.sh 330` exits 1 with:

```
Error: ENOENT: no such file or directory, open 'C:\c\Users\brown\.claude\scratch\tmp_project_item_<pid>.json'
```

## Root cause

Line 26 sets `SCRATCH="${HOME}/.claude/scratch"`. Under Git Bash, `$HOME` is the MSYS path `/c/Users/brown`. The `gh ... > "$TMPFILE"` redirect (line 32) is performed by Git Bash, which maps `/c/Users/...` to `C:\Users\...` correctly. But the `node -e` call (lines 34–39) reads the same `$TMPFILE` string `/c/Users/brown/.claude/scratch/...`, and Node on Windows resolves a leading `/c/...` against the current drive, producing `C:\c\Users\...` — a different, nonexistent path. The write and read target different files → ENOENT.

## Secondary bug

Line 40 `EXIT_CODE=$?` is dead under `set -e` (line 15): if the `node -e` on line 34 exits non-zero, the script aborts before reaching line 40, so the `if [[ $EXIT_CODE -ne 0 ]]` guard never runs as intended.

## Fix direction

Make the scratch path one that both Git Bash and Node resolve identically. Repo convention (global + project CLAUDE.md, `DEFAULT_LOG` in `claude/scripts/session-mode-report.py`) is the literal Windows-style path `C:/Users/brown/.claude/scratch`. Hardcode that, and replace the dead `EXIT_CODE` guard with a `trap … EXIT` cleanup so `set -e` propagates the node failure correctly.

Verify with `bash claude/scripts/get-project-item.sh 330` returning a `PVTI_...` id.
@

brownm09:
## Summary

Fixes `claude/scripts/get-project-item.sh`, which exited 1 with `ENOENT … C:\c\Users\brown\.claude\scratch\…` on every invocation on Windows + Git Bash.

**Root cause.** Line 26 set `SCRATCH="${HOME}/.claude/scratch"`. Under Git Bash `$HOME` is the MSYS path `/c/Users/brown`. The `gh … > "$TMPFILE"` redirect is performed by Git Bash, which maps `/c/Users/…` → `C:\Users\…` correctly. But the `node -e` read of the same `$TMPFILE` string ran under Node-on-Windows, which resolves a leading `/c/…` against the *current drive* → `C:\c\Users\…`. Write and read targeted different files → ENOENT.

**Fix.** Hardcode the literal Windows-style path `C:/Users/brown/.claude/scratch`, which Git Bash and Node resolve identically. This matches the existing repo convention — `DEFAULT_LOG` in `claude/scripts/session-mode-report.py` and the scratch-dir guidance in global + project `CLAUDE.md`.

**Secondary fix.** The old `EXIT_CODE=$?` guard (line 40) was dead code: under `set -e`, a non-zero `node -e` exit aborts the script before the assignment, so the `if [[ $EXIT_CODE -ne 0 ]]` block never ran. Replaced it with a `trap 'rm -f "$TMPFILE"' EXIT` so the temp file is cleaned up on *every* exit path, and a no-match case still exits 1 via `set -e` propagating node's exit code.

Closes #334

## Testing

Per dev-env `## Testing`. This is a `.sh` utility with no unit-test harness; verification is the syntax checks plus the live repro (success + failure paths).

- **`bash -n claude/scripts/get-project-item.sh`** → PARSE OK
- **`py -3 -c "import ast … ast.parse …" claude/scripts/*.py`** (hook-script syntax check) → AST OK (unaffected; .py scripts untouched)
- **Live repro — success:** `bash claude/scripts/get-project-item.sh 330` → `PVTI_lAHOAjEKvM4BWKFezgu8RAQ` (previously ENOENT exit 1) ✅
- **Live repro — failure path:** `bash claude/scripts/get-project-item.sh 99999999` → stderr `get-project-item: no item found for #99999999`, **exit 1** ✅
- **Temp-file cleanup:** scratch-dir temp-file count held steady across both success and failure runs (the `trap` removes the file on every exit). Removed two stale orphan temp files left by the *original* broken script (08:02 today), which are exactly the bug's symptom — set -e aborted before the old `rm`.

`Tests: no automated test suite for .sh utilities — manual verification above; 4 checks passed, 0 skipped, 0 failed`

**Coverage gate:** No JS/TS/py test framework covers these standalone `.sh` utilities; behavior verified via the live repro above (success + failure + cleanup). Deferring an automated harness for shell utilities as out of scope for a one-line path fix.

**Suppression / test-integrity greps:** Suppression grep clean. Test-integrity grep produced one match — `process.exit(1)` inside the unchanged `node -e` snippet matched the `xit\(` Jest-marker pattern as a substring of "e**xit(**". This is a documented false-positive class (substring over-match), not a `xit(` pending-test marker. No real violation.

## ADR-warrant

N/A — bug fix to an existing utility script. No rule/hook contract/skill/settings change; rationale is captured in the issue, code comments, and commit message.

## Doc-reconciliation

N/A — invocation syntax unchanged, so the `docs/REFERENCE.md` Utilities row and `README.md` need no update.

`/review` (claude-opus-4-8) returned **0 blocking, 0 non-blocking** findings — nothing to fix or file. The `reviewed-by-claude` label is applied. ([review comment](https://github.com/brownm09/dev-env/pull/335#issuecomment-4643304139))

### DIFF
--- a/claude/scripts/get-project-item.sh
+++ b/claude/scripts/get-project-item.sh
@@ -23,25 +23,28 @@ fi
 PROJECT_NUMBER="${2:-${PROJECT_NUMBER:-3}}"
 OWNER="${3:-${PROJECT_OWNER:-brownm09}}"
 
-SCRATCH="${HOME}/.claude/scratch"
+# Literal Windows-style path (not "${HOME}/...") so Git Bash and Node resolve it
+# identically. Under Git Bash $HOME is the MSYS path /c/Users/brown; Node on Windows
+# resolves a leading /c/... against the current drive (-> C:\c\Users\...), so the
+# redirect-write and the node-read would target different files (dev-env#334). This
+# matches the repo convention (DEFAULT_LOG in session-mode-report.py).
+SCRATCH="C:/Users/brown/.claude/scratch"
 TMPFILE="${SCRATCH}/tmp_project_item_$$.json"
+trap 'rm -f "$TMPFILE"' EXIT
 
 gh project item-list "$PROJECT_NUMBER" \
   --owner "$OWNER" \
   --format json \
   --limit 1000 > "$TMPFILE"
 
+# Under `set -e` a non-zero exit from node (e.g. no matching item) aborts the script
+# with that status; the EXIT trap above removes the temp file either way. (The former
+# `EXIT_CODE=$?` guard here was dead code — set -e aborted before it could run.)
 ITEM_ID=$(node -e "
   const d = JSON.parse(require('fs').readFileSync('$TMPFILE', 'utf8'));
   const item = d.items.find(i => i.content && i.content.number === $ISSUE_NUMBER);
   if (!item) { process.stderr.write('get-project-item: no item found for #$ISSUE_NUMBER\n'); process.exit(1); }
   console.log(item.id);
 ")
-EXIT_CODE=$?
-rm -f "$TMPFILE"
-
-if [[ $EXIT_CODE -ne 0 ]]; then
-  exit 1
-fi
 
 echo "$ITEM_ID"


brownm09:
## Review: Fix get-project-item.sh scratch-path resolution on Windows + Git Bash (#334)

### Summary
**Merge.** This is a tight, correct fix for a real cross-interpreter path-resolution bug, with a well-targeted secondary cleanup. The diff is appropriately scoped (one file, 17 lines). No blocking or non-blocking findings.

The core fix is sound: a `${HOME}`-derived path (`/c/Users/brown/...`) that Git Bash writes via redirect but Node-on-Windows reads back resolves to two different files (`C:\Users\...` vs `C:\c\Users\...`). Hardcoding the literal `C:/Users/brown/.claude/scratch` makes both interpreters agree, and it matches the established `DEFAULT_LOG` precedent in `session-mode-report.py`.

The secondary fix is also correct: the old `EXIT_CODE=$?` / `if [[ $EXIT_CODE -ne 0 ]]` guard was genuinely dead under `set -e` (the script aborts on a non-zero `node -e` before reaching the assignment). Replacing it with `trap 'rm -f "$TMPFILE"' EXIT` is the right idiom — the trap is registered immediately after `TMPFILE` is assigned and before the `gh` redirect, so the temp file is now cleaned up on *every* exit path including a `gh`/`set -e` abort, which the old inline `rm` never guaranteed. The no-match case still exits 1 via `set -e` propagating node's exit code (the `ITEM_ID=$(...)` is a plain, non-`local` assignment, so the substitution failure is not masked). Both paths were verified live in the PR body (success → `PVTI_...`, no-match → exit 1, temp-file count steady).

### Gate checks
- **Doc reconciliation (2b/2c):** `claude/scripts/` was touched, but this is a content modification — not an add/remove/rename — and the script's invocation signature is unchanged, so the `docs/REFERENCE.md` Utilities row still describes it accurately. No README/REFERENCE update warranted.
- **Test coverage (2d):** Bug fix to a `.sh` utility with no automated harness; PR body documents manual verification (success + failure + cleanup paths) and a valid deferral rationale. Not blocking.
- **Test integrity (2e):** No test files or test config touched. The lone integrity-grep hit (`process.exit(1)` matching the `xit(` pattern) is the documented substring false-positive class, correctly called out in the PR body. No violation.

---

PR #338, opened 2026-06-07
Add execution-level verification for shell scripts (#337)

### LINKED ISSUE #334: get-project-item.sh broken on Windows: $HOME scratch path resolves differently in Git Bash vs Node

### LINKED ISSUE #337: Add execution-level verification for claude/scripts shell utilities (smoke test + path-class lint + shellcheck)
## Problem

dev-env#334 (`get-project-item.sh` failing on every invocation on Windows + Git Bash) went undetected because nothing **executes** the `.sh` utility scripts. The `## Testing` suite runs `bash -n` (syntax only — the bug parsed fine) and `ast.parse` on the Python scripts. There is no CI, and even if there were, GitHub's Linux runners would not reproduce a Windows+Git-Bash+Node path-resolution bug.

## Goal

Add execution-level guards so this class of breakage is caught before merge. Three complementary checks (one PR):

1. **Path-class lint** (hermetic, generalizes): flag any `claude/scripts/*.sh` that derives a scratch/temp path from `$HOME`/`${HOME}` and also invokes `node` — the exact failure class from #334 (Git Bash writes `/c/Users/...` but Node-on-Windows re-resolves it to `C:\c\Users\...`). Recommend the literal `C:/Users/brown/.claude/scratch` form. Wire into `## Testing`.
2. **Smoke test** for `get-project-item.sh`: run it against a known issue number and assert a `PVTI_` id; assert the no-match path exits 1. Skip gracefully when `gh` is unauthenticated/offline (network-dependent, so a local pre-PR check, not CI).
3. **shellcheck** pass over `claude/scripts/*.sh` + `claude/hooks/*` in `## Testing` — catches the broader class of shell bugs (unquoted expansions, etc.). Skip-if-absent with a documented install command if shellcheck cannot be installed.

## Acceptance

- New checks added to the dev-env `## Testing` section and the relevant reference docs.
- Path-class lint passes clean on the current tree (post-#334).
- Smoke test passes locally (or skips cleanly without gh auth).

brownm09:
## Summary

Follow-up to #335 (the `get-project-item.sh` fix). That bug — a Windows + Git Bash path-resolution failure that broke the script on *every* invocation — went undetected because nothing **executes** the `.sh` utility scripts: `bash -n` checks syntax only, and the bug parsed cleanly. This PR adds three complementary, execution-level guards, wired into the dev-env `## Testing` section.

| Guard | File | Catches |
|---|---|---|
| **Path-class lint** (hermetic) | `tests/check-script-path-hygiene.sh` | The exact #334 class: a `$HOME`-rooted scratch/temp path passed to `node`. Recommends the literal `C:/Users/brown/.claude/scratch`. Comment mentions of `$HOME` are stripped so a script can *document* the hazard without tripping. |
| **Smoke test** | `tests/test-get-project-item.sh` | `get-project-item.sh` regressing — runs it end-to-end, asserts a known issue → `PVTI_` id, no-match → exit 1 + diagnostic, and temp-file cleanup. Network-dependent → SKIPs cleanly without `gh` auth. |
| **shellcheck gate** | `tests/run-shellcheck.sh` | The broader class of shell bugs. Blocking at `--severity=error` (tree is error-clean); warnings/info advisory. SKIPs with an install hint when shellcheck is absent. |

Why not just "add CI": dev-env has no CI, and GitHub's Linux runners would not reproduce a Windows + Git-Bash + Node path bug — so the path-class lint (hermetic) and the smoke test (local, network-gated) are the guards that actually catch this class.

All three new scripts are themselves shellcheck-clean at info severity, route diagnostics to stderr, and use deliberate exit codes (0 = pass/skip, non-zero = fail) per the dev-env `## Observability` contract.

## Testing

dev-env `## Testing` suite (run with `SHELLCHECK_BIN` pointed at a portable shellcheck 0.10.0):

- **Hook-script syntax check** (`ast.parse` on `claude/scripts/*.py`) → AST OK
- **`pyw -3` stdio test** → exit 0
- **Path-hygiene lint** → `clean (10 shell files scanned)`
- **Smoke test** → `Tests: 4 passed, 0 failed` (resolved #334 → `PVTI_...`; no-match exit 1 + diagnostic; no leftover temp files)
- **shellcheck gate** → `PASS (0 error-severity findings)`; 3 pre-existing warnings + info printed advisorily (all in pre-existing files, unrelated to this change)
- **Self-check:** ran the new shellcheck gate against the 3 new scripts at `--severity=info` → clean (no warnings/info introduced)

`Tests: dev-env has no JS/py unit suite for .sh utilities — these scripts ARE the verification layer; 5 suite checks + 4 smoke assertions passed, 0 skipped, 0 failed`

**Coverage gate:** This PR *is* test coverage — it adds execution verification for the `.sh` scripts that previously had none. The new scripts are verified by running them (above).

**Suppression / test-integrity greps:** Both clean. (No JS/TS test markers; the new files are shell.)

**shellcheck install note:** shellcheck is not installed on this machine and `choco install shellcheck` requires an elevated shell (fails with `UnauthorizedAccessException` non-elevated). I verified the gate by running a portable shellcheck 0.10.0 binary via `SHELLCHECK_BIN`. The gate SKIPs (exit 0) with an install hint when the binary is absent, so it never blocks a PR on a missing tool.

## Pre-existing shellcheck findings (advisory, out of scope)

The shellcheck advisory pass surfaced pre-existing **warnings** in files this PR does not touch — not gated, noted here for visibility:
- `merge-stale-pr.sh:53` — SC2010 (`ls | grep`)
- `hooks/tests/test-pre-push-lockfile.sh:19` — SC2034 (unused var), `:125` — SC2046 (word-splitting)
- plus several SC2015 (`A && B || C`) and SC2086/SC2012 info-level notes in existing test scripts.

These predate this branch and are unrelated to the #334 class. Filing them as a follow-up cleanup would be the right call rather than absorbing them here (scope guard). Let me know if you'd like that issue opened.

## ADR-warrant

N/A — adds verification *tooling* implementing the project's existing "`## Testing` is the canonical verification set" convention; not a new cross-cutting policy referenced by other CLAUDE.md files. Rationale captured in #337 + script headers + this PR.

## Doc-reconciliation

`claude/scripts/` changed → `docs/REFERENCE.md` updated (new "Script verification suite" subsection) and the dev-env `## Testing` section (steps 5–7). README Hooks table intentionally **not** touched — these are test helpers, not runtime hooks, matching the existing `test-merge-findings-gate.sh` / `test-pre-push-lockfile.sh` precedent (neither is in the README Hooks table).

Closes #337

`/review` (claude-opus-4-8) → marker `blocking=0 non_blocking=0`. One correctness finding was self-caught during review and **fixed in-PR** (commit on this branch: path-hygiene lint now ignores comments in node-detection and reports real file line numbers; verified with a planted-offender positive test). Nothing left to file. `reviewed-by-claude` label applied. ([review comment](https://github.com/brownm09/dev-env/pull/338#issuecomment-4643356541))

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -47,6 +47,37 @@ before PR" rule in [`claude/CLAUDE.md`](claude/CLAUDE.md) defers to this section
    `grep -n 'date -u' claude/CLAUDE.md` and confirm every match is in an internal operational
    artifact context (lock files, log timestamps) — not in stub filename or branch name descriptions.
 
+5. **Script path-hygiene lint** — required when adding or changing any `claude/scripts/*.sh` or
+   `claude/hooks/*` shell script. Flags the [dev-env#334](https://github.com/brownm09/dev-env/issues/334)
+   failure class: a `$HOME`-rooted scratch/temp path passed to `node` (Git Bash writes
+   `/c/Users/...` → `C:\Users\...`, but Node-on-Windows re-resolves the same string to
+   `C:\c\Users\...` → ENOENT). Scripts must use the literal `C:/Users/brown/.claude/scratch`
+   instead. Hermetic; comments mentioning `$HOME` are ignored.
+
+   ```bash
+   bash claude/scripts/tests/check-script-path-hygiene.sh
+   ```
+
+6. **`get-project-item.sh` smoke test** — required when changing `claude/scripts/get-project-item.sh`.
+   Actually *runs* the script (which `bash -n` cannot — #334 parsed cleanly yet failed at runtime):
+   asserts a known issue resolves to a `PVTI_` id, the no-match path exits 1, and the temp file is
+   cleaned up. Network-dependent — SKIPs (exit 0) when `gh` is unauthenticated/offline, so it is a
+   local pre-PR check, not a CI gate.
+
+   ```bash
+   bash claude/scripts/tests/test-get-project-item.sh
+   ```
+
+7. **shellcheck gate** — recommended when changing any shell script. Blocking at `--severity=error`
+   (the tree is error-clean as of 2026-06-07); pre-existing warnings/info are printed advisorily and
+   not gated. SKIPs (exit 0) with an install hint when `shellcheck` is absent — it is not installed by
+   default here and `choco install shellcheck` needs an elevated shell, so set `SHELLCHECK_BIN` to a
+   [portable binary](https://github.com/koalaman/shellcheck/releases) to run it locally.
+
+   ```bash
+   bash claude/scripts/tests/run-shellcheck.sh
+   ```
+
 ## Observability
 
 dev-env has **no long-running runtime to instrument** — it is a configuration repo whose
--- a/claude/scripts/tests/check-script-path-hygiene.sh
+++ b/claude/scripts/tests/check-script-path-hygiene.sh
@@ -33,14 +33,18 @@ mapfile -t FILES < <(
   } | sort -u
 )
 
+# grep the file for $1 with real line numbers, dropping full-line comments so a
+# script that merely *documents* the hazard (mentions node or $HOME in a comment)
+# is not flagged. Real file line numbers are preserved (grep -n on the file, then
+# filter out `N:   # ...` comment lines) so offender output points at the true line.
 OFFENDERS=0
 for f in "${FILES[@]}"; do
   [ -f "$f" ] || continue
-  # (a) does it call node?
-  grep -qE '\bnode\b' "$f" || continue
-  # (b) non-comment line assigning a $HOME-rooted path? Strip full-line comments
-  # first so a documentation mention of $HOME does not match.
-  hits=$(grep -vE '^[[:space:]]*#' "$f" | grep -nE '=[[:space:]]*"?'"'"'?\$\{?HOME\}?/' || true)
+  code_lines() { grep -nE "$1" "$f" | grep -vE '^[0-9]+:[[:space:]]*#' || true; }
+  # (a) does it actually call node (outside comments)?
+  [ -n "$(code_lines '\bnode\b')" ] || continue
+  # (b) does it assign a $HOME-rooted path (outside comments)?
+  hits=$(code_lines '=[[:space:]]*"?'"'"'?\$\{?HOME\}?/')
   if [ -n "$hits" ]; then
     OFFENDERS=$((OFFENDERS + 1))
     {
--- a/claude/scripts/tests/check-script-path-hygiene.sh
+++ b/claude/scripts/tests/check-script-path-hygiene.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+# check-script-path-hygiene.sh — lint for the dev-env#334 failure class.
+#
+# Bug class: a scratch/temp path derived from $HOME is written by a Git Bash
+# redirect (which maps the MSYS `/c/Users/...` to `C:\Users\...`) but then read
+# back by Node-on-Windows, which resolves a leading `/c/...` against the current
+# drive (`C:\c\Users\...`). The write and the read target different files →
+# ENOENT on every invocation. See dev-env#334 and the fix in get-project-item.sh.
+#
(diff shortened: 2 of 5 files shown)


brownm09:
## Review: Add execution-level verification for shell scripts (#337)

### Summary
**Merge.** Three small, self-contained verification scripts plus docs; 226 additions, 0 deletions, appropriately scoped. The scripts are themselves shellcheck-clean, route diagnostics to stderr with deliberate exit codes, and degrade gracefully (smoke test SKIPs without `gh`; shellcheck gate SKIPs when absent). I reviewed the new code adversarially — one precision gap found and fixed in-PR (below); no remaining blocking or non-blocking findings.

### Findings

**[correctness] — `tests/check-script-path-hygiene.sh` (fixed in this PR)**
The lint stripped comments before the `$HOME`-path check but **not** before the `node` check, so a script with a `$HOME`-rooted path plus the word "node" appearing only in a comment could false-positive. Fixed by routing both checks through a `code_lines()` helper that greps the file directly (preserving **true file line numbers**) and drops full-line comments. Verified with a planted offender — flagged at the real assignment line, while a comment mentioning `${HOME}` and `node` on an adjacent line was correctly ignored; and a negative run on the current tree stays clean. Fixed in the latest commit on this branch.

### Verification observed
- Path-hygiene lint: positive test (planted offender → exit 1, real line number) **and** negative test (clean tree → exit 0).
- Smoke test: `4 passed, 0 failed` against live `gh` (resolved #334 → `PVTI_...`; no-match → exit 1 + stderr diagnostic; no temp-file leak).
- shellcheck gate: `PASS (0 error-severity findings)` via a portable shellcheck 0.10.0; new scripts clean at info severity.

### Design notes (non-blocking, no action needed)
- **Smoke-test fixture coupling:** `KNOWN_ISSUE=334` is a real board item; if it is ever removed the success assertion fails (loudly) rather than skips. This is documented in the script with the remedy (point it at any current item) and is the right failure mode for a smoke test.
- **shellcheck not gated in CI:** it can only run where a shellcheck binary is present (not installable here without elevation). The skip-if-absent design is correct — it documents the install path and never blocks on a missing tool.

### Pre-existing shellcheck findings
The advisory pass surfaces warnings/info in files this PR does not touch (e.g. `merge-stale-pr.sh` SC2010, `test-pre-push-lockfile.sh` SC2034/SC2046). Out of scope here, flagged in the PR body for a possible follow-up cleanup — not gated.

---

Issue #356, opened 2026-06-13
usage-snapshot: self-healing OAuth token refresh (spec — deferred behind verification gate)


brownm09:
Follow-up to #355. **Spec only — do not implement until the verification gate below is cleared.**

## Goal

When the stored access token is expired but a `refreshToken` is present, mint a fresh access token, write it back atomically, and proceed with the snapshot — falling back to the #355 advisory if refresh fails. This makes the hook self-heal instead of requiring an interactive `claude` launch to rewrite the credentials file.

## Proposed design

1. **Config gate** — add `"self_heal_token": false` (default off) to `claude/usage-config.json`; document in `docs/REFERENCE.md` Configuration. Self-heal runs only when explicitly enabled.
2. **Refresh function** — `refresh_access_token(refresh_token) -> dict | None`: `POST` with `{grant_type: "refresh_token", refresh_token, client_id}`, returning `{accessToken, expiresAt(ms), refreshToken}` ([RFC 6749 §6](https://datatracker.ietf.org/doc/html/rfc6749#section-6)). Any error → `None`.
3. **Atomic write-back** — `_write_credentials()`: write `.credentials.json.tmp` then `Path.replace()` (atomic on same volume); preserve all other keys, update only `claudeAiOauth` subfields. Never log token / refresh-token values.
4. **`main()` composition** — on `state == "expired"`: if `self_heal_token` and a refresh token exists → attempt refresh → on success continue with the new token → else #355 advisory + `exit 2`.

## Verification gate — MUST resolve before implementing

- **Undocumented endpoint + client_id.** The token endpoint (`https://console.anthropic.com/v1/oauth/token`) and Claude Code public `client_id` are **observed from the OAuth flow, not a published Anthropic API**. Verify against the live flow; treat as liable to change without notice.
- **Rotating refresh tokens (dominant risk).** If Anthropic rotates the refresh token on each use, the hook consuming it could invalidate / race the primary client's session and force a re-login. Verify empirically (does a second refresh with the old token fail?) before enabling. This is why the gate defaults off.
- **Concurrent writes.** Desktop client + hook both writing `.credentials.json` → corruption / lost updates. Atomic replace mitigates corruption, not a logical race; document the residual.

Because the file is *already* 8 days disconnected from the live client's token lineage, the likely practical outcome is that the hook maintains an independent token in this file — but confirm, don't assume.

## ADR

[...]

PR #357, opened 2026-06-13
Make usage-snapshot surface expired/missing OAuth token instead of silent exit

### LINKED ISSUE #355: usage-snapshot: surface expired/missing OAuth token instead of silent exit
## Problem

The post-merge `### Usage Snapshot` block silently stopped appearing for ~8 days. Root cause (confirmed): `claude/scripts/usage-snapshot.py` reads the OAuth token from `C:/Users/brown/.claude/.credentials.json`, finds it **expired**, and hits the silent `sys.exit(0)` at usage-snapshot.py:446. The current client (desktop/Cowork) refreshes its token elsewhere and never writes back to that file, so the on-disk token the hook reads is frozen (last written 2026-06-04, ~8.3 days stale).

Two defects:
1. **Silently dead feature** — every merge since the lapse emitted nothing; journal Section 7 usage telemetry has been blank.
2. **Failure hides itself** — the *expired* branch exits 0 silently, while the *expiring-within-1h* branch right below (line 448) prints a refresh warning. The moment the token actually lapsed, the warning meant to prompt a refresh also went quiet.

## Fix (this issue)

Make the expired-token (and missing-token-when-creds-present) paths emit a one-line stderr advisory + `exit 2`, mirroring the existing within-1h pattern (`exit 2` + stderr is already this hook's output contract). Extract a pure `classify_token(expires_at_ms, now_ms)` helper to make it unit-testable offline, and add `claude/scripts/tests/test_usage_snapshot.py` following the `test_pyw_stdio.py` harness. Keep `creds is None` (file absent) silent — legitimate in unconfigured environments.

## Out of scope

Self-healing token refresh via the stored refresh token — tracked separately (follow-up issue, spec only, deferred behind a verification gate).

## Impact

Medium — post-merge usage telemetry silently degraded for ~8 days with no signal; the advisory makes the failure self-reporting.

brownm09:
Closes #355. Follow-up self-healing spec tracked in #356.

## Problem

The post-merge `### Usage Snapshot` block silently stopped appearing for ~8 days. `claude/scripts/usage-snapshot.py` reads the OAuth token from `C:/Users/brown/.claude/.credentials.json`, found it **expired**, and hit a silent `sys.exit(0)` (old line 446). The current client refreshes its token elsewhere and never writes back to that file, so the on-disk token the hook reads is frozen (~8.3 days stale, confirmed this session). Worse, the *expiring-within-1h* branch right below it printed a warning — so the moment the token actually lapsed, the warning meant to prompt a refresh also went quiet.

## Change

- Extract a pure `classify_token(expires_at_ms, now_ms) -> (state, advisory)` helper (`no_expiry` / `ok` / `expiring` / `expired`).
- Rewire `main()`: `expired` and `expiring` print a one-line stderr advisory + `exit 2` (this hook's existing output contract); `ok` / `no_expiry` proceed to `fetch_usage` as before.
- Surface the missing-token-when-creds-present case (was a silent `exit 0`); `creds is None` (file absent/corrupt) stays silent — legitimate in unconfigured environments.
- New offline test `claude/scripts/tests/test_usage_snapshot.py` + registered as command 8 in the dev-env `## Testing` section.

## Testing

Ran from the worktree root:

- Syntax check (`ast.parse` over `claude/scripts/*.py`): **OK**
- `py -3 claude/scripts/tests/test_usage_snapshot.py` — **Tests: 5 passed, 0 skipped, 0 failed**
- `py -3 claude/scripts/tests/test_pyw_stdio.py` (unchanged invariant; also confirms the modified hook still parses and resolves) — **Tests: 6 passed, 0 skipped, 0 failed**

Pre-PR greps: suppression grep clean; test-integrity grep clean (the `sys.exit(` hits are the known `exit(` ⊃ `xit(` false positive, not skip markers). No `package.json` touched — lockfile policy N/A.

## Coverage

New behavior (the four classifier states + expiry boundary) is covered by the new offline test. The live `fetch_usage` network call is intentionally not covered — the repo avoids urllib mocks, consistent with existing script tests.

## Risk-dimension audit

- **Testing** — covered (above).
- **Observability** — the change *is* an observability fix: routes the expired/missing-token diagnostic to stderr with a deliberate `exit 2`, per the dev-env `## Observability` contract.
- **Security** — no token values logged; the advisory names the file, never its contents. No new credential reads/writes (self-healing write-back is deferred to #356).
- **Resilience** — strictly widens visibility; `ok`/`no_expiry` paths and the outer `except: sys.exit(0)` guard are unchanged, so a malformed payload still fails safe.
- **Performance** — N/A, pure local branch logic.
- **Data integrity** — N/A, no persisted state changed.
- **ADR warrant** — consistency fix within the existing hook-output contract (ADR-027); no new ADR. The deferred self-healing work (#356) introduces a credential-writing convention and will warrant an ADR when implemented.
- **Doc reconciliation** — no script added/removed, so README/REFERENCE Hooks tables are unchanged; only the new `## Testing` command line was added.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,6 +78,17 @@ before PR" rule in [`claude/CLAUDE.md`](claude/CLAUDE.md) defers to this section
    bash claude/scripts/tests/run-shellcheck.sh
    ```
 
+8. **usage-snapshot classifier test** — required when changing `claude/scripts/usage-snapshot.py`.
+   Exercises the pure `classify_token()` helper offline (no network, no credentials file): pins the
+   `no_expiry` / `ok` / `expiring` / `expired` states and the expiry boundary, asserting an expired
+   token now yields a user-facing advisory rather than the silent skip it did before
+   [#355](https://github.com/brownm09/dev-env/issues/355). The live usage API call is not covered
+   (the repo avoids urllib mocks).
+
+   ```bash
+   py -3 claude/scripts/tests/test_usage_snapshot.py
+   ```
+
 ## Observability
 
 dev-env has **no long-running runtime to instrument** — it is a configuration repo whose
--- a/claude/scripts/tests/test_usage_snapshot.py
+++ b/claude/scripts/tests/test_usage_snapshot.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python3
+"""Unit tests for usage-snapshot.py token classification.
+
+`usage-snapshot.py` is a PostToolUse hook that fires after `gh pr merge` and
+emits a usage snapshot. Before the fix in dev-env#355, an already-expired OAuth
+token made the hook `sys.exit(0)` silently — the feature went dead with no
+signal, even though an "expires within 1 hour" warning sat one branch below.
+
+The fix extracts the expiry decision into the pure `classify_token()` helper so
+it can be exercised offline (no network, no credentials file), matching the
+repo's fixture-only test convention. These tests pin the four states and assert
+the previously-silent `expired` path now yields a user-facing advisory.
+
+The live network call (`fetch_usage`) is intentionally not tested — the repo
+avoids urllib mocks, consistent with the other script tests.
+
+Usage:
+    py -3 claude/scripts/tests/test_usage_snapshot.py
+
+Exit 0 = all pass.
+"""
+
+import importlib.util
+import sys
+from pathlib import Path
+
+# tests/ -> scripts/ -> claude/ -> repo root
+REPO_ROOT = Path(__file__).resolve().parents[3]
+SCRIPT = REPO_ROOT / "claude" / "scripts" / "usage-snapshot.py"
+
+# The script filename is hyphenated, so import it by path rather than `import`.
+_spec = importlib.util.spec_from_file_location("usage_snapshot", SCRIPT)
+assert _spec and _spec.loader, f"cannot load module spec from {SCRIPT}"
+usage_snapshot = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(usage_snapshot)  # safe: main() is guarded by __main__
+classify_token = usage_snapshot.classify_token
+
+NOW_MS = 1_700_000_000_000  # fixed synthetic "now"; real time never consulted
+HOUR_MS = 3_600_000
+DAY_MS = 86_400_000
+
+
+def test_no_expiry_proceeds_silently() -> str:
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: Make usage-snapshot surface expired/missing OAuth token instead of silent exit

### Summary
**Merge.** Small (168-line), well-scoped, behavior-preserving change that converts two silent `sys.exit(0)` paths into visible stderr advisories + `exit 2`, matching the hook's existing output contract. The expiry decision is correctly extracted into a pure `classify_token()` and covered by a deterministic offline test (5 cases incl. the inclusive `<=` boundary). No blocking findings. The highest-severity item is informational only (Step 2e language-scope), already satisfied manually.

### Gate results
- **Doc reconciliation (2b):** Not warranted. The Documentation Maintenance trigger is *add/remove/rename a script*; this PR adds a **test** file (test files are not listed in the README Hooks table) and **modifies** `usage-snapshot.py` without renaming it (its README row stays accurate). The mechanical `claude/scripts/**` match is a false positive here.
- **Doc coverage (2c):** No README index exists under `claude/scripts/tests/`; no gap.
- **Test coverage (2d):** New behavior (`classify_token`) is tested. Pass.
- **Test integrity (2e):** Python diff — JS/TS scanners don't apply. Manually verified: no skip markers, deletions, lowered thresholds, or bypass flags introduced; PR body carries `Tests: N passed, 0 skipped, 0 failed`. No violation.

---

### Questions for Author

**Question:** Was preserving the "expiring within 1h ⇒ warn and skip the snapshot" behavior (no `fetch_usage`) intentional, rather than still fetching while the token is valid?
**Context:** In `classify_token` ([usage-snapshot.py:469-490](claude/scripts/usage-snapshot.py)), `state == "expiring"` exits 2 before `fetch_usage`, so during a token's final hour of validity no snapshot is emitted — this faithfully preserves the pre-PR inline branch, so it's not introduced here.
**Tradeoffs:** (a) Keep as-is — zero behavior change vs. main, smallest diff, but you lose snapshots in the last hour even though the token would still work. (b) For `expiring`, warn *and* continue to `fetch_usage` — captures the snapshot while the token is valid, but expands this PR's scope beyond the stated "make failures visible" goal and would deserve its own test case. Out of scope for #355; raising only to confirm the carry-over was deliberate.

---

### Findings Disposition
No blocking or non-blocking findings — disposition section not required.

---

PR #360, opened 2026-06-15
Keep-warm scheduled task to keep OAuth token fresh for usage-snapshot (#359)

### LINKED ISSUE #356: usage-snapshot: self-healing OAuth token refresh (spec — deferred behind verification gate)

### LINKED ISSUE #359: Automate OAuth token freshness so post-merge usage snapshot doesn't need a manual refresh
## Problem

The post-merge `usage-snapshot.py` hook reads the short-lived OAuth **access token** from `~/.claude/.credentials.json`. The desktop/Cowork client refreshes its token *elsewhere* and never writes that file back, so the on-disk token goes stale (it sat ~8 days stale before #357). PR #357 made a stale token **visible** (advisory instead of a silent skip), but restoring the snapshot still requires a **manual** interactive `claude` run on each machine to rewrite the file. That manual chore recurs on every lapse and on every machine.

## Goal

Keep `~/.credentials.json`'s token fresh **automatically**, per machine, so the post-merge usage snapshot works without manual intervention.

## Approaches investigated

- **Long-lived `setup-token` (`CLAUDE_CODE_OAUTH_TOKEN`)** — rejected. A valid `sk-ant-oat…` token is **HTTP 403** at `https://api.anthropic.com/api/oauth/usage` (authenticated but not authorized; the long-lived token class lacks the subscription-usage scope the `/login` token carries).
- **Self-healing refresh inside the hook (#356)** — rejected. Re-implements a private OAuth refresh client and risks refresh-token rotation invalidating the desktop client's session. **This issue supersedes #356.**

## Chosen approach

A **local Windows scheduled task** that periodically runs the Claude CLI to trigger a token refresh — the same action the user does manually today — keeping `.credentials.json` fresh. Local by necessity (it runs the local `claude.exe` writing the local credentials file), so each machine registers its own task.

## Open verification (instrument, don't assume)

Interactive `claude` is confirmed to rewrite `.credentials.json`; **headless `claude -p` refresh-and-write near expiry is unverified**. The payload script will log token expiry + file mtime before/after each run so a few days of observation either confirm or disprove the mechanism. #357's advisory is the safety net during the trial.

brownm09:
Closes #359. Supersedes/closes #356.

## Problem

The post-merge `usage-snapshot.py` hook reads the short-lived OAuth access token from `~/.claude/.credentials.json`. The desktop/Cowork client refreshes its token elsewhere and never writes that file back, so it goes stale and the snapshot stops working until a manual interactive `claude` run rewrites it — on every lapse, on every machine. #357 made the failure visible (advisory instead of silent skip); this removes the manual chore.

## What changed

- **`claude/scripts/keep-token-warm.ps1`** — scheduled-task payload. Resolves `claude.exe` dynamically, runs `claude -p 'ok' --model haiku` (timeout-bounded) to trigger the CLI's own token refresh, and logs the credentials-file mtime + token minutes-to-expiry **before/after** each run to `Documents\LOGS\keep-token-warm_<date>.txt`. The token value is never read or logged. Always exits 0 (best-effort).
- **`claude/scripts/register-keep-token-warm.ps1`** — per-machine registration of the `ClaudeKeepTokenWarm` task: every 4h, **non-elevated** (RunLevel Limited, Interactive logon), **hidden** (ADR-041-compliant). Idempotent; `-Unregister` removes it. Action points at the junctioned `~/.claude/scripts/keep-token-warm.ps1` so it tracks `main`.
- **ADR-043 + INDEX** — records the HTTP 403 finding that killed the long-lived-token path, why keep-warm over self-healing (#356, superseded), cadence, and the unverified-headless-refresh caveat the logging instruments.
- **REFERENCE.md** — Utilities entries for both scripts + a token-freshness cross-link on the `usage-snapshot.py` PostToolUse row.

## Rejected alternatives

- **Long-lived `setup-token` (`CLAUDE_CODE_OAUTH_TOKEN`)** read by the hook — verified **HTTP 403** at `https://api.anthropic.com/api/oauth/usage` (token class lacks the subscription-usage scope).
- **Self-healing refresh in the hook (#356)** — credential write-back from a hook + refresh-token rotation risk against an undocumented endpoint.

## Testing

Ran the relevant dev-env `## Testing` commands (touched no Python; these confirm no regression):

- `py -3 -c "ast.parse ..." claude/scripts/*.py` → **PASS** (all parse)
- `py -3 claude/scripts/tests/test_pyw_stdio.py` → `Tests: 6 passed, 0 skipped, 0 failed`
- `py -3 claude/scripts/tests/test_usage_snapshot.py` → `Tests: 5 passed, 0 skipped, 0 failed`

PowerShell scripts (no Python tests apply):

- Parse-validated both `.ps1` with `[System.Management.Automation.Language.Parser]::ParseFile` → clean.
- **Live smoke test** of the payload: `claude.exe` resolved, ran headlessly (exit 0, 18.6s), logged `before[mtime=… expiry=455.6min] after[… expiry=455.3min] | no-change` — correct (token healthy → lazy refresh is a no-op; no secrets in the log).
- **Registration validated** non-elevated: task `ClaudeKeepTokenWarm` State=Ready, interval PT4H (indefinite duration), Principal Interactive/**Limited**, Hidden=True.

**Combined: 11 passed, 0 skipped, 0 failed.**

### Coverage gate

New behavior is in PowerShell; dev-env has no PowerShell unit-test harness. Coverage is deferred to the parse check + live smoke tests above, plus the **multi-day observation built into the payload's logging** (it records whether a near-expiry run actually `REFRESHED` the token — the one unverified mechanism). #357's advisory is the safety net during that observation. This is the honest verification path for a per-machine OS-automation script.

### Other gates

- Suppression grep (ADR-026) and test-integrity grep (ADR-029): **clean** (diff is PowerShell + Markdown).
- Lockfile (ADR-036): **N/A** — no `package.json`.
- ADR-warrant: **ADR-043 included** in this PR.

---

`/review` (self) on this PR found 1 blocking + 1 non-blocking + 1 style note — all resolved in-PR:

- **[correctness]** `Resolve-ClaudeExe` sorted `claude.exe` versions lexically (would pick 2.1.99 over 2.1.170) → **fixed in `5b57279`** (sort by parsed `[version]` of the parent dir, mirroring `sort -V`).
- **[reliability]** `Start-Job` orphaned the grandchild `claude.exe` on the timeout path → **fixed in `5b57279`** (replaced with a directly-killable `ProcessStartInfo` child; also restored a reliable exit code, which `Start-Process -PassThru` did not provide; async stdout/stderr drain; stdin closed for EOF).
- **[style]** Misleading literal backticks in `register-keep-token-warm.ps1` strings → **fixed in `63b2d91`**.

Re-verified after fixes: payload parses clean and smoke-tests `claude_exit=0`, ~14s, no orphan.


brownm09:
## Review: Keep-warm scheduled task to keep OAuth token fresh for usage-snapshot (#359)

### Summary
Merge with minor changes. The design is sound — a local scheduled task that lets the CLI own the token refresh sidesteps the credential-write-back risk that killed #356, and the before/after logging honestly instruments the one unverified assumption. One **blocking correctness** issue: the `claude.exe` resolver sorts versions lexically, so it can pick an *older* binary than intended (the very bug `~/bin/claude` avoids with `sort -V`). One non-blocking reliability edge on the timeout path. Docs/ADR are thorough and accurate.

---

### Blocking Findings

**[correctness]** `claude/scripts/keep-token-warm.ps1:75` (`Resolve-ClaudeExe`)
`Sort-Object FullName | Select-Object -Last 1` orders the candidate `claude.exe` paths **lexically** by full path, then takes the last. Version directory names don't lexically sort in version order: `"...\claude-code\2.1.170\claude.exe"` sorts *before* `"...\2.1.99\claude.exe"` (char-by-char, `2.1.1…` < `2.1.9`), so once two versions coexist after an upgrade, this selects the **older** binary — the opposite of the "newest version" the function documents. The `~/bin/claude` shim this mirrors deliberately uses `sort -V | tail -1` to avoid exactly this.
**Fix:** Sort by the parsed version of the parent directory, not the path string, e.g. `Sort-Object { try { [version]$_.Directory.Name } catch { [version]'0.0.0' } } | Select-Object -Last 1`.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/keep-token-warm.ps1:96-112` (timeout path)
On the `TIMEOUT` branch, `Stop-Job` terminates the background job's PowerShell but not its **grandchild** `claude.exe` (spawned via `&` inside the job). The script then logs and exits, the task process ends, and a hung `claude.exe` is re-parented and persists — the task's `ExecutionTimeLimit` no longer applies once the task itself has finished. Low probability (the smoke test returned in 18.6s; a hang needs claude to exceed its own network timeouts), but unbounded if it occurs.
**Suggestion:** Run the CLI as a directly-killable child instead of a job grandchild — `Start-Process -PassThru -NoNewWindow`, `WaitForExit($TimeoutSeconds*1000)`, and `Stop-Process -Force` on timeout (redirect stdin from an empty file so `-p` never blocks). This also drops the `Start-Job` process overhead.

---

### Style Notes
In `register-keep-token-warm.ps1`, the backtick-wrapped words in the `Write-Warning` and `-Description` strings (`` `main` ``, `` `claude` ``) are inside double-quoted strings, where backtick is the escape char — they render as `main` / `claude` without the intended literal backticks. Harmless, but if literal backticks are wanted, use single quotes or escape them (`` `` ` ``).

---

### Findings Disposition
Per ADR-028, each finding must be fixed in this PR or filed and linked before merge. Author to record disposition (fixed in `<sha>` | filed #N) in a "Review findings disposition" section of the PR body before `gh pr merge`.

---

PR #365, opened 2026-06-18
Pre-install free-space gate + prompt post-merge node_modules reclaim (#364)

### LINKED ISSUE #364: Investigate recurring C: disk-full (ENOSPC) that silently corrupts npm installs
## Problem

During a lifting-logbook session on 2026-06-17, the `C:` drive reached **0 GB free** (475.92 GB used) with no deliberate large write. The exhaustion struck mid-`npm install` in a Claude Code worktree and surfaced indirectly — as **corrupted/truncated native dependencies**, not an obvious "disk full" dialog — which cost significant time to diagnose.

This is an environment-level reliability problem (it can hit any project's worktree installs), so it belongs in dev-env rather than a single product repo.

## Evidence / symptoms observed

The `ENOSPC` did not fail loudly; a background `npm install` reported exit 0 but left several packages partially extracted. Downstream failures were misleading:

- **Jest** failed with `Preset ts-jest not found` — the real cause (traced per the Error-Message-Diligence rule) was `bs-logger/dist/index.js` `MODULE_NOT_FOUND` inside ts-jest's load chain.
- **`next dev`** crashed on boot because `@next/swc-win32-x64-msvc` was truncated to **32.5 MB** (valid binary is **136.8 MB**), producing a downstream `next.config.compiled.js` "Unexpected token 'export'". This failed all 15 Playwright tests with `ERR_CONNECTION_REFUSED` before the root cause was found.

Node was 20.11.1 throughout — this is **not** the documented Node-24 incomplete-tarball issue (lifting-logbook #373); it was genuine disk exhaustion truncating writes.

## Band-aids applied (not a fix)

- `npm cache clean --force` reclaimed ~2 GB.
- Re-extracted individual truncated packages (`rm -rf node_modules/<pkg> && npm install <pkg> --no-save`) and ran a clean `npm ci`.
- Free space later rose to ~10 GB on its own, which is still tight.

**Root cause of the disk filling to 0 GB is unknown.** The recovery only restored a small margin.

## What this issue should establish

1. **Find what consumed the disk.** Identify the top space consumers on `C:` — likely suspects given this environment: npm cache, accumulated Claude Code worktrees under `.claude/worktrees/`, Docker Desktop (Testcontainers images/volumes from the API DB E2E suite), Playwright browser bundles, `.turbo` caches, Windows temp.
2. **Add a guard / early warning.** A cheap free-space check (e.g. a `SessionStart` or pre-`npm install` hook, or a periodic routine) that warns when `C:` drops below a threshold (e.g. < 20 GB), so exhaustion surfaces directly instead of as truncated binaries.
3. **Document a cleanup runbook** in dev-env (`docs/REFERENCE.md` or a runbook): worktree pruning (`git worktree prune` + removing stale `.claude/worktrees/*`), `npm cache clean`, `docker system prune`, clearing `.turbo`, and how to recognize ENOSPC-truncated-binary symptoms quickly.

## Acceptance criteria

- [ ] The dominant disk consumer(s) on `C:` are identified and documented.
- [ ] A free-space early-warning mechanism exists (hook or routine) with a defined threshold.
- [ ] A cleanup runbook is documented and linked from the relevant CLAUDE.md / REFERENCE.
- [ ] The ENOSPC-as-truncated-binary failure signature is captured so it's recognized immediately next time.

## References

- Surfaced during lifting-logbook PR [#560](https://github.com/brownm09/lifting-logbook/pull/560) (onboarding TM-direct); the disk event was an unrelated environment failure during that work.
- Related but distinct: lifting-logbook [#373](https://github.com/brownm09/lifting-logbook/issues/373) (Node-24 incomplete tarballs — a different root cause with overlapping symptoms).

brownm09:
Closes #364.

## Problem

`C:` saturated to **0 GB free** mid-`npm install` (dev-env#364), silently truncating `node_modules`: npm reported **exit 0** but left partial native binaries (`@next/swc-win32-x64-msvc` 32.5 MB vs the valid 136.8 MB), surfacing hours later as misleading downstream errors (`Preset ts-jest not found` → really `bs-logger` `MODULE_NOT_FOUND`). ADR-037's `disk-space-check.py` only samples at prompt boundaries, and `worktree-npm-install.py` ran the install **unattended with no free-space check** — the actual corruption vector.

**Dominant consumer (criterion 1), measured this session:** `lifting-logbook/.claude/worktrees/` = **60 worktrees** × ~1–2 GB `node_modules` (~30–60+ GB). Secondary: Docker ~5.9 GB, Playwright ~685 MB, npm cache ~684 MB. dev-env worktrees negligible (17 MB, no `node_modules`).

## Changes

- **`worktree-npm-install.py` — pre-install free-space gate** with a *synchronous* reclamation ladder (Tier 1 idle-worktree reclaim → Tier 2 `npm cache clean`), re-measuring after each; **refuses** the install below a 5 GB hard floor rather than truncate silently. Pure `install_decision()` helper; fails open on any measurement error. (`docker system prune` stays a *manual* lever — destructive, not auto-run.)
- **`post-pr-merge-reclaim.py` — new PostToolUse hook** spawns detached reclamation of now-idle merged worktrees' `node_modules` (the dominant consumer) **at the merge event** instead of waiting for the 6-hourly routine. Worktree directory/branch removal stays the daily `prune-stale-worktrees` job (the active worktree can't remove itself — Windows cwd lock).
- **`docs/REFERENCE.md` → Disk-Full (ENOSPC) Recovery** runbook: failure signature (criterion 4), consumer table (criterion 1), recovery steps (criterion 3). Linked from `claude/CLAUDE.md`.
- **ADR-045** + INDEX; README/REFERENCE hooks tables (added the previously-undocumented `worktree-npm-install.py` row + the new hook); project `CLAUDE.md` `## Testing` items 9–10.

Thresholds (10 / 5 GB) stay hardcoded named constants, consistent with ADR-037.

## Acceptance criteria

- [x] Dominant disk consumer identified + documented (consumer table in runbook + ADR-045)
- [x] Free-space early-warning with a defined threshold — pre-`npm install` gate (10 GB ladder, 5 GB hard floor) closing the gap ADR-037's prompt-boundary check left open
- [x] Cleanup runbook documented + linked from CLAUDE.md
- [x] ENOSPC-as-truncated-binary failure signature captured

## Testing

Ran the relevant items from dev-env `## Testing`:

- **Hook-script syntax check** (`ast.parse` over `claude/scripts/*.py`) — ALL PARSE OK
- **`pyw -3` stdio test** — 6 passed (confirms the new hook is wired: 28 commands all `pyw -3`, all subprocess hooks `import _winsubp`, none spawn Python via `py`)
- **`test_worktree_npm_install.py`** — 6 passed (`install_decision` proceed/reclaim-first/abort + 10 GB & 5 GB boundaries)
- **`test_post_pr_merge_reclaim.py`** — 4 passed (`is_successful_merge` predicate)
- **settings.json** — valid JSON

`Tests: 16 passed, 0 skipped, 0 failed`

**Coverage gate:** new behavior (both pure decision helpers) is covered by the two new offline tests. The shell-out paths (reclamation ladder, real install, detached spawn) are not unit-tested — the repo avoids subprocess/urllib mocks (same convention as `test_usage_snapshot.py`).

**Suppression / test-integrity greps:** suppression grep clean. The test-integrity grep flags two `+        sys.exit(0)` lines — this is the **known `xit(` false positive** (`sys.e`**`xit(`**`0)` matches the `xit\(` skip-marker pattern). Both lines are the **safe-exit guards** I added to the two hooks (advisory hooks must exit 0 on every path, ADR-027), not skip markers. No deleted tests, no lowered thresholds, no `package.json`/lockfile change.

`/review` (claude-opus-4-8) returned **0 blocking, 1 non-blocking** finding + 1 question. Both addressed in-PR per ADR-028:

- **[reliability] `worktree-npm-install.py:108` — mid-ladder `_free_gb` not guarded (fail-closed gap):** **Fixed in 15b8e57.** Wrapped the mid-ladder measurement in `try/except OSError` so a `disk_usage` error falls through to Tier 2 rather than propagating to the safe-exit guard and silently suppressing the install — aligning the code with the documented fail-open contract.
- **[question] Is `timeout=120` for Tier 1 reclaim deliberately conservative?** **Resolved in 15b8e57.** Raised 120s → 300s (matching the install's own ceiling) + added a comment: Windows `rmtree` over dozens of worktree `node_modules` trees is slow, and `reclaim-worktree-disk` early-exits at `--min-free-gb`, so the cap only bites when the disk is deep underwater — exactly when reclamation should continue rather than abort prematurely.

Post-fix verification: `Tests: 16 passed, 0 skipped, 0 failed` (test_worktree_npm_install 6, test_post_pr_merge_reclaim 4, pyw_stdio 6; ast.parse all scripts OK).


brownm09:
## Review: Pre-install free-space gate + prompt post-merge node_modules reclaim (#364)

### Summary
**Merge with one minor change.** This closes the real gap ADR-037 left — an unattended worktree install on a near-full disk silently truncating `node_modules` — by gating the install on free space and refusing below a hard floor rather than corrupting. The design is sound: fail-open on measurement error, synchronous reclaim (correctly *not* detached, since it guards a synchronous install), `--protect-cwd` on both spawns, safe-exit guards on both hooks, and the abort path is the safe direction (refuse, never truncate). The only substantive item is a single un-guarded `disk_usage` call mid-ladder that breaks the documented fail-open contract in a low-probability path; everything else is clean and well-documented. No security or correctness defects.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/worktree-npm-install.py:108`
The mid-ladder re-measurement `if _free_gb(TARGET_DRIVE) >= HARD_FLOOR_GB:` (inside `_run_reclaim_ladder`) is **not** wrapped in `try/except OSError`, unlike the two `_free_gb` calls in `_gate_install` (lines 131, 148) which deliberately fail open. If `shutil.disk_usage` raises here, the exception propagates out of `_run_reclaim_ladder` → out of `_gate_install` → out of `main()` and is swallowed by the `__main__` safe-exit guard as `sys.exit(0)`. That path skips the install *and* emits no advisory — i.e. it fails **closed**, contradicting the module/ADR-045 promise that "any disk-usage error fails open." Low probability (it's the same fixed `C:` drive measured successfully ~one line earlier, and the hook self-heals on the next prompt since `node_modules` is still absent), but it's the one call that doesn't honor the contract the rest of the function advertises.
**Suggestion:** Wrap the line-108 measurement in `try/except OSError` and treat a failure as "fall through to Tier 2" (or `return`), so a measurement error never silently suppresses the install. Aligns the code with its own docstring and ADR-045.

---

### Questions for Author

**Question:** Is `timeout=120` for the Tier 1 synchronous reclaim deliberately conservative?
**Context:** `worktree-npm-install.py:102` caps the `reclaim-worktree-disk.py` subprocess at 120 s. The motivating scenario is the dominant consumer — dozens of lifting-logbook worktrees × ~1–2 GB of `node_modules` (hundreds of thousands of tiny files), and Windows `rmtree` over many small files is slow. If the reclaim is killed at 120 s having freed less than the 5 GB floor, `_gate_install` aborts the install.
**Tradeoffs:** (a) Keep 120 s — abort-on-timeout is the *safe* direction (refuse rather than truncate), so the only cost is a user occasionally hitting the manual `docker system prune` advisory when more reclaim time would have cleared the floor; simplest, no corruption risk. (b) Raise the timeout (e.g. 300 s, matching the install's own timeout) — fewer "false" aborts in exactly the high-worktree-count case this PR targets, at the cost of a longer worst-case prompt stall while reclaim runs. Since the reclaim script early-exits once it reaches `--min-free-gb 10`, the timeout only bites when the disk is genuinely deep underwater — which is precisely when you'd want it to keep going. Worth a one-line note in the code if 120 s is the intended ceiling.

---

### Style Notes
Clean and consistent with the existing hook conventions — `_winsubp` import, `sys.executable` (never `py`) for detached spawns, hardcoded named thresholds matching ADR-037's deliberate choice, thorough module docstrings cross-referencing the issue/ADR. The pure-helper-plus-offline-test pattern mirrors `test_usage_snapshot.py` well. No style concerns.

---

### Findings Disposition
One non-blocking reliability finding (line-108 fail-open gap) — to be fixed in this PR. The timeout item is a question, not a defect (abort-on-timeout is safe-by-design); will be resolved by a code comment or a small bump per the author's judgment. Disposition recorded in the PR body before merge per ADR-028/ADR-039.

---

Issue #369, opened 2026-06-21
Capture post-merge follow-ups as background-task tiles (standing rule)


brownm09:
## Problem

Follow-ups noticed while finishing a piece of work — out-of-scope fixes, deferred work, tech debt, or ideas spotted in passing — are easy to lose once a session ends. The merge of a PR is a natural session boundary where these crystallize, but there is no standing rule to capture them at that moment, so they silently drop.

The Claude Code harness already provides the `spawn_task` chip/tile mechanism for flagging out-of-scope work into its own session + worktree, but no workflow rule ties it to the post-merge checkpoint.

## Desired outcome

A standing, **global** workflow rule: after merging, if the work surfaced any follow-ups, create a background-task tile (`spawn_task`) for each. Per the Durable Preferences rule ([ADR-038](../docs/adr/038-durable-preferences-documented-in-repo.md)), it must be documented in the version-controlled repo — global `claude/CLAUDE.md` plus an ADR extending the post-merge checklist ([ADR-012](../docs/adr/012-post-merge-checklist-board-done-roadmap-update.md)) — not left only in agent memory.

Tiles are a low-friction capture (ephemeral); a follow-up that needs durable tracking still gets an issue. Where `spawn_task` is unavailable (e.g. some terminal sessions), file a follow-up issue instead.

Reported as a process update in session on 2026-06-20.

PR #379, opened 2026-06-21
fix(post-tool-use): read command output from stdout, not output (#377)

### LINKED ISSUE #377: post-tool-use.py silently no-ops on gh issue/pr create (reads tool_response.output; Bash payload uses stdout)
## Problem

The `post-tool-use.py` PostToolUse hook is documented (dev-env CLAUDE.md → GitHub Project, [ADR-023](docs/adr/023-generic-required-fields-issue-hook.md)) to fire after `gh issue create` / `gh pr create`, add the new item to the Dev Env GitHub Project (#3), and exit 2 printing the `gh project item-edit` commands for Impact/Why.

In practice it **never does this** — it silently no-ops, so every issue/PR has to be added to the board manually via the fallback. This was noticed when issue #369 (created in a worktree session) did not auto-add, but the failure is **not** worktree-, session-, or desktop-specific: it affects every session.

## Evidence

- The hook *runs* (wired in `claude/settings.json` PostToolUse/Bash) but takes a silent `sys.exit(0)` path. In the #369 transcript, `usage-snapshot` hook output is captured (14×) but `[project-hook]` appears 0×.
- Across **all** retained transcripts (dev-env + lifting-logbook, months), there is not one real `[project-hook]` fire — every project-add was done manually.

## Root cause

`post-tool-use.py:179` reads `data["tool_response"]["output"]`, but Claude Code's Bash hook payload exposes a command's output under **`stdout`/`stderr`**, not `output` (observed tool-result shape: `interrupted,isImage,noOutputExpected,stderr,stdout` — no `output`/`exitCode` — identical across May and June transcripts). So `output` is always `""`:

1. `extract_github_url("", repo)` → `None`
2. Because the dev-env `hook-config.json` sets `repo`, the `if repo: sys.exit(0)` branch (lines 204-206) fires — **silently**, which is why nobody ever saw a failure.

Verified by running the real hook code against both payload shapes: with a `stdout`-shaped payload `output` is empty and extraction returns `None`; with an `output`-shaped payload it works.

## Scope

This issue covers the fix to **`post-tool-use.py`** (the issue/PR project-add hook). The same empty-`output` read also affects `post-pr-merge-project.py`, `post-pr-merge-pull.py`, `post-pr-merge-reclaim.py`, and `stub-push-archive-reminder.py` — and `post-pr-merge-project.py` additionally needs to derive the PR number from the command rather than the output (`gh pr merge` output carries no `/pull/N` URL). Those are tracked as a separate follow-up.

## Acceptance

- `post-tool-use.py` reads the command output from `stdout`/`stderr` (with a legacy `output` fallback) so the hook fires on `gh issue create` / `gh pr create`.
- The hidden-failure path is de-silenced: a successful create that yields **no** GitHub URL at all emits an advisory rather than exiting silently (while still staying silent for the legitimate different-repo case).
- A regression test covers the output-reading helper and the URL extraction.
- The Bash-hook-payload field constraint is recorded so the sibling hooks and future hooks don't repeat it.

brownm09:
## Summary

`post-tool-use.py` is documented to add new issues/PRs to the Dev Env GitHub Project (#3) after `gh issue create` / `gh pr create` and print the Impact/Why commands. **It never did** — it silently `sys.exit(0)`'d on every invocation, so every project-add was done manually via the fallback. This was surfaced by issue #369 not auto-adding in a worktree session, but the failure is **not** worktree/session-specific.

**Root cause:** the hook read `data["tool_response"]["output"]`, but Claude Code's Bash hook payload exposes a command's output under **`stdout`/`stderr`**, not `output`. So `output` was always `""` → `extract_github_url("", repo)` → `None` → and because the dev-env config sets `repo`, the hook took the **silent** `if repo: sys.exit(0)` branch. Confirmed: zero `[project-hook]` fires exist in months of transcripts (dev-env + lifting-logbook), and creating #377 itself did not auto-add (verified `PRESENT => NO`).

## Changes

- **`claude/scripts/post-tool-use.py`** — route output through a pure `read_command_output(data)` helper (reads `stdout`+`stderr`, legacy `output` fallback). De-silence the no-URL path: a successful create that yields **no** GitHub URL at all now emits an advisory (exit 2); the legitimate different-repo case still exits silently.
- **`claude/scripts/tests/test_post_tool_use.py`** — new offline regression test (9 cases): pins that the real `stdout`-shaped payload yields the URL (the pre-fix `output` read was `""`), the legacy fallback works, and the de-silence predicate distinguishes a different-repo miss from a genuine empty.
- **`docs/adr/049-hook-payload-output-field.md`** + INDEX — records the Bash-payload field constraint so sibling/future hooks don't repeat it.
- **`CLAUDE.md`** — register the new test as Testing item 12.

## Scope / follow-up

This PR fixes `post-tool-use.py` (verified: `gh issue create` prints the full issue URL to stdout). The same empty-`output` read also affects `post-pr-merge-project.py`, `post-pr-merge-pull.py`, `post-pr-merge-reclaim.py`, and `stub-push-archive-reminder.py`. `post-pr-merge-project.py` additionally needs to derive the PR number from the **command** (the `gh pr merge` output carries no `/pull/N` URL — verified in the #369 transcript), so its fix is more than a field rename. **Tracked as a follow-up** (so the move-to-Done hook the #369 stub believed had fired — it had not — gets a correct, verified fix).

## Testing

- `py -3 claude/scripts/tests/test_post_tool_use.py` → **Tests: 9 passed, 0 skipped, 0 failed**
- Hook-script syntax check (`ast.parse` all scripts) → **OK: 32 scripts parse**
- Coverage gate: new behavior (`read_command_output`, de-silence predicate, URL extraction) is covered by the new test.

## Suppression / test-integrity

- One suppression added: `except Exception as e:  # noqa: BLE001` in the test **runner** — a deliberate catch-all that reports an unexpected exception as a test ERROR rather than crashing the run. It mirrors the established convention in `test_usage_snapshot.py` (identical line) and applies only to the runner loop, not the assertions.
- Pre-PR greps: JS/TS suppression grep clean; test-integrity grep clean (no skip markers, no deleted tests, no lowered thresholds).

## ADR-warrant

Warranted — fixes a documented hook and records a durable platform constraint (Bash hook payload field names) → ADR-049 added.

`/review` completed (comment [#issuecomment-4760889231](https://github.com/brownm09/dev-env/pull/379#issuecomment-4760889231)): **0 blocking, 0 non-blocking findings** — clean review, nothing to disposition. `reviewed-by-claude` applied.

Closes #377

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -120,6 +120,17 @@ before PR" rule in [`claude/CLAUDE.md`](claude/CLAUDE.md) defers to this section
     py -3 claude/scripts/tests/test_memory_write_advisory.py
     ```
 
+12. **post-tool-use test** — required when changing `claude/scripts/post-tool-use.py`. Exercises the pure
+    `read_command_output()` and `extract_github_url()` helpers offline: pins that the real `stdout`-shaped
+    Bash payload yields a non-empty output (the pre-fix `output` read was `""` — the [#377](https://github.com/brownm09/dev-env/issues/377)
+    silent no-op), that the legacy `output` field still works, and that the de-silenced no-URL path
+    distinguishes a different-repo miss from a genuine empty ([ADR-049](docs/adr/049-hook-payload-output-field.md)).
+    The live `gh project item-add` call is not covered (the repo avoids subprocess mocks).
+
+    ```bash
+    py -3 claude/scripts/tests/test_post_tool_use.py
+    ```
+
 ## Observability
 
 dev-env has **no long-running runtime to instrument** — it is a configuration repo whose
--- a/claude/scripts/post-tool-use.py
+++ b/claude/scripts/post-tool-use.py
@@ -51,6 +51,23 @@ def load_config(cwd: str) -> dict | None:
         return None
 
 
+def read_command_output(data: dict) -> str:
+    """Return the Bash command's output from a PostToolUse hook payload.
+
+    Claude Code exposes a Bash command's output under ``tool_response.stdout``
+    (and ``stderr``), NOT ``output``. Reading ``output`` yields ``""`` and
+    silently breaks URL extraction (dev-env #377). Read stdout/stderr first and
+    fall back to the legacy ``output`` key for forward/backward compatibility.
+    """
+    tr = data.get("tool_response") or {}
+    if not isinstance(tr, dict):
+        return ""
+    parts = [p for p in (tr.get("stdout"), tr.get("stderr")) if p]
+    if parts:
+        return "\n".join(parts)
+    return str(tr.get("output", "") or "")
+
+
 def extract_github_url(output: str, repo: str | None = None) -> str | None:
     """Return the last GitHub URL found in command output, or None.
 
@@ -176,7 +193,7 @@ def main() -> None:
         sys.exit(0)
 
     command = data.get("tool_input", {}).get("command", "")
-    output = data.get("tool_response", {}).get("output", "")
+    output = read_command_output(data)
     exit_code = data.get("tool_response", {}).get("exitCode", 0)
     cwd = data.get("cwd", "")
 
@@ -200,9 +217,13 @@ def main() -> None:
 
     url = extract_github_url(output, repo)
     if not url:
-        # If a repo filter is configured, a missing URL most likely means the
-        # command targeted a different repo — exit silently rather than warning.
-        if repo:
+        # A configured repo filter can legitimately miss: the command may have
+        # created the item in a *different* repo than this cwd's project. Stay
+        # silent only in that case — some GitHub URL is present, just not ours.
+        # If a successful create produced no GitHub URL at all, that is the
+        # symptom of a real failure (e.g. reading the wrong payload field — the
+        # bug behind #377) and must surface rather than be swallowed silently.
+        if repo and extract_github_url(output, None):
             sys.exit(0)
         print(
             f"[project-hook] {item_type} created but no GitHub URL found in output.\n"
--- a/claude/scripts/tests/test_post_tool_use.py
+++ b/claude/scripts/tests/test_post_tool_use.py
@@ -1,0 +1,151 @@
+#!/usr/bin/env python3
+"""Unit tests for post-tool-use.py command-output reading and URL extraction.
+
+`post-tool-use.py` is a PostToolUse hook that fires after `gh issue create` /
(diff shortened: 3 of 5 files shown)


brownm09:
## Review: fix(post-tool-use): read command output from stdout, not output (#377)

### Summary
Merge-ready. A tight, well-scoped bug fix: `post-tool-use.py` read `tool_response["output"]` (always empty — Claude Code's Bash payload exposes output under `stdout`/`stderr`), so with `repo` set the hook hit a silent `sys.exit(0)` and **never fired**. The fix routes output through a pure `read_command_output()` helper (stdout+stderr, legacy `output` fallback) and de-silences the no-URL path, with a 9-case offline regression test and ADR-049 recording the payload-field constraint. No blocking or non-blocking findings.

### Gate checks
- **Doc reconciliation (2b):** `claude/scripts/post-tool-use.py` changed; README.md (hooks table) and docs/REFERENCE.md already describe the hook's behavior accurately, and this is a behavior *fix* (not add/remove/rename) plus a test — no hooks-table update warranted. The mechanical "scripts changed → update README" trigger is a false positive here.
- **Doc coverage (2c):** new test indexed in the CLAUDE.md Testing section (item 12); ADR-049 added to docs/adr/INDEX.md; no `tests/README.md` exists. Covered.
- **Test coverage gate (2d):** new behavior (`read_command_output`, de-silence predicate, URL extraction) is covered by `test_post_tool_use.py` (9 cases). ✓
- **Test integrity (2e):** no skip markers, no deleted tests, no lowered thresholds. A `.py` file is present — the JS/TS scanners don't apply, so manually verified the diff adds no `@pytest.mark.skip` / `pytest.skip(` / equivalent. PR body carries `Tests: 9 passed, 0 skipped, 0 failed`. The single `# noqa: BLE001` is the test-runner catch-all (justified in PR body; mirrors `test_usage_snapshot.py`). ✓

### Notes (intentional non-findings)
- `read_command_output` joins `stdout`+`stderr`; for `gh issue create` / `gh pr create` the URL is on stdout and stderr is empty, so `extract_github_url` resolves it unambiguously — the both-streams-carry-a-URL edge cannot arise for these commands.
- `exit_code` still reads `tool_response.get("exitCode", 0)`; PostToolUse fires only after success and a missing field defaults to 0 (proceed), so this is correct and intentionally left unchanged.

No blocking or non-blocking findings.

Issue #384, opened 2026-06-22
Prune/reclaim routines can remove a worktree that has a live Claude session


brownm09:
## Problem

A maintenance routine removed a Claude-managed worktree **while a Claude Code session was actively running inside it**, breaking the session mid-task.

Observed 2026-06-21 (~23:42 local), worktree `pensive-taussig-448bc7` (branch `claude/pensive-taussig-448bc7`, sitting *at* `origin/main` — not merged-and-gone):

- The worktree was pruned out from under the live session: its `.git` link file was removed, the admin dir `…/.git/worktrees/pensive-taussig-448bc7/` was deleted, and it was deregistered from `git worktree list`.
- The session's transcript was still being written at the moment of removal (last line `23:42:10`).
- The orphaned session then had every `Write`/`Edit` blocked by the ADR-024 worktree-path guard, and bare `git` from the session cwd silently resolved to the canonical main checkout — unsafe.

## Root cause

`prune-merged-worktrees.py` is the remover (`reclaim-worktree-disk.py` and `post-pr-merge-reclaim.py` only strip `node_modules`/`.turbo`). Two factors combined:

1. `is_merged()` runs `git merge-base --is-ancestor <branch> origin/main`, which returns true when the branch tip **is** `origin/main` (a commit is its own ancestor) — so a fresh worktree that hasn't committed yet (0 commits ahead) is classified "merged."
2. The only active-session protection is `path == os.getcwd()`, which shields only the *pruning process's own* worktree. The `prune-stale-worktrees` routine runs out-of-process and passes no `--protect-cwd`, so it cannot see a live session in any *other* worktree. With the tree momentarily clean, `git worktree remove` succeeded.

## Goal

Prune/reclaim routines must **skip a worktree that has a live/active Claude session** — detected via recent activity in its per-worktree transcript dir (`~/.claude/projects/<worktree-slug>/*.jsonl` mtime within a window) — **in addition to** (not instead of) the existing merged/clean/cwd checks. The guard must only ever *add* skips, never cause more deletion.

Related: ADR-024 (worktree-path guard / orphaned-worktree), ADR-037 (worktree disk reclamation), ADR-045 (pre-install free-space gate). Discovered alongside #381 (PostToolUse hooks not firing).

Issue #381, opened 2026-06-21
PostToolUse hooks do not execute in some sessions (UserPromptSubmit/PreToolUse do) — harness-level


brownm09:
## Observation

In some Claude Code sessions, **PostToolUse hooks do not execute at all**, while UserPromptSubmit and PreToolUse hooks in the same session do. This means none of the 6 registered PostToolUse hooks run — including `post-tool-use.py` (GitHub Project add + Impact/Why prompt), `pr-merge-reminder.py` (journal-stub reminder), `stub-push-archive-reminder.py`, `post-pr-merge-pull.py`, `post-pr-merge-project.py`, and `usage-snapshot.py`.

Discovered while investigating #378 (the `post-tool-use.py` project-add failing in worktree sessions). This is a **distinct, deeper cause** from the gitignored-config bug fixed in #378.

## Evidence

Comparing this session (`pensive-taussig-448bc7`) to a peer Claude-managed worktree session (`happy-brown-ace7c4`) on the same machine:

| Signal | This session | Peer (happy-brown) |
|---|---|---|
| `"hookEvent":"PostToolUse"` records (exit-2 attachments) | **0** | 4 |
| `"hookEvent":"UserPromptSubmit"` records | 3 | 3 |
| `userType` | `external` | `external` |
| `version` | `2.1.181` | `2.1.181` |
| Invocation binary (settings.json) | `pyw` (all events) | `pyw` (all events) |
| Project-level `.claude/settings.json` | none | none |

- `gh issue create` (#378) in this session should have triggered `post-tool-use.py` to add the item and exit 2 (the worktree HAS `.claude/hook-config.json`; the captured tool output was `https://github.com/brownm09/dev-env/issues/378` + an `EXIT: 0` line, from which `extract_github_url` resolves the URL; the `project` OAuth scope is present). Running the hook MANUALLY with that exact payload succeeds (adds the item, prints the Impact/Why reminder, exits 2). Yet during the live `gh issue create` the item was NOT added and NO PostToolUse hook attachment was recorded.
- PreToolUse hooks DID fire in this session (the worktree-path-check hook blocked a later Write), and `pyw -3` resolves and runs here — so it is not an invocation-binary failure and not a blanket "no hooks" state. It is specifically the PostToolUse event.

## Ruled out

[...]

Issue #371, opened 2026-06-21
Document: remote git ops hang on Git Credential Manager GUI in agent/worktree sessions (+ the gh auth setup-git fix)


brownm09:
## Problem

In Claude-managed worktree sessions on Windows, every *remote* git operation (`git push`, `git fetch`, `git ls-remote`) hangs. The operation invokes the Git Credential Manager, which launches the `GitHub.UI.exe` GUI login dialog; in a non-interactive agent session that dialog never resolves, so the command blocks until timeout. In one session ~15 stuck `GitHub.UI.exe` dialogs accumulated and had to be force-killed (`taskkill //F //IM GitHub.UI.exe`), burning tokens on recovery.

Note: `gh` itself stays authenticated (keyring) the whole time — only git-over-https is affected.

## Why this matters

When unmitigated, the constraint causes manual recovery work and token waste on every occurrence. A persistent fix was applied during the originating session (2026-06-20), but it lives only as a global git-config change with no in-repo record — a future session could revert it without understanding why it exists, or a new machine could hit the hang and rediscover the workaround from scratch.

## Goal

Make the constraint and its fix durable and explained, per ADR-038 (durable decisions documented in-repo, not memory-only):

1. Add a **Git Workflow Runbooks** entry to `docs/REFERENCE.md` covering the symptom, the per-command workaround, and the `gh auth setup-git` standardization (with the revert command).
2. Decide whether a global git-config standardization warrants an **ADR** (a workflow decision `git log` won't explain) and author it if so.

## Context

Originating incident: engineering-journal `sessions/dev-env/2026-06-20_212416.stub.md` → "Platform snag (resolved this session)". Sibling precedent for a git platform constraint documented this way: ADR-035 (`git push --delete` fails in web sessions).

PR #386, opened 2026-06-22
fix(hooks): resolve canonical hook-config in worktree sessions (#378)

### LINKED ISSUE #378: post-tool-use project-add hook silently no-ops in worktree sessions (missing .claude/hook-config.json)
## Problem

Issues created with `gh issue create` in **Claude-managed worktree sessions** are not auto-added to the Dev Env GitHub Project (#3), and the `post-tool-use.py` hook's Impact/Why exit-2 reminder never surfaces. The session proceeds to file edits with no project item and no field prompt, forcing the manual project-add fallback documented in `CLAUDE.md → GitHub Project`.

Observed twice on 2026-06-20: issue #369 (originating session) and issue #371 (GCM-credential-helper docs session) both required the manual fallback.

## Root cause

`post-tool-use.py:load_config(cwd)` reads `<cwd>/.claude/hook-config.json`. That file is **gitignored** (`.gitignore` ignores the whole `.claude/` dir), so `git worktree add` never checks it out. The harness copies it into new worktrees only **inconsistently** — roughly half of the live Claude-managed worktrees, and **all** sibling worktrees (`dev-env-188`, …), are missing it.

When the config is absent, `load_config` returns `None` and the hook hits `if config is None: sys.exit(0)` — a silent exit *before* the project-add and *before* the Impact/Why exit-2 message. This is not ADR-015 (that guard only lives in the two `UserPromptSubmit` hooks; `post-tool-use.py` has no worktree guard and its output is not suppressed). When the config *is* present, `load_config` and `extract_github_url` both work correctly.

## Desired outcome

Issues created in worktree sessions are reliably added to project #3 with the Impact/Why prompt surfaced — i.e. worktree sessions behave like main-checkout sessions. The hook should resolve the canonical checkout's machine-local config when the worktree-local copy is absent.

## Impact

High — every issue created in a config-less worktree silently skips the board-add + Impact/Why workflow, costing manual recovery on each occurrence.

brownm09:
## Summary

`post-tool-use.py`'s `load_config(cwd)` read only `<cwd>/.claude/hook-config.json`. That file is **gitignored** (`.gitignore` ignores all of `.claude/`), so it is machine-local and lives only in the canonical checkout — `git worktree add` never checks it out, and the harness copies it into Claude-managed worktrees only inconsistently (≈ half of live worktrees, and **never** into sibling worktrees like `dev-env-188`). When absent, `load_config` returned `None` and the hook hit `if config is None: sys.exit(0)` — a silent exit **before** the project-add and **before** the Impact/Why exit-2 reminder, forcing the manual board-add (observed on #369, #371).

## Fix

`load_config` now falls back to the **canonical checkout's** config when the cwd-local copy is absent:

- **Claude-managed worktree** (`<root>/.claude/worktrees/<name>`) → pure regex `canonical_root_from_worktree` (mirrors the proven prefix regex in `pre-tool-use-worktree-path-check.py`; no I/O, offline-testable; preserves the cwd path format → no #334 hazard).
- **Sibling worktree** (`dev-env-188`, not under `.claude/worktrees/`) → `canonical_root_via_git` via `git rev-parse --git-common-dir`; the common dir is the canonical `.git`, whose parent is the root. The pure resolution is extracted into `_canonical_root_from_common_dir` (offline-tested); only the `subprocess.run` shells out. Degrades to `None` on any git failure.
- The cwd-local read still wins first, so main-checkout / non-worktree sessions are unchanged; a project with **no** config anywhere still silently skips (opt-in preserved).

Verified end-to-end against the real config: `canonical_root_via_git('C:/Users/brown/Git/dev-env-188')` → `C:\Users\brown\Git\dev-env`, and `load_config` on that config-less sibling resolves the real `project_number=3` config.

## Testing

Ran the dev-env `## Testing` checks relevant to this change:

- **Hook-script syntax check** (`py -3 -c "import ast…" claude/scripts/*.py`) — all parse.
- **post-tool-use test** (item #12, extended) — `py -3 claude/scripts/tests/test_post_tool_use.py`:

  `Tests: 22 passed, 0 skipped, 0 failed`

  9 pre-existing + **13 new**: `canonical_root_from_worktree` across forward-slash / backslash / subdir / main-checkout / sibling / POSIX / empty+None; a hermetic temp-dir test of `load_config`'s canonical-worktree fallback end-to-end; and 5 tests of the extracted `_canonical_root_from_common_dir` resolver (relative `.git`, absolute `<root>/.git`, whitespace-stripped stdout, non-`.git` basename, empty).

**Coverage gate (ADR-022):** new behavior is covered. Only the `subprocess.run` call in `canonical_root_via_git` is uncovered — it shells out, matching the repo's fixture-only convention (cf. `add_to_project`); also validated by a one-off smoke test against a real config-less `dev-env-188` (above).

**Suppression / test-integrity greps (ADR-026 / ADR-029):** clean. The integrity grep's only hits are `` `sys.exit(0)` `` substrings (`xit(` ⊃ `exit(`) — a known false positive. No skip markers, no deleted tests, no lowered thresholds.

`/review` (self, via Claude Code) → `blocking=0 non_blocking=1`. [Review comment](https://github.com/brownm09/dev-env/pull/386#issuecomment-4765039382).

- **[maintainability] `canonical_root_via_git` fused its subprocess call with pure path-resolution, leaving that logic untested** — **fixed in `42b01ce`**: extracted the pure `_canonical_root_from_common_dir(cwd, common)` and added 5 offline tests; updated ADR-052 and the `## Testing` item.

## ADR-052

Records the gitignored-config-in-worktree failure mode. Rebased onto `main` after #384/#385 (ADR-051) landed, so the INDEX reads `050 → 051 → 052` in order.

**Out of scope:** the separate harness-level cause where PostToolUse hooks don't execute *at all* in some sessions (#381) — orthogonal to config resolution. This PR makes the hook correct *when it runs*.

Closes #378


brownm09:
## Review: fix(hooks): resolve canonical hook-config in worktree sessions (#378)

### Summary
**Merge with minor changes.** The fix is correct and well-targeted: `load_config` now resolves the canonical checkout's gitignored `hook-config.json` when the worktree-local copy is absent, restoring the project-add + Impact/Why hook in worktree sessions (the silent `sys.exit(0)` behind the manual board-adds on #369/#371). Both worktree shapes are handled, every new failure mode degrades to the pre-existing silent skip (opt-in preserved), and the change ships 8 new offline tests plus a verified end-to-end smoke test of the git path. The single finding is non-blocking — a testability gap in the sibling-worktree resolver.

### Non-Blocking Findings

**[maintainability]** `claude/scripts/post-tool-use.py` — `canonical_root_via_git`
The function fuses the `git rev-parse --git-common-dir` subprocess call with the pure resolution of its output (relative/absolute join, `normpath`, the `.git`-basename guard, `dirname`). Because the whole function shells out, that 4-branch resolution logic has **no** automated coverage — its only proof is a one-off manual smoke test that won't run in CI or the pre-PR gate. Every other I/O helper in this repo extracts its pure core for offline testing (`install_decision`, `classify_token`, `read_command_output`, `is_successful_merge`), so this is the only path-resolution logic in the change with zero regression coverage.
**Suggestion:** Extract a pure `_canonical_root_from_common_dir(cwd, common) -> str | None` that takes the already-captured stdout string and returns the canonical root (or `None`), leaving `canonical_root_via_git` as the thin `subprocess.run` wrapper. Add offline tests for relative `.git`, absolute `<root>/.git`, a non-`.git` basename, and empty input.

### Findings Disposition
- **[maintainability] extract pure `_canonical_root_from_common_dir` + offline tests** — fixing in this PR (disposition recorded in the PR body before merge).

---

PR #385, opened 2026-06-22
fix(worktrees): skip active-session worktrees in prune/reclaim (#384)

brownm09:
## Problem

A maintenance routine removed the Claude-managed worktree `pensive-taussig-448bc7` **while a Claude Code session was actively running inside it** (2026-06-21 ~23:42; transcript's last line `23:42:10` = the removal instant). `git worktree remove` deleted the `.git` link + admin dir and deregistered the worktree; the orphaned session then had every `Write`/`Edit` blocked by the ADR-024 guard and bare `git` silently resolved to the canonical main checkout.

Closes #384.

## Root cause

`prune-merged-worktrees.py` is the only remover (`reclaim-worktree-disk.py` / `post-pr-merge-reclaim.py` only strip regenerable artifacts). Two factors combined:

1. `is_merged()` runs `git merge-base --is-ancestor <branch> origin/main`, which is **true when the branch tip *is* origin/main** (a commit is its own ancestor) — so a fresh worktree with 0 commits is classified "merged."
2. The only liveness protection was `path == os.getcwd()`, which shields just the *pruning process's own* worktree. The `prune-stale-worktrees` routine runs out-of-process and passes no `--protect-cwd`, so it cannot see a live session in any *other* worktree.

## What changed

- **New shared helper `claude/scripts/_worktree_liveness.py`** — `worktree_session_is_live()` skips a worktree whose Claude transcript (`~/.claude/projects/<slug>/*.jsonl`, slug = cwd with `:`/`\`/`/`/`.`→`-`) was written within a window. Import-only (offline-testable). Fail-safe: a missing/empty transcript dir ⇒ not-live ⇒ eligible (unchanged behavior for idle worktrees). Exact-slug lookup with a unique-basename suffix fallback hedging an encoding change; recursive `*.jsonl` so an active `subagents/` keeps a worktree live.
- **`prune-merged-worktrees.py`** — guard with a **24h** window (removal severs the session → long guard).
- **`reclaim-worktree-disk.py`** — guard with a **6h** window (stripping node_modules is self-healing and must stay aggressive vs ENOSPC; matches the 6-hourly routine). `--protect-cwd` only covers *this* session; the guard covers live sessions in *other* worktrees.
- Both windows override-able via `--liveness-window-min N`. The guard sits before the merged/dirty checks, so its skip reason wins. **It only ever adds skips — never removes/strips more than before.**
[...]

Issue #383, opened 2026-06-22
Prune/reclaim routine deleted a LIVE session's worktree, orphaning it mid-task


brownm09:
## Problem

A reclaim/prune routine deleted a Claude-managed worktree **while a Claude Code session was actively running inside it**, orphaning the session mid-task.

Observed 2026-06-21, worktree `pensive-taussig-448bc7`. During a live session whose cwd was `C:/Users/brown/Git/dev-env/.claude/worktrees/pensive-taussig-448bc7`:

- the `.git` link file in the worktree was removed,
- the admin dir `C:/Users/brown/Git/dev-env/.git/worktrees/pensive-taussig-448bc7/` was deleted,
- the worktree was deregistered from `git worktree list`.

Earlier in the same session, writes to the worktree succeeded — so the prune happened mid-session, not before it.

## Impact (High)

The session was orphaned: the ADR-024 worktree-path PreToolUse guard then (correctly) blocked **all** Write/Edit, and bare `git` from the session cwd silently resolved to the **canonical** main checkout (`git rev-parse --show-toplevel` → `C:/Users/brown/Git/dev-env`) — unsafe to run. The session could no longer edit files or safely run git, forcing a handoff to a fresh session.

## Investigate / fix

Determine which routine pruned an **unmerged, actively-used** worktree (its branch `claude/pensive-taussig-448bc7` was at origin/main — not merged-and-gone): `claude/scripts/prune-merged-worktrees.py`, `claude/scripts/reclaim-worktree-disk.py`, and any prune routine under `claude/routines/`.

Add a **liveness guard** so a worktree with an active session is skipped — e.g. recent activity in its transcript dir `C:/Users/brown/.claude/projects/<worktree-slug>/*.jsonl` (mtime within N minutes), or a session lock/heartbeat — **in addition to** (not instead of) the existing merged-branch check.

## Related

ADR-024 (worktree-path guard / orphaned-worktree #328), ADR-037 (worktree disk reclamation), ADR-045 (pre-install free-space gate). Discovered alongside #381 (PostToolUse hooks not firing) during the #378 investigation.

PR #387, opened 2026-06-22
docs(adr): document PostToolUse-inert-in-background-sessions limitation (ADR-053)

### LINKED ISSUE #381: PostToolUse hooks do not execute in some sessions (UserPromptSubmit/PreToolUse do) — harness-level

brownm09:
## Summary

Documents a **third, distinct** `post-tool-use.py` failure mode (after [ADR-049](docs/adr/049-hook-payload-output-field.md)/[ADR-050](docs/adr/050-shared-hookio-sibling-hook-fixes.md) and #378): in some sessions **every** PostToolUse hook is silently inert, while `UserPromptSubmit`, `PreToolUse`, and `Stop` hooks from the same `settings.json` fire normally. Root cause is an **upstream Claude Code Desktop harness behavior** tied to how the session was launched (background task / `spawn_task` vs. interactive) — **no hook-code change in this repo can fix it**, so the deliverable is a documented limitation + recovery, not a code change.

Closes #381.

## Root cause (verified)

PostToolUse settings hooks do not execute in **background-task / SDK-launched** sessions; the other three event classes still run.

Evidence chain:
1. **Confirmed asymmetry** — affected session `pensive-taussig` (`ea0b2f4e`): **0** PostToolUse attachment records vs. an interactive peer's 4; UserPromptSubmit/PreToolUse/Stop fired in both.
2. **Rules out the #377/#380 `output`→`stdout` bug** — the affected session created issues #378 (*before* those fixes merged to `main`), #381, and #383 (*after*); **all** `gh issue create` calls were PostToolUse-inert. Hooks run from `~/.claude/scripts` (junctioned to `main`), so the code changed under the session while the inert behavior did not → a per-session launch state, not a code version. Peers emitted PostToolUse attachments even with the pre-#377 code (the output-field bug has a different, non-silent signature).
3. **Launch fingerprint** — the affected session's `stop_hook_summary` carries two `{"command":"callback"}` in-process hooks (SDK-injected); the interactive peer has no `system` records at all.
4. **Second inert integration, same session** — `mcp__ccd_session__spawn_task` was called but its chip never rendered: a different desktop integration that also depends on the foreground UI, isolating the common cause to the background/SDK launch context.
5. **Upstream corroboration (verified to exist)** — [anthropics/claude-code#42336](https://github.com/anthropics/claude-code/issues/42336) "[BUG] PostToolUse hooks not triggering in Desktop App" and [#53494](https://github.com/anthropics/claude-code/issues/53494) "Cloud-spawned tasks … invisible in desktop UI".
6. **Intermittent within the background class** — a scan of ~745 transcripts shows most background/SDK sessions *do* fire PostToolUse, so it is not deterministically reproducible from config. The `settingSources`-skips-the-file theory is ruled out: the same file's other-event hooks load and run.

## What changed (docs only)

- **`docs/adr/053-…md`** — new ADR: mechanism, detection signature, recovery, links to #381 + the two upstream issues + ADR-049/050.
- **`docs/adr/INDEX.md`** — ADR-053 row.
- **`docs/REFERENCE.md`** — a note under `### PostToolUse` flagging the background/SDK-session limitation.
- **`CLAUDE.md`** — a pointer from the GitHub Project § Fallback (the manual project-add *is* the recovery).

## Disposition & proposed follow-up

A true fix can only land upstream. A reliable-event safety net (a `Stop`/`UserPromptSubmit` hook that surfaces a `gh issue create`/`gh pr merge` whose PostToolUse side-effect did not occur) would convert the silent gap into a visible advisory — proposed as a **follow-up** (separate feature, own test burden; recovery already exists), not built here.

## Testing

Docs-only change — no runtime, no `.py`/`.sh`/`.json` touched, so none of the dev-env `## Testing` script self-tests apply (each is gated on changing its target script). Verification run (the reference-integrity equivalent per the project Observability convention):

- **ADR reference integrity:** 129 ADR markdown links across the 4 changed files checked → **0 broken**.
- **Suppression grep** (`git diff origin/main`): clean.
- **Test-integrity grep:** clean.
- **INDEX ordering:** `050 → 051 → 053` (052 reserved for concurrent open PR #386).

**Coverage gate (ADR-022):** N/A — no code or testable behavior changed; documentation only.

## Notes for the reviewer

- **ADR numbering:** originally drafted as ADR-052, **renumbered to 053** because the concurrent open PR #386 (#378 worktree-config fallback) already adds `052-worktree-config-canonical-fallback.md`. This branch's INDEX therefore reads `050 → 051 → 053`; #386's 052 row slots in on its merge.
- PR #385 (ADR-051, worktree-liveness-guard) merged to `main` mid-session; this branch was fast-forwarded onto it before re-applying the edits, so ADR-051 is present with no gap below 053.
- **Risk audit (Pass 3):** Testing — reference-integrity (above); Observability — the ADR *improves* it, converting a silent automation gap into a documented, recognizable signature; Security / Performance / Data-integrity / Accessibility — N/A (docs only); Resilience — documents the manual recovery path.

`/review` (claude-opus-4-8): **0 blocking, 1 non-blocking** — all addressed in-PR.

- **[correctness]** ADR Evidence §, point 1 — peer-timing overclaim ("Peer sessions emitted PostToolUse attachments even with the pre-#377 code"): **fixed in `11cdf2f`**. Reframed to the supported distinction — the #377 output-field bug suppresses `post-tool-use.py`'s URL *extraction*, not PostToolUse *dispatch* (it exits 2 with a "no URL" advisory = an attachment), versus the missing-attachment signature seen here — and dropped the unverified timing clause. The load-bearing argument (affected session inert before AND after the fixes) is unchanged.


brownm09:
## Review: docs(adr): document PostToolUse-inert-in-background-sessions limitation (ADR-053)

### Summary
**Merge with one minor revision.** Docs-only ADR (137 lines, 4 files) recording an upstream Claude Code Desktop limitation: PostToolUse settings hooks don't dispatch in background/`spawn_task`-launched sessions while the other event classes do. Reference integrity (129 ADR links, 0 broken), INDEX update, and the gate checks all pass — coverage/integrity gates are N/A (docs-only), doc-reconciliation is satisfied (no `claude/{skills,hooks,scripts,routines}` touched; the `docs/adr/INDEX.md` index *is* updated). The cited Claude Agent SDK hooks reference was verified to resolve and to support the hook-model claims (callback hooks, additive `settingSources`-gated merge). One non-blocking accuracy nit; the load-bearing argument is sound.

---

### Non-Blocking Findings

**[correctness]** `docs/adr/053-posttooluse-hooks-inert-in-background-sessions.md` (Evidence §, point 1 — last sentence)
The clause "Peer sessions emitted PostToolUse attachments even with the *pre*-#377 code" asserts that the peer's attachments fired *before* the #377 fix merged (06-21 00:21 EDT). The evidence actually gathered is a peer attachment *count* (4), not timestamped against that merge — so the "pre-#377" timing is not established. This is a corroborating sentence only; the load-bearing argument (the affected session produced **zero** PostToolUse attachments across `gh issue create` calls spanning both before #378 and after #381/#383 the fixes) stands without it.
**Suggestion:** Soften to what's supported — the output-field bug suppresses `post-tool-use.py`'s URL *extraction*, not PostToolUse *dispatch*, so its signature is a fired-but-wrong hook (exit 2 with a "no URL" advisory), not a missing attachment — and drop the unverified peer-timing clause.

---

### Style Notes
None — the ADR is clearly structured and consistent with the ADR-049/050 sibling format.

---

### Findings Disposition
- **[correctness]** peer-timing overclaim → **fixed in this PR** (follow-up commit; see the "Review findings disposition" section of the PR body).

---

PR #402, opened 2026-06-22
chore(scripts): track merge-ready.sh + commit settings UI prefs

brownm09:

`/review` by `claude-opus-4-8` found 1 blocking + 1 non-blocking finding; both fixed in commit `e74b409`.

- **[correctness] `sum(p.statusCheckRollup)` null TypeError** — fixed in `e74b409`: changed to `sum(p.statusCheckRollup || [])`. JS default params only trigger on `undefined`; GitHub returns `null` for repos with no CI checks, causing `for (const x of null)` to throw.
- **[reliability] no EXIT trap** — fixed in `e74b409`: added `trap 'rm -f "$TMPFILE"' EXIT` after TMPFILE assignment so temp file is cleaned up even on early exit.


brownm09:
## Review: chore(scripts): track merge-ready.sh + commit settings UI prefs

### Summary

Nearly merge-ready — one correctness fix needed before squash. The `sum(p.statusCheckRollup)` call will crash with a `TypeError` for any PR from a repo that has never had a CI workflow (GitHub returns `null`, not `[]`, for `statusCheckRollup` in that case, and JS default parameters only apply to `undefined`). Fix is a one-character addition; everything else — the temp-path fix, the settings commit, the REFERENCE.md placement — is correct and well-justified.

---

### Blocking Findings

**[correctness]** `claude/scripts/merge-ready.sh:38`
`sum(p.statusCheckRollup)` is called with the raw API value. JavaScript default parameters (`roll=[]`) only fire when the argument is `undefined`; GitHub returns `null` for `statusCheckRollup` on a PR from a repo with no CI checks at all. `for (const x of null)` then throws `TypeError: null is not iterable`, the `node -e` process exits non-zero, and `set -e` aborts the loop — without reaching the `rm -f` at line 47. The smoke-test ran against `brownm09/dev-env`, which has checks, so this branch was never exercised.
**Fix:** Change `sum(p.statusCheckRollup)` → `sum(p.statusCheckRollup || [])`.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/merge-ready.sh:14`
No `EXIT` trap is set after `TMPFILE` is defined. If either `gh pr list` (network failure, auth error) or the `node -e` invocation exits non-zero mid-loop, `set -e` causes the script to abort before reaching the `rm -f "$TMPFILE"` at line 47. The scratch dir is user-accessible and the files are small, but the cleanup intention is clearly stated in the comment at line 12 and the explicit `rm` at line 47 — an inconsistency.
**Suggestion:** Add `trap 'rm -f "$TMPFILE"' EXIT` on the line after `TMPFILE=` is assigned (line 14). The explicit `rm` at line 47 can stay or be removed — the trap fires either way.

---

### Findings Disposition

Per ADR-028, before `gh pr merge` record each finding's disposition in a "Review findings disposition" section of the PR body:

- **[correctness] `sum(null)`** — fix in-PR: `sum(p.statusCheckRollup || [])` (1-char change)
- **[reliability] no EXIT trap** — fix in-PR or file and link

---

PR #416, opened 2026-06-28
feat(hooks): add post-merge tile checkpoint hook

brownm09:
## Summary

- Adds `claude/scripts/post-merge-tile-checkpoint.py`: a PostToolUse Bash hook that fires on every successful `gh pr merge` and exits 2 with a blocking reminder to spawn follow-up tiles via `spawn_task`
- Enforces ADR-046's "Capture post-merge follow-ups as tiles" rule mechanically — same model as `pr-merge-reminder.py`
- Uses `is_successful_merge` (exit 0 or stdout marker) so worktree merges that exit non-zero are also covered
- Adds test, ADR-060, settings.json entry, and doc updates per Documentation Maintenance table

Closes #415

## Testing

Hook syntax check:
```
py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py
```
→ all scripts syntax OK

Safe-exit on empty stdin:
```
py -3 claude/scripts/post-merge-tile-checkpoint.py < /dev/null; echo "exit: $?"
```
→ exit: 0

Unit test:
```
py -3 claude/scripts/tests/test_post_merge_tile_checkpoint.py
```
→ Tests: 4 passed, 0 skipped, 0 failed

## Suppression / test-integrity check

No suppressions, no skip markers, no deleted tests, no threshold changes.

## ADR

ADR-060 documents this decision: `docs/adr/060-post-merge-tile-checkpoint-hook.md`.

Issue #275, opened 2026-05-27
post-pr-merge-pull hook skips local main update when gh pr merge runs from a worktree


brownm09:
## Problem

`claude/scripts/post-pr-merge-pull.py` is meant to fast-forward local `main` after every `gh pr merge`, so the canonical clone stays current without waiting for the next `UserPromptSubmit` `dev-env-sync` pass. In practice, **it never fires for merges run from a worktree** — which is the documented workflow for most repos.

## Root cause

`post-pr-merge-pull.py:114`:

```python
# Only pull when the merge actually succeeded
if exit_code != 0:
    sys.exit(0)
```

When `gh pr merge` runs from a worktree, the *remote* merge succeeds but `gh`'s local cleanup steps fail with:

```
failed to run git: fatal: 'main' is already used by worktree at 'C:/Users/brown/Git/<repo>'
```

`gh` exits 1. CLAUDE.md → Git Workflow → "Merging a PR developed in a worktree" explicitly documents this as a benign, expected failure mode. But the hook trusts the exit code and bails before calling `pull_main()`.

## Observed impact

- **dev-env:** masked by the `UserPromptSubmit` `dev-env-sync` hook (ADR-006), which catches up on the next prompt. Today's session: PR #274 merged from a worktree, local `main` was stale until `dev-env-sync` pulled `f56e95f` on the next prompt.
- **Other repos (engineering-journal, lifting-logbook, engineering-playbooks):** no equivalent `UserPromptSubmit` syncer. Local `main` stays stale until the next manual `git pull` or the next session that touches the repo. New worktrees cut from a stale local `main` will diverge.

## Suggested fix

Trust `gh`'s stdout success marker instead of (or in addition to) the exit code:

```python
output = data.get(\"tool_response\", {}).get(\"output\", \"\")
merge_succeeded = (
    exit_code == 0
    or \"Merged pull request\" in output
    or \"Squashed and merged\" in output
    or \"Rebased and merged\" in output
)
if not merge_succeeded:
    sys.exit(0)
```

`gh pr merge` prints \"✓ Squashed and merged pull request\" (or rebased/merged variant) on success regardless of the local cleanup outcome, so the stdout check is reliable and adds no network call.

## Acceptance criteria

- [ ] Hook fires `pull_main()` after a successful `gh pr merge` even when `gh` exits non-zero due to worktree-conflict errors.
- [ ] Hook still skips on actual merge failures (e.g., merge conflicts, auth errors) — no false positives.
[...]

PR #420, opened 2026-06-28
fix(hooks): pr-merge-reminder fires on worktree merges

### LINKED ISSUE #419: pr-merge-reminder.py silently misses worktree merges (exit non-zero on local cleanup)
## Problem

`pr-merge-reminder.py` has a global early-exit guard at line 279:

```python
exit_code = data.get("tool_response", {}).get("exitCode", 0)
if exit_code != 0:
    sys.exit(0)
```

When `gh pr merge` is run from a worktree, the remote GitHub merge succeeds but the local checkout cleanup fails with `fatal: 'main' is already checked out at ...`. This causes `gh pr merge` to exit non-zero, and the global guard silently swallows the event — the journal-update reminder never fires.

## Root Cause

The guard is applied unconditionally to all commands (`gh pr create`, `gh pr merge`, `git push`) before we know which command triggered the hook. Worktree merges are a known case where exit non-zero ≠ failed (issue #275).

## Fix

The sibling hook `post-merge-tile-checkpoint.py` (added in PR #416) already handles this correctly via the `is_successful_merge` predicate (`exit_code == 0 OR output_has_merge_marker(output)` from `_hookio`).

For `pr-merge-reminder.py`:
- Remove the global `exit_code != 0` bail
- Apply per-command success guards:
  - `gh pr merge`: check `exit_code == 0 or output_has_merge_marker(output)`
  - `gh pr create` and `git push`: keep the strict `exit_code == 0` guard

Extract the merge gate as a named pure predicate (`_is_successful_merge_call`) for testability, following the same pattern as the sibling hook.

brownm09:
## Problem

`pr-merge-reminder.py` had a global early-exit guard:

```python
exit_code = data.get("tool_response", {}).get("exitCode", 0)
if exit_code != 0:
    sys.exit(0)
```

When `gh pr merge` runs from a worktree, the remote GitHub merge succeeds but the local checkout cleanup fails with `fatal: 'main' is already checked out at ...`. This exits non-zero, and the global guard silently dropped the event — the journal-update reminder never fired.

## Fix

Applies the same pattern as `post-merge-tile-checkpoint.py` (ADR-060, PR #416):

- Removed the global `exit_code != 0` bail
- Added per-command guards after command detection:
  - `gh pr merge`: `exit_code == 0 or output_has_merge_marker(output)` — imports `output_has_merge_marker` from `_hookio`
  - `gh pr create` / `git push`: strict `exit_code == 0` (unchanged semantics)
- Extracted `_is_successful_merge_call(exit_code, output)` as a named pure predicate for testability

## Testing

**Hook-script syntax check:**
```
py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/pr-merge-reminder.py claude/scripts/tests/test_pr_merge_reminder.py
```
→ `Syntax OK`

**pr-merge-reminder test suite** (4 new cases covering `_is_successful_merge_call`, 16 existing cases unchanged):
```
```
→ `Tests: 20 passed, 0 skipped, 0 failed`

## ADR-warrant check

No new ADR needed. This is a bug fix extending the pattern already established in ADR-060 (`output_has_merge_marker` for worktree-merge detection). Rationale is recoverable from ADR-060 and issue #275.

## Suppression / test-integrity greps

No suppressions, skip markers, deleted tests, or lowered thresholds.

Closes #419

### DIFF
--- a/claude/scripts/pr-merge-reminder.py
+++ b/claude/scripts/pr-merge-reminder.py
@@ -30,7 +30,7 @@ import subprocess
 import sys
 from collections.abc import Callable
 
-from _hookio import read_command_output
+from _hookio import output_has_merge_marker, read_command_output
 
 # Matches the start of a statement token against `gh pr merge`, `gh pr create`,
 # or `git push`.
@@ -262,6 +262,16 @@ def _create_shard_step(output: str) -> str:
     )
 
 
+def _is_successful_merge_call(exit_code: int, output: str) -> bool:
+    """Return True iff a gh pr merge call completed the remote merge.
+
+    Worktree merges exit non-zero on local cleanup (e.g. 'main is already
+    checked out') even when the remote merge succeeded — check the success
+    marker too (issue #275 behaviour, mirrors post-merge-tile-checkpoint.py).
+    """
+    return exit_code == 0 or output_has_merge_marker(output)
+
+
 def main() -> None:
     raw = sys.stdin.read().strip()
     if not raw:
@@ -275,12 +285,9 @@ def main() -> None:
     if data.get("tool_name") != "Bash":
         sys.exit(0)
 
-    exit_code = data.get("tool_response", {}).get("exitCode", 0)
-    if exit_code != 0:
-        sys.exit(0)
-
     command = data.get("tool_input", {}).get("command", "")
     cwd = data.get("cwd", "<unknown>")
+    exit_code = data.get("tool_response", {}).get("exitCode", 0)
 
     is_create = is_pr_create_command(command)
     is_merge = is_pr_merge_command(command)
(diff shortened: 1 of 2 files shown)


brownm09:
## Review: fix(hooks): pr-merge-reminder fires on worktree merges

### Summary

Ready to merge. The fix is surgical and correct: the global `exit_code != 0` guard is replaced with a per-command guard that correctly applies `output_has_merge_marker` for `gh pr merge` while leaving the strict exit-0 gate intact for `gh pr create` and `git push`. The approach mirrors `post-merge-tile-checkpoint.py` (ADR-060) exactly, and four new tests cover all meaningful branches of `_is_successful_merge_call`. No blocking or non-blocking findings.

---

### Documentation Reconciliation

Changed paths are under `claude/scripts/` but are modifications to existing scripts (not add/remove/rename), so the Documentation Maintenance table entry does not trigger. The REFERENCE.md entry for `pr-merge-reminder.py` ("fires on `gh pr merge`") remains accurate — the fix makes the hook match its documented intent, not change it.

No README exists in `claude/scripts/` or `claude/scripts/tests/` — no coverage gap.

---

### Test Coverage and Integrity

PR body includes `Tests: 20 passed, 0 skipped, 0 failed`. The diff adds four new test cases covering: clean exit (exit 0), worktree merge with success marker (exit 1 + marker), genuine failure (exit 1, no marker), and exit-0-trumps-no-marker. All branches of `_is_successful_merge_call` are exercised. No skip markers, deleted tests, lowered thresholds, or bypass flags in the diff.

---

PR #443, opened 2026-06-30
[fix] Scope pr-merge-reminder git-push reminder to the push-target repo

### LINKED ISSUE #442: [fix] pr-merge-reminder fires misattributed journal reminders for cross-repo pushes
## Problem

`claude/scripts/pr-merge-reminder.py` (PostToolUse Bash hook) fires a journal-stub
reminder on every `git push` whenever the session's **cwd repo** has an open PR — but it
keys off `cwd` and `_open_pr_for_cwd(cwd)`, **not** off which repo the push actually
targeted.

### Symptom

During a session whose cwd was `lifting-logbook` (open PRs #622 and #628), every
`git push` to a *different* repo — `dev-env` and `engineering-journal` — fired the
lifting-logbook journal reminder. It fired ~6 times in one session on pushes that had
nothing to do with lifting-logbook, including the journal-stub pushes themselves. Pure
noise: the reminder names a PR unrelated to what was pushed.

### Root cause

In `main()`, the `is_push and not (is_create or is_merge)` branch calls
`_open_pr_for_cwd(cwd)` and fires if any open PR exists for the **cwd** repo. It never
checks whether the `git push` actually targeted the cwd repo's remote. A push to another
repo via `cd <other-repo> && git push` still trips it, and the `_EJ_REPO_FRAGMENT` skip
(intended to silence engineering-journal pushes) checks `cwd`, not the push target, so it
also misses.

ADR-021 (which introduced this branch) assumed `cwd` == the pushed repo and explicitly
accepted the per-push subprocess cost; cross-repo pushes were never considered.

## Goal

The push reminder should be both **correctly attributed** and **quiet**:

1. **Scope to the push target.** Evaluate the open-PR lookup against the directory the
   push actually runs in (honor a `cd <path> && … git push` chain), not the session cwd.
   This naturally routes engineering-journal pushes into the existing skip.
2. **Once per PR per session.** The reminder is idempotent guidance — fire it at most once
   per open PR per session (per-session sentinel) instead of on every push.

The hook must stay non-blocking-safe: any internal error exits 0 and never crashes the
push flow.

## Acceptance criteria

- [ ] A push that targets a *different* repo than the session cwd does not fire that cwd's
      reminder.
- [ ] A given open PR's push reminder fires at most once per session.
- [ ] Offline unit test (`tests/test_pr_merge_reminder.py`) pins both behaviors.
- [ ] Test registered in the dev-env `CLAUDE.md` `## Testing` section.
- [ ] ADR-warrant check run; firing-semantics change recorded.

brownm09:
## Summary

`pr-merge-reminder.py`'s `git push` branch called `_open_pr_for_cwd(cwd)`, keyed off the
**session cwd**, never checking which repo the push actually targeted. From a `lifting-logbook`
cwd (open PRs #622/#628), every `cd <other-repo> && git push` to dev-env / engineering-journal
fired a **misattributed lifting-logbook reminder** — ~6× in one session, including on the
journal-stub pushes themselves. The `_EJ_REPO_FRAGMENT` skip also checked `cwd`, not the push
target, so it missed. This was a refinement gap in ADR-021 (which assumed cwd == the pushed repo).

The defect was **misattribution**, not the per-push cadence. The fix scopes the lookup to the
repo the push targets; the reminder still fires on every qualifying push (each carries new
journalable content), now attributed to the correct repo.

## What changed

1. **Scope to the push target** — `_effective_push_dir(command, cwd)` parses a
   `cd <path> && git push` (or `;`) chain and the open-PR lookup runs against that directory.
   A cross-repo push is evaluated against *its* repo; engineering-journal pushes now route into
   the existing EJ skip. Conservative: any shape it can't parse falls back to `cwd` (pre-ADR-021
   behavior) — never a wrong-repo positive, and a mis-resolved dir is a silent no-op downstream.
2. **Every-push firing retained (no dedup)** — once attribution is correct, the reminder fires
   on every push to an open-PR branch *in the repo actually pushed*, matching ADR-021's
   update-trigger intent (a multi-push review cycle wants a nudge per push). A
   once-per-PR-per-session sentinel was considered and **rejected** — it would suppress exactly
   those wanted later-push nudges; scoping alone removes the cross-repo noise.
3. **Non-blocking hardening** — `__main__` wrapped in `try/except -> exit 0` (it had none before);
   `sys.exit(2)` still propagates as `SystemExit`, so the reminder path is unchanged.

ADR-065 (refines ADR-021) documents the decision; ADR-021 `Refined by`, `docs/adr/INDEX.md`, the
`docs/REFERENCE.md` hook row, and the CLAUDE.md `## Testing` item 28 are updated to match.

## Testing

`py -3 claude/scripts/tests/test_pr_merge_reminder.py` — offline (no network/gh):

```
Tests: 27 passed, 0 skipped, 0 failed
```

7 tests pin `_effective_push_dir` (bare->cwd, `cd <repo> && push`->that repo, multi-segment
chain, quoted path, relative-resolved, `;` chain, cd-after-push ignored). The live
`_open_pr_for_cwd` subprocess boundary is not covered (repo convention: no subprocess mocks).

Hook-script syntax check (`ast.parse` over `claude/scripts/*.py`): **SYNTAX OK**.

### Test-integrity note
The 2 sentinel tests (`test_push_reminder_once_per_pr_per_session`,
`test_push_sentinel_filename_is_per_pr_and_session`) were **removed because the sentinel they
covered was removed** (the canonical "deleted tests for removed functionality" case under
ADR-029 Rule 1). Net count 29 → 27, all passing; no skip markers added, no thresholds lowered.
The test-integrity grep's only matches are the documented `xit\(` ⊂ `exit(` false positive on
`sys.exit(...)` control-flow lines. Suppression grep clean. Lockfile N/A (no `package.json`).

## ADR-warrant
Warranted — changes hook firing semantics documented in ADR-021. **ADR-065** added (refines
ADR-021).

## Acceptance criteria
- [x] A push targeting a different repo than cwd does not fire that cwd's reminder.
- [x] A push to an open-PR branch still nudges on every push, attributed to the correct repo.
- [x] Offline unit tests pin the scoping behavior.
- [x] Test registered in the CLAUDE.md `## Testing` section.
- [x] ADR-warrant check run; firing-semantics change recorded (ADR-065).

Closes #442

---

Self-review via `/review` (`claude-opus-4-8`) — [review comment](https://github.com/brownm09/dev-env/pull/443#issuecomment-4846622054). **0 blocking, 1 non-blocking, 1 question.**

- **[correctness] ADR-065 limits omission** — a relative `cd` in a multi-`cd` chain (`cd /a && cd b && git push`) resolves against cwd, not the preceding `/a`. **Fixed in 16ef97d** (ADR-065 "Trade-off / limits" bullet + code comment; behavior unchanged).
- **Question — reminder cadence** — **Resolved**: maintainer chose to re-fire on every push rather than dedup once-per-session. **Implemented in 894eaf3** — the once-per-PR-per-session sentinel (helpers, constant, `_hookutil` import, dedup gate, and its 2 tests) was removed; the reminder now fires on every qualifying push, scoped to the correct repo. ADR-065 reframes the sentinel as the rejected alternative.


brownm09:
## Review: [fix] Scope pr-merge-reminder git-push reminder to the push-target repo, once per PR per session

### Summary
Merge-ready. This is a clean, well-scoped fix for the cross-repo misattribution in #442: the open-PR lookup now keys off the repo the push *targets* (`_effective_push_dir`) instead of the session cwd, and a per-PR-per-session sentinel collapses the ~6×/session noise to ≤1×. The `__main__` `try/except → sys.exit(0)` wrapper is a strict reliability improvement — the hook previously had no top-level guard. All gates pass: tests are included and green (29 passed, 0 skipped), doc-reconciliation and ADR-coverage are satisfied (REFERENCE.md hook row + INDEX.md + ADR-065 + ADR-021 `Refined by` cross-ref are all in the diff), and the conservative-fallback design is genuinely backed by `_open_pr_for_cwd`'s `except Exception: return None`, so a mis-resolved push dir is a real silent no-op rather than a wrong-repo positive. No blocking findings. The one point worth your call is the once-per-session reminder granularity (Question below).

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/pr-merge-reminder.py:275-278`
In a multi-`cd` chain with a *relative* second segment — `cd /a && cd b && git push` — the parser keeps the last `cd` token (`b`) and resolves it against `cwd` (→ `cwd/b`), not against the preceding `/a` (→ `/a/b`). ADR-065's "Trade-off / limits" list enumerates the quoting and `git -C` escape hatches but not this one, so the documented limit set is slightly incomplete. Practical impact is negligible and in the safe direction: a non-existent `cwd/b` makes `_open_pr_for_cwd` raise → `None` → no reminder; a wrong-repo *positive* would require `cwd/b` to coincidentally be a checked-out repo with its own open PR.
**Suggestion:** Add this case to the ADR-065 limits list (or a one-line comment at the `finditer` loop) so the known-limit set is complete. Threading the running directory through each `cd` is not worth the complexity for a shape this rare.

---

### Questions for Author

**Question:** Is once-per-PR-per-session the granularity you want for the push reminder, given that successive pushes in a review cycle (initial push → push after review-fix-1 → push after review-fix-2) each carry *new* journalable content but only the first now fires the nudge?
**Context:** The `is_push` branch (`pr-merge-reminder.py:410-437`) marks the sentinel on the first push for a PR and suppresses every later push for that PR in the same session. ADR-021's update-trigger framing ("PR updated → update the journal immediately") fired on every push; ADR-065 deliberately narrows that to once-per-session on the grounds that the reminder is idempotent guidance.
**Tradeoffs:** (a) Keep once-per-session (current) — kills the 6× noise cleanly and relies on Claude keeping the stub current without a re-nudge after later pushes; (b) reset the sentinel on a materially-later push (or key on push count) — re-nudges each push but reintroduces some of the noise the ADR set out to remove; (c) once-per-session for stub/EJ repos but per-push for the feature repo — more moving parts, and the EJ skip already covers the dominant stub case. The ADR picked (a); I'm flagging it because reminder cadence is a journaling-workflow preference you own, not something the code can settle.

---

### Style Notes
The diff matches surrounding conventions. The `_hookutil` import sits between `_winsubp` (which must load first to suppress console windows) and the stdlib group — that is this file's established order, not a deviation. New-helper docstrings are thorough and consistent with the existing ones.

---

### Findings Disposition
- **[correctness] ADR-065 limits omission** — to be fixed in this PR (add the multi-`cd`-relative case to the limits list / loop comment).
- **Question (reminder granularity)** — pending author decision; default is to keep ADR-065's once-per-session unless you prefer otherwise.

---

PR #451, opened 2026-06-30
[feat] reconcile-project-board: backstop for issues that never reach project #3 (ADR-068)

brownm09:
## What & why

Backstop for the gap [ADR-053](docs/adr/053-posttooluse-hooks-inert-in-background-sessions.md) documents: `post-tool-use.py` auto-adds each new dev-env issue to the **Dev Env** board (#3) and prompts for Impact/Why — but PostToolUse hooks are **inert in background / `spawn_task` / SDK-launched sessions**, so issues filed there are silently never boarded (motivating instance: #434/#435/#436, filed by the background memory-audit session #363).

Adds a **config-driven, add-only + report-only** reconcile engine plus a **nightly routine**, mirroring the `prune-merged-worktrees.py` (script) + `prune-stale-worktrees` (routine) pattern:

- **`claude/scripts/reconcile-project-board.py`** — reads `.claude/hook-config.json` (the same config `post-tool-use.py` reads, so field IDs never drift), lists open issues + board items, computes the **set difference** (orphans = open issues not on the board), adds each orphan via `gh project item-add`, then **surfaces** the orphans + any pre-existing **open** board items still missing a required field, printing the exact `gh project item-edit` commands. It **never guesses** a field value and **never** mutates single-select *options* (so it can't trip the option-mutation hazard). `--dry-run` reports without adding.
- **`claude/routines/reconcile-project-board/`** — nightly `0 6 * * *` routine: Step 0 `sync-routine-worktree`, run the script live, push-notify when issues still need Impact/Why **or** when an add failed.
- **ADR-067** + INDEX; README (Routines), REFERENCE (Routines + Utilities), CLAUDE.md (`## Testing` #29).

`Closes #447`

## Testing

- **Syntax check** (`## Testing` #1): `py -3 -c "import ast..." claude/scripts/*.py` → **SYNTAX OK**.
- **New unit test** (`## Testing` #29): `py -3 claude/scripts/tests/test_reconcile_project_board.py` → **`Tests: 15 passed, 0 skipped, 0 failed`**. Covers `canonical_repo_root`, `field_key`, `colliding_required_fields`, `board_issue_numbers` filtering, the `compute_orphans` set difference, `item_missing_fields` / `board_items_missing_fields`, `looks_like_scope_error`, `is_truncated`, and `render_report` — including the **no-guessing contract** (emits `item-edit` commands, never assigns a value), the `RESULT:` line, and a partial-add-failure scenario (header vs. RESULT consistency).
[...]

Issue #447, opened 2026-06-30
[feat] reconcile-project-board: backstop for dev-env issues that never reach project #3 (inert PostToolUse hooks in background sessions)


brownm09:
## Problem

The `post-tool-use.py` PostToolUse hook auto-adds each newly-created dev-env issue to
project #3 (the **Dev Env** board) and prompts for its Impact / Why fields. But **PostToolUse
hooks are inert in background / `spawn_task` / SDK-launched sessions** ([ADR-053](docs/adr/053-posttooluse-hooks-inert-in-background-sessions.md)).

So any issue filed from such a session is **silently never added to the board** — it misses its
Impact rating, its Why description, and the board's Status workflow entirely. The drift is
invisible until someone notices an issue isn't tracked.

## Concrete instance (2026-06-30)

The cross-project memory audit ([#363](https://github.com/brownm09/dev-env/issues/363)) ran in a
background session and filed #434, #435, #436. All three had `projectItems: []` — none were on the
board. A later session porting them had to manually `gh project item-add` + set Impact/Why for all
three before work could proceed. This recurs for any audit- or automation-filed issue. (Live check
while scoping this issue found #439 and #368 already orphaned, confirming the leak is ongoing.)

## Proposed direction

A **config-driven, add-only + report-only** reconcile engine plus a nightly backstop routine — built
to mirror the existing `prune-merged-worktrees.py` (script) + `prune-stale-worktrees` (routine)
pattern:

1. **`claude/scripts/reconcile-project-board.py`** — reads `.claude/hook-config.json` (same config
   `post-tool-use.py` uses, so field IDs never drift). Lists open dev-env issues, lists project #3
   items, computes the set difference (orphans), adds each orphan to the board, then **surfaces** the
   orphans (and any pre-existing open board items) still missing Impact and/or Why — emitting the
   exact `gh project item-edit` commands. It **never guesses Impact/Why** and **never** mutates
   single-select *options* (avoids the option-mutation hazard). `--dry-run` reports without adding.
2. **`claude/routines/reconcile-project-board/`** — nightly routine that runs the script and
   push-notifies when issues need Impact/Why.

This is the remediation for the gap ADR-053 documented (and complements the upstream re-check tracked
in #424): the hook is the happy path; this is the backstop for the sessions where the hook can't fire.

## Out of scope (follow-up)

Generalizing the engine to scan every project-configured repo under `C:/Users/brown/Git` (a
[...]

PR #465, opened 2026-07-01
Generalize reconcile-project-board.py to reconcile every configured repo (--scan-dir)

### LINKED ISSUE #462: Generalize reconcile-project-board.py to reconcile every configured repo (--scan-dir)
## Problem

`claude/scripts/reconcile-project-board.py` only reconciles the single repo it's invoked
from (default: the canonical checkout of whichever repo the script lives in). Every repo
that adopts the `required_fields` hook-config pattern (ADR-023) — or more generally, any
repo with a valid `.claude/hook-config.json` — has the same orphaned-issue gap this tool
was built to catch (ADR-068), but today each would need its own separate invocation, and
the nightly `reconcile-project-board` routine only ever reconciles dev-env's own board.

## Goal

Add a `--scan-dir PATH` mode that discovers every git repo directly under `PATH` with a
`.claude/hook-config.json` present, and reconciles each one — mirroring the existing
`--repo-path` / `--scan-dir` split already shipped in `claude/scripts/prune-merged-worktrees.py`
and `claude/scripts/reclaim-worktree-disk.py` (both routines already call
`--scan-dir C:/Users/brown/Git`). Update the `reconcile-project-board` routine to use the
new mode so the nightly sweep covers every project-configured repo, not just dev-env.

Follow-up to #447 (deferred scope named explicitly in PR #451 / ADR-068's "Alternatives
Considered -> Multi-repo scan now").

## Acceptance

- `reconcile-project-board.py --scan-dir PATH [--dry-run]` reconciles every discovered repo
  that has a `.claude/hook-config.json` (with `repo`/`project_number`/`project_owner` set),
  skips repos without one, isolates a single repo's `gh` failure from the rest of the scan,
  and reports an aggregate result line.
- `claude/routines/reconcile-project-board/SKILL.md` invokes `--scan-dir C:/Users/brown/Git`.
- New pure-helper test coverage in `test_reconcile_project_board.py`.

brownm09:
## Summary

- Adds `--scan-dir PATH` to `reconcile-project-board.py`: discovers every git repo directly
  under `PATH` with a usable `.claude/hook-config.json` (`repo`/`project_number`/
  `project_owner` set) and reconciles each one's project board — mirroring the
  `--repo-path`/`--scan-dir` split already shipped in `prune-merged-worktrees.py` and
  `reclaim-worktree-disk.py`.
- Single-repo mode's behavior and every stderr message it can print are unchanged
  byte-for-byte (verified by diff-reading the refactor, not just by inspection).
- Scan-dir mode silently skips repos with no config (board tracking is opt-in — most repos
  won't have one), isolates one repo's non-scope `gh` failure from the rest of the scan, and
  fails the whole scan fast on a `gh` `project`-scope error (token-level, would repeat
  identically for every remaining repo — including one that only surfaces at add time, see
  Review findings disposition below).
- Updates the `reconcile-project-board` routine to scan `C:/Users/brown/Git` nightly instead
  of reconciling dev-env's board alone.
- New ADR-070 documents the scope decision (any repo with hook-config.json, confirmed with
  the user during scoping — see the ADR) and the fatal-vs-skip design.

Closes #462 (follow-up to #447, deferred scope named in PR #451 / ADR-068).

## Testing

1. Syntax check (dev-env `## Testing` #1): `py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py` — clean.
2. `py -3 claude/scripts/tests/test_reconcile_project_board.py` — all pass: the original 15 pure-helper tests, 3 for `render_scan_summary`, plus 8 added in review response (3 for the extracted `_added_and_failed`, 5 for `_validated_config` — ok / no-config / missing-required-keys / colliding-required-fields / non-dict-JSON, the last being the regression test for the scan-crash fix below).
3. Manual regression: `py -3 claude/scripts/reconcile-project-board.py --dry-run` (single-repo,
   default canonicalized root) — output identical to pre-PR behavior.
4. Manual smoke test: `py -3 claude/scripts/reconcile-project-board.py --dry-run --scan-dir C:/Users/brown/Git`
   — found 29 repos, skipped 27 (no config), reconciled 2 (dev-env, lifting-logbook), correct
   per-repo `Repo:` headers, correct aggregate `RESULT:` line summing both repos' counts.
5. Post-review smoke tests against throwaway directories (not just unit tests): (a) a scan
   directory containing a repo with a bare `[]` hook-config.json no longer crashes the sweep —
   it's skipped and the scan continues to the remaining repos; (b) a nonexistent `--scan-dir`
   path now prints a distinct failure message, still emits a `RESULT:` line, and exits 1; (c) a
   genuinely empty (but valid) scan directory prints its own distinct message, still emits a
   `RESULT:` line, and exits 0 — (b) and (c) are now distinguishable, where before both looked
   identical.

**Tests: 26 passed, 0 skipped, 0 failed.**

## Coverage gate (ADR-022)

Pure logic (`render_scan_summary`, `_added_and_failed`, `_validated_config`) has direct unit
tests — `_validated_config` was added in review response specifically because it needs no
gh/filesystem mocking (just a real tmp-dir hook-config.json) and its non-dict-JSON case is the
regression test for a Blocking finding (see below). The remaining orchestration functions
(`find_git_repos`, `_reconcile_repo`, `_main_single_repo`, `_main_scan_dir`) stay integration-
tested via `--dry-run` rather than unit-tested, consistent with this file's existing convention
of not mocking the gh/filesystem boundary — matching `find_git_repos`'s siblings in
`prune-merged-worktrees.py`/`reclaim-worktree-disk.py`.

## Suppression / test-integrity grep

Both pre-PR greps ran clean against added lines, for the original diff and again after the
review-response commit. The test-integrity grep matched one **unchanged context line**
(`sys.exit(main())`, via the JS-oriented `xit\(` pattern substring-matching Python's `exit(` —
a false positive, not a real skip marker); confirmed with `git diff --cached origin/main -- . |
grep '^+' | grep -E '...'` returning zero matches. No suppressions or test-integrity violations
were added.

## ADR-warrant

New [ADR-070](docs/adr/070-reconcile-project-board-scan-dir.md) — the change alters the
nightly routine's blast radius (every configured repo instead of dev-env alone) and makes a
non-obvious design call (any-repo-with-config scope, fail-fast-on-scope-error vs.
per-repo-isolation-on-other-errors) that `git log` won't recover. `docs/adr/INDEX.md` and
ADR-068's Consequences section are updated in this PR.

## Doc reconciliation

`docs/REFERENCE.md` (routine description + Utilities entry), `README.md` (routine table), and
`claude/routines/reconcile-project-board/SKILL.md` (which is the live scheduled-task copy too,
via junction — no separate re-registration needed since only step content changed, not the
`schedule:`/`name:` frontmatter) are all updated in this PR.

`/review` ([comment](https://github.com/brownm09/dev-env/pull/465#issuecomment-4855872432)) ran
as two parallel subagents (correctness/security; reliability/performance/maintainability) given
the 595-line diff. All findings fixed in [e4ef398](https://github.com/brownm09/dev-env/commit/e4ef398) — none filed, none out of scope:

- **[correctness, blocking]** Non-dict `hook-config.json` crashed `_validated_config` with an
  unhandled `AttributeError`, aborting the entire `--scan-dir` sweep on one malformed file —
  **fixed**: `_validated_config` now returns `bad-config` for a non-dict payload; regression
  test added.
- **[correctness, blocking]** An add-time `project`-scope failure (list-scope ok, add-scope
  missing) wasn't treated as the documented scan-aborting `scope-error` — only fetch-time scope
  failures were — **fixed**: `_reconcile_repo` now surfaces `add_scope_error`, and `--scan-dir`
  aborts on it the same way it aborts on a fetch-time scope-error. Single-repo mode unchanged.
- **[reliability, non-blocking]** A nonexistent/unreadable `--scan-dir` looked identical to a
  successful empty scan (no `RESULT:` line either way) — **fixed**: `find_git_repos` now
  returns `None` (scan failed) vs `[]` (scanned, empty); `_main_scan_dir` always prints the
  aggregate `RESULT:` line and exits 1 on the failure case.
- **[maintainability, non-blocking]** `added_count`/`add_failed` computed identically in two
  places (`render_report`, `_reconcile_repo`) — **fixed**: extracted `_added_and_failed()`,
  used by both.
- **[maintainability, non-blocking]** `os.scandir` handle in `find_git_repos` never explicitly
  closed — **fixed**: wrapped in a `with` block.
- **[maintainability, non-blocking]** `config['repo']` subscript inconsistent with
  `config.get("repo", "")` used elsewhere in the file — **fixed**.
- **[maintainability, non-blocking]** New orchestration functions had zero unit coverage,
  notably `_validated_config` which needs no mocking — **fixed**: added 8 unit tests (see
  Testing above).
- **[question]** Whether excluding symlinked/junctioned repo dirs from `--scan-dir` discovery
  was intentional — **resolved by documentation**: confirmed no repo under `C:/Users/brown/Git`
  is currently a junction (latent, not active), and added a docstring note in `find_git_repos`
  making the exclusion explicit rather than incidental.


brownm09:
## Review: Generalize reconcile-project-board.py to reconcile every configured repo (--scan-dir)

### Summary
Solid, well-tested refactor (18/18 tests pass, byte-identical single-repo behavior verified, real `--dry-run --scan-dir` smoke test against 29 repos) — but two blocking correctness issues undercut the PR's own core promise of per-repo failure isolation. The most serious: a malformed (valid-JSON-but-non-dict) `hook-config.json` in any one repo under the scan directory crashes the *entire* nightly sweep with an unhandled `AttributeError`, silently skipping every repo after the bad one — exactly the failure mode `--scan-dir` was built to avoid. Needs a small fix before merge; everything else here is non-blocking polish.

---

### Blocking Findings

**[correctness]** `claude/scripts/reconcile-project-board.py:545` (`_validated_config`)
`load_config()` only maps a missing file or a `JSONDecodeError` to `None` — a `hook-config.json` containing valid-but-non-object JSON (a bare array, string, or number) parses cleanly and is returned as-is. `_validated_config` immediately calls `config.get("repo", "")` on that value with no type check, which raises `AttributeError` for anything that is not a dict (confirmed: `[].get(...)` raises `AttributeError: 'list' object has no attribute 'get'`).

In single-repo mode this was already true pre-PR (unchanged behavior — a crash there just means "this one invocation fails," same as any other operational failure). `--scan-dir` mode changes the blast radius: `_validated_config` is called from inside `_main_scan_dir`'s per-repo loop with no try/except around it, so this exception propagates up and aborts the whole scan — every repo after the malformed one silently never gets reconciled that night, with no counted `repos_failed`, no aggregate `RESULT:` line, nothing. That is precisely the single-point-of-failure this PR's `repos_failed`/`scope-error` isolation design exists to prevent.

**Fix:** In `_validated_config`, treat a non-dict `load_config()` result the same as `bad-config` (or `no-config`) — e.g. `if not isinstance(config, dict): return None, "bad-config", "<message>"` — before calling `.get()` on it. Add a unit test for `_validated_config` with a non-object JSON payload (e.g. `[]`) alongside the fix, since `_validated_config` is pure and cheaply testable and this exact case would have caught the bug.

**[correctness]** `claude/scripts/reconcile-project-board.py:505-517` (`_reconcile_repo` add-loop) vs. `claude/routines/reconcile-project-board/SKILL.md` / PR description
The PR and routine both document a hard guarantee: a missing `project` gh scope is a token-level failure that "aborts the entire scan immediately... every remaining repo would fail identically." That guarantee only holds when the scope failure surfaces from the fetch path (`fetch_board_items`, line ~495) — `looks_like_scope_error` there returns `status: "scope-error"`, which `_main_scan_dir` correctly treats as fatal. But gh's read (list) and write (add) project scopes are distinct (as `add_to_project`'s own docstring notes) — a token with read-but-not-write project scope sails through the fetch, then fails on every `add_to_project` call in the loop at line 510. That path only sets a local `scope_warned` flag, prints a per-repo stderr warning, and returns `status: "ok"` with `add_failed = len(orphans)` — indistinguishable from a transient add failure. In `--scan-dir` mode, this means the exact same scope failure repeats silently for every remaining repo (the very thing the fetch-path check exists to prevent), miscounted as ordinary `add_failed` instead of triggering the fast-abort plus `gh auth refresh -s project` hint the routine keys on.

**Fix:** Have the add-loop's scope detection also return `status: "scope-error"` (instead of continuing to accumulate `add_failed`) so scan-dir aborts on the first affected repo, matching the documented invariant — or, if silent per-repo continuation is actually acceptable for the add-scope case, correct the SKILL.md/PR wording so "aborts the entire scan immediately" is not overstated.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/reconcile-project-board.py:393-410` (`find_git_repos`) / `:593-597` (`_main_scan_dir`)
When `scan_dir` doesn't exist or isn't readable, `find_git_repos` prints a stderr `WARNING: cannot scan ...` and returns `[]`. `_main_scan_dir` can't distinguish that from a directory that scanned fine and simply contains zero repos — both print `"no git repos found under {scan_dir}"` to stdout and `return 0`, with no `RESULT:` line at all. For an unattended nightly job, a mistyped or unmounted `--scan-dir` path (currently a hardcoded string in the routine) would look identical to "ran fine, nothing to do" rather than surfacing as the misconfiguration it is.
**Suggestion:** Have `find_git_repos` signal the scan-failure case distinctly (e.g. raise, or return `None` vs `[]`), and have `_main_scan_dir` still emit a `RESULT:` line (or a clearly distinct message) on that path so the routine's "read the final RESULT: line" instruction has something to read either way.

**[maintainability]** `claude/scripts/reconcile-project-board.py:521-522` vs `render_report` (`:265-266` in the pre-PR file)
The `added_count`/`add_failed` computation (`sum(1 for o in orphans if o.get("item_id")) if not dry_run else 0`, etc.) is duplicated verbatim between `render_report` (to build its own RESULT line) and `_reconcile_repo` (to populate its return dict for the scan-dir aggregate). If either side's definition of "added" ever changes, the per-repo report and the scan-wide aggregate can silently disagree.
**Suggestion:** Extract a single helper (e.g. `_added_and_failed(orphans, dry_run) -> tuple[int, int]`) used by both, or have `render_report` return the counts it already computes.

**[maintainability]** `claude/scripts/reconcile-project-board.py:399` (`find_git_repos`)
`os.scandir(scan_dir)` is passed directly into `sorted(...)`; the scandir iterator/handle is never explicitly closed (no `with` block). Mirrors an identical pattern already shipped in `prune-merged-worktrees.py`, so this is not a new mistake, but since scan-dir now runs this nightly against the same directory and the PR is already touching this exact shape, it is a low-cost opportunity to close it properly.
**Suggestion:** `with os.scandir(scan_dir) as it: entries = sorted(it, key=...)`.

**[maintainability]** `claude/scripts/reconcile-project-board.py:615`
`print(f"\nRepo: {config['repo']} ({repo_root})")` uses direct subscript access, while every other read of the same key in this file (`:485`, `:545`) uses `config.get("repo", "")`. Safe today only because `_validated_config` already guarantees `repo` is present by the time this line runs — but that invariant lives two functions away, so the direct index is inconsistent with the defensive style used a few lines earlier and would raise `KeyError` if the validation contract ever loosened.
**Suggestion:** Use `config.get("repo", "")` here for consistency with the rest of the file.

**[maintainability]** `claude/scripts/tests/test_reconcile_project_board.py`
The new orchestration functions carrying this PR's entire behavioral change — `find_git_repos`, `_validated_config`, `_reconcile_repo`, `_main_scan_dir`, `_main_single_repo` — have zero unit coverage; only the new pure `render_scan_summary` (3 tests) and the pre-existing helpers are exercised. The PR body's Coverage-gate justification (matching the repo's no-subprocess-mock convention) is valid for the gh/filesystem-boundary functions, but `_validated_config` specifically is a pure function over `load_config`'s return value — no mocking needed, just synthetic dict/non-dict inputs — and a table-driven test over its three statuses (`ok`/`no-config`/`bad-config`) would have caught the first Blocking finding directly.
**Suggestion:** Add direct unit tests for `_validated_config` (valid config, missing-field config, colliding-required-fields config, and — per the fix above — a non-dict config).

---

### Questions for Author

**Question:** Is excluding symlinked/junctioned repo directories from `--scan-dir` discovery (`find_git_repos`'s `entry.is_dir(follow_symlinks=False)`, line 401) an intentional generalization, or specifically meant only to exclude worktrees?
**Context:** Worktrees are already excluded by a different mechanism — a worktree's `.git` is a file, not a directory, so the subsequent `os.path.isdir(.../".git")` check excludes them regardless of `follow_symlinks`. The `follow_symlinks=False` on the outer `entry.is_dir()` check additionally skips any top-level entry that is itself a symlink/junction to a directory — a distinct case from worktree exclusion. Verified no repo directly under `C:/Users/brown/Git` is currently a junction, so this is latent, not an active bug.
**Tradeoffs:** (a) Keep as-is — safest default (only reconcile primary, non-aliased checkouts), but a future junctioned repo root would be silently invisible to reconciliation with no warning printed anywhere. (b) Relax to `follow_symlinks=True` on the outer check — would surface junctioned repo dirs too, but risks double-reconciling if a junction and its target both happen to sit under the same scan dir. Given dev-env's own heavy use of directory junctions elsewhere in `~/.claude/`, worth a one-line comment either way so the exclusion reads as deliberate rather than incidental.

---

### Findings Disposition
Per ADR-028, each finding above will be fixed in this PR or filed and linked before merge — recorded in a "Review findings disposition" section in the PR body prior to `gh pr merge`.

---

*Reviewed by `claude-opus-4-8` (parallel correctness/security + reliability/performance/maintainability subagents) via Claude Code*

PR #468, opened 2026-07-01
feat(hooks): block git-mutating Bash commands in a canonical checkout

### LINKED ISSUE #453: Two Claude Code sessions sharing one non-worktree checkout: HEAD thrashing swallowed a commit into the wrong PR
## Symptom

A session working in `career-playbook` (not a worktree — the main checkout at `C:/Users/brown/Git/career-playbook`) created a branch, edited files, and committed. Immediately after, `git branch --show-current` and `git log --oneline -1` returned a **different branch and different HEAD** than the one just created — across consecutive tool calls within the same turn, with no intervening user action.

## Root cause (reconstructed via `git reflog`)

A second, concurrent Claude Code session was also operating directly in the same main checkout (not a worktree) on unrelated work (a cover-letter execution-time-audit feature, issues career-playbook#579/#580/#581). The reflog showed:

1. Session A: `git checkout -b chore/port-durable-rules-575-576 origin/main`, edited 4 files, committed (`747be13`).
2. Session B (concurrently, same checkout): `git checkout feat/exec-structural-audit` — this silently moved HEAD *out from under session A*, so session A's commit `747be13` landed as a child of `feat/exec-structural-audit` instead of staying on `chore/port-durable-rules-575-576`.
3. Session B committed its own work on top (`1d17c47`, `4c69d30`), rebased onto `origin/main`, and eventually squash-merged via [career-playbook#587](https://github.com/brownm09/career-playbook/pull/587) — carrying session A's 4-file docs change in as an unattributed rider. The PR title/description referenced only session B's issues (#579–581); session A's issues (career-playbook#575, career-playbook#576) went unmentioned in the merging PR.

No commit was lost — `git cat-file -t 747be13` still resolved after the fact, and the content was recoverable — but only because session A stopped to diagnose via reflog instead of trusting the (thrashing) working-tree state.

## Impact

- **Attribution scrambled.** A rules/docs fix (career-playbook#575, #576) merged to `main` under a PR about unrelated feature work, with no `Closes #575` / `Closes #576` anywhere in the merging PR. Both issues had to be closed and given resolution comments manually, after the fact, pointing at the actual carrying commit.
- **Near-miss on a destructive action.** Session A's own branch (`chore/port-durable-rules-575-576`) was now based on a stale `origin/main` (missing session B's since-merged work). Had session A opened a PR from that branch without first diffing against the new `origin/main`, merging it would have **reverted session B's legitimate work**. The only thing that prevented this was session A pausing to check `git diff origin/main <branch> -- <touched files>` before opening the PR and discovering the diff was empty (the changes were already upstream).
- **Wasted diagnostic effort.** Recovering the true sequence of events required reading `git reflog`, cross-referencing commit hashes against branch-contains queries, and diffing against `origin/main` — meaningfully more expensive than the original task (a 2-file docs port).

## Why this isn't covered by existing guidance

[ADR-066](https://github.com/brownm09/dev-env/blob/main/docs/adr/066-worktree-session-safety-rules.md) (worktree session safety) addresses the case of **one session in a worktree, one in main** — e.g. "Worktree Bash commands must not `cd` to the main repo" and the split-brain tell "`git status` clean but `Read` shows uncommitted edits → two different checkouts." This incident is a different shape: **two sessions both directly in the main checkout, no worktree involved at all.** Nothing in the current rules says two sessions should never share one non-worktree checkout concurrently, and nothing detects or prevents it.

## Motivating incident

career-playbook, 2026-07-01. Session transcript: porting durable rules career-playbook#575 and career-playbook#576 into `CLAUDE.md` / `context/prose_style.md`. Commit `747be13` on the intended branch `chore/port-durable-rules-575-576`; landed on `main` via `fc1fa91` (career-playbook#587) instead. Both issues closed with a resolution comment tracing the actual path: [career-playbook#575](https://github.com/brownm09/career-playbook/issues/575#issuecomment-4848539782), [career-playbook#576](https://github.com/brownm09/career-playbook/issues/576#issuecomment-4848539142).

## Open questions for discussion (not prescribing a fix yet)

- Should a `SessionStart` (or `PreToolUse`) hook detect that another Claude Code session already has an active branch/uncommitted state in the *same* non-worktree checkout, and warn or force worktree isolation before proceeding?
- Is the right invariant "never run two sessions directly in a project's main checkout concurrently — the second one always gets a worktree," analogous to how worktrees already isolate session-vs-main? That would extend ADR-066's model rather than replace it.
- At minimum: should the recovery pattern used here (reflog reconstruction → `git branch --contains <sha>` → diff-against-`origin/main` before opening/trusting a PR from a possibly-stale branch) be written up as a runbook next to the existing worktree-deregistration-recovery runbook in `docs/REFERENCE.md`, so a future session facing HEAD thrashing doesn't have to re-derive the diagnostic sequence from scratch?

Surfaced during a routine memory→instructions porting task; no data was lost, but the near-miss (destructive PR from a stale branch) and the attribution loss are worth closing off.

brownm09:
## Summary

Closes #453.

Two Claude Code sessions can both work directly in the same repo's **canonical (non-worktree) checkout** at once — nothing in the harness prevents it. When they do, one session's `git checkout` (branch switch, not necessarily `-b`) silently thrashes HEAD and the working tree out from under the other, mid-turn. #453 documents two occurrences (both 2026-07-01, both career-playbook):

1. **Attribution scrambled** — Session A's commit landed as a rider on Session B's unrelated PR, with no `Closes #N` for Session A's issues; both had to be closed manually after the fact.
2. **Near-miss + a second failure dimension** — a stale-branch PR could have reverted a concurrent session's already-merged work; separately, two sessions sharing one checkout exhausted the GitHub GraphQL API's 5,000/hr rate-limit bucket mid-session, disabling `gh pr merge`/`gh pr comment`/`gh pr view --json` for both.

Neither existing worktree ADR covers this: [ADR-066](https://github.com/brownm09/dev-env/blob/main/docs/adr/066-worktree-session-safety-rules.md) addresses a session already in a worktree `cd`-ing to main (a different shape); [ADR-024](https://github.com/brownm09/dev-env/blob/main/docs/adr/024-worktree-path-guard-hook.md)'s `pre-tool-use-worktree-path-check.py` is a **complete no-op** here, since it only fires on `Write`/`Edit`/`NotebookEdit` and only when `cwd` matches the worktree pattern — the colliding sessions in #453 are never in a worktree at all.

## Changes

- **New hook** `claude/scripts/pre-tool-use-canonical-mutate-guard.py` — a `PreToolUse(Bash)` hard-block, the inverse complement of ADR-024's hook. Blocks git-mutating commands (`checkout`/`-b`, `switch`, `commit`, `merge`, `rebase`, `reset`, `cherry-pick`, `revert`, `stash pop`/`apply`, `branch -d`/`-D`, bare `pull`) whenever `cwd` resolves to a canonical (non-worktree) root. Read-only commands (`status`, `log`, `diff`, `show`, `fetch`, `branch --show-current`, `rev-parse`, `ls-tree`, `blame`, `remote -v`, plain `branch`, `stash list`/`show`, `checkout -- <path>`, `pull --ff-only`) and anything from inside a worktree are unaffected. `ALLOW_CANONICAL_MUTATE=1` is the deliberate, visible override — same pattern as career-playbook's `block-artifact-merge.py`/`block-letter-violations.py` and ADR-024's own recovery-command style.
- **New test file** `claude/scripts/tests/test_canonical_mutate_guard.py` — pure-function coverage of the full mutating-verb matrix + read-only allowlist + segment-split/anchor + two-tier redirect-scope behavior, plus subprocess end-to-end coverage of block/allow/override/fail-open paths against a real throwaway git repo.
- **`claude/settings.json`** — wired as the 5th entry in the existing `Bash` matcher array (same commit as the hook, per this repo's atomic-commit invariant).
- **`claude/CLAUDE.md`** — new Git Workflow bullet stating the invariant, right after "Worktree Bash commands must not `cd` to the main repo."
- **`docs/REFERENCE.md`** — new recovery runbook ("Concurrent-session HEAD thrashing in a canonical (non-worktree) checkout") under Git Workflow Runbooks, right after "Worktree deregistration recovery" — the exact reflog-reconstruction / remote-only-recovery sequence both #453 incidents actually used.
- **`docs/adr/071-canonical-checkout-mutate-guard-hook.md`** (+ INDEX row) — new ADR, not an ADR-066 addendum (#453 was discovered independently and this ships a hook + tests + settings wiring, the same shape ADR-024 itself flagged as ADR-warranting).

## Judgment calls (documented in the ADR)

- `cd <path>` anywhere in the command takes the **whole command** out of scope (a `cd` persists across `&&`-chained segments); `git -C <path>`/`--git-dir=<path>` takes only **its own segment** out of scope (it doesn't change the shell's directory). These are genuinely different redirect shapes — the first implementation collapsed them into one per-segment check and a test caught the bug (a `cd X && git checkout -b foo` was incorrectly flagged, since the redirect and the mutation land in different segments).
- Bare `git checkout <path>` without `--` is conservatively classified as mutating (could be a branch name).
- `git pull` is blocked unless the same segment has `--ff-only`.
- Both are v1 deferrals, not solved: a command that `cd`s/`-C`s into the canonical root from elsewhere is not caught — same class of scope-limiting judgment ADR-024 made for its own Bash coverage.

All 12 findings from the [code review](https://github.com/brownm09/dev-env/pull/468#issuecomment-4856927874) addressed in commit `044c02f`, plus one filed as a separate out-of-scope issue:

1. **Fixed in `044c02f`** — block reason now written to `sys.stderr` (was `print(...)` to stdout, discarded by Claude Code on exit code 2). E2E test updated to assert against `proc.stderr` instead of `proc.stdout`.
2. **Fixed in `044c02f`** — override check replaced with `_has_override()`, an anchored leading-prefix match on the command or on any `&&`/`||`/`;`/`|`-split segment, instead of a bare substring test. `ALLOW_CANONICAL_MUTATE=1` merely mentioned inside a commit message no longer bypasses the block. New tests: genuine-prefix-bypasses and mention-only-does-not-bypass.
3. **Fixed in `044c02f`** — removed the undocumented `canonical-mutate-approved` second override token; only `ALLOW_CANONICAL_MUTATE=1` remains (`OVERRIDE_TOKENS` tuple → `OVERRIDE_TOKEN` string), matching what's documented in the docstring/CLAUDE.md/ADR everywhere else.
4. **Fixed in `044c02f`** — stash subcommand detection now scans all tokens after the verb for `pop`/`apply` instead of trusting a fixed `tokens[1]` position. This needed a deeper fix than the finding's framing: git-level flags before the subcommand (`git -c gc.auto=0 stash pop`, `git --no-optional-locks stash apply`) were pushing the actual verb off `tokens[0]` entirely, not just pushing `pop`/`apply` off `tokens[1]` — a new `_skip_git_level_flags()` helper strips leading git-level options first. All three of the finding's examples (`-c gc.auto=0 stash pop`, `--no-optional-locks stash apply`, `stash --quiet pop`) now correctly classify as mutating; verified with a 4-case test.
5. **Fixed in `044c02f`** — `cd` detection in `classify()` now runs against the env-stripped segment (reusing `_strip_leading_env`), so `FOO=1 cd /tmp && git checkout -b x` is correctly recognized as a cd-redirect (whole command out of scope) instead of falsely blocking the checkout. New test with two env-prefixed cd cases.
6. **Fixed in `044c02f`** — `main()` now checks `isinstance(data, dict)` immediately after `json.loads`, guarding against valid-but-non-dict JSON (`[]`, `"x"`, `123`, `null`) that previously raised an uncaught `AttributeError` on `data.get(...)`, contradicting the hook's own fail-open contract. New e2e test covers all four non-dict payload shapes.
7. **Fixed in `044c02f`** — ADR-071's Observability section rewritten to describe the stderr mechanism (was inaccurate before this fix too, describing stdout).
8. **Fixed in `044c02f`** — `classify()` now splits the command once into a local `segments` variable, reused by both the cd-scan loop and the classification loop.
9. **Fixed in `044c02f`** — promoted the two inline regexes (git-invocation match, cd detection) to named module-level compiled constants (`_GIT_INVOCATION_RE`, `_CD_RE`), alongside the file's other `_*_RE` patterns.
10. **Fixed in `044c02f`** — added a cross-reference comment at the top of the verb-branch chain in `is_mutating_segment()` naming the four prose locations (module docstring, `claude/CLAUDE.md`, ADR-071, `docs/REFERENCE.md`) that must stay in sync with the mutating-verb list. Did not attempt the optional single-source refactor — the per-verb conditions are non-uniform (some unconditional, some flag-gated with different flag sets), so a single frozenset would lose information rather than simplify.
11. **Fixed in `044c02f`** — added a unit test for `_resolve_git_toplevel`'s fail-open guarantee, monkeypatching `cmg.subprocess.run` with stubs that raise `subprocess.TimeoutExpired` and `OSError` respectively, asserting both paths return `None`.
12. **Fixed in `044c02f`** — `_init_throwaway_repo()`'s `git init` call now passes `-c init.templateDir= -c core.hooksPath=` to neutralize any global template directory / hooks path.
- **Filed [#469](https://github.com/brownm09/dev-env/issues/469)** — the reviewer's out-of-scope note that `claude/scripts/pre-tool-use-worktree-path-check.py` (ADR-024's hook, already merged/live, not touched by this PR) likely has the same stdout-discard-on-exit-2 bug. Not fixed here per the explicit scope boundary; issue describes the suspected bug, points at this PR/ADR-071 as precedent, and gives verification + fix steps for a follow-up PR.

Two manual sanity checks confirmed both bugs are actually gone (not just test-covered):
- `git commit -m "ALLOW_CANONICAL_MUTATE=1 mentioned only"` run against a throwaway canonical-root repo is now **BLOCKED** (exit 2), not bypassed.
- The block message now appears on **stderr**, not stdout, when the hook is invoked directly with mutating-command stdin.

## Testing

Both required suites, run from the worktree:

```
```
```
PASS: mutating verbs classified as mutating
      16 mutating-verb cases correctly classified
PASS: read-only commands classified as safe
      24 read-only cases correctly classified as safe
PASS: classify() finds first mutating segment
      classify() finds the mutating segment among safe ones
PASS: classify() returns None when all safe
      classify() returns None when every segment is read-only
PASS: heredoc mention does not trigger
      heredoc/prose mention of a mutating verb does not trigger (career-playbook #442 lesson)
PASS: env-prefixed mutating command still classified
      leading env-var assignment does not prevent verb classification
PASS: cd redirect takes whole command out of scope
      2 cd-redirect commands correctly out of scope in full (not just the cd segment)
PASS: -C/--git-dir redirect skips only its own segment
      git -C/--git-dir skips only its own segment; other segments still classified
PASS: flag-prefixed stash pop/apply classified as mutating
      4 flag-prefixed stash pop/apply cases correctly classified as mutating
PASS: env-prefixed cd recognized as redirect
      2 env-prefixed cd commands correctly recognized as a redirect (out of scope)
PASS: override anchored prefix bypasses
      2 genuine leading-prefix override cases correctly bypass
PASS: override mention-only does not bypass
      3 mention-only (non-prefix) occurrences correctly do NOT bypass
PASS: _resolve_git_toplevel fails open on timeout/OSError
      _resolve_git_toplevel returns None on both TimeoutExpired and OSError (fail-open guarantee)
PASS: main() blocks mutating command from canonical root
      mutating command from a canonical (non-worktree) git repo blocked (exit 2), reason on stderr
PASS: main() allows read-only command from canonical root
      git status from a canonical root is allowed (exit 0)
PASS: main() allows pull --ff-only, blocks bare pull
      git pull --ff-only allowed, bare git pull blocked
PASS: main() allows any command from worktree cwd
      mutating command from a worktree-pattern cwd allowed (out of scope, exit 0)
PASS: main() override token bypasses block
      ALLOW_CANONICAL_MUTATE=1 override bypasses the block (exit 0)
PASS: main() fails open on non-git cwd
      non-git cwd fails open (exit 0) -- cannot resolve a canonical root
PASS: main() fails open on malformed JSON
      malformed JSON fails open (exit 0)
PASS: main() fails open on non-dict JSON
      non-dict JSON payloads ([], "x", 123, null) all fail open (exit 0)
PASS: main() fails open on missing cwd
      missing cwd fails open (exit 0)
PASS: main() fails open on empty cwd
      empty cwd fails open (exit 0)
PASS: main() no-ops on non-Bash tool
      non-Bash tool_name is a no-op (exit 0)
PASS: main() no-ops on empty stdin
      empty stdin is a no-op (exit 0)

Tests: 25 passed, 0 skipped, 0 failed
```

```
py -3 claude/scripts/tests/test_worktree_path_check.py   # regression: ADR-024's hook unaffected
```
```
PASS: _worktree_is_live decision table
      6 liveness combinations classified correctly
PASS: missing .git short-circuits before git
      missing .git link blocks without spawning git
PASS: main() blocks Edit from orphaned worktree
      Edit from orphaned worktree cwd blocked (exit 2) with recovery recipe
PASS: main() no-op outside worktree
      non-worktree cwd is a no-op (exit 0)

Tests: 4 passed, 0 skipped, 0 failed
```

## Risk-dimension audit

1. **Testing** -- 25 tests (pure-function verb matrix + subprocess e2e, up from 19 after the review-finding fixes) + the ADR-024 regression suite re-run unmodified, both green.
2. **Observability** -- the block reason now correctly reaches the model via stderr (review finding #1); previously it was silently discarded on stdout, a real observability gap this PR closes.
3. **Security** -- the `ALLOW_CANONICAL_MUTATE=1` override is a deliberate, visible bypass typed inline on the command being run, matching the two existing precedent hooks (`block-artifact-merge.py`, `block-letter-violations.py`) and ADR-024's own recovery-command pattern -- not a hidden bypass. Review finding #2 closed a real bypass gap (substring match instead of anchored prefix).
4. **Resilience / failure modes** -- fails open on every ambiguous/unresolvable input (malformed JSON, non-dict JSON, missing/empty cwd, non-git cwd, git unavailable); can only ever block when it positively identifies a mutating segment in a positively-resolved canonical root. Revert is a straightforward hook removal from `settings.json` + file deletion -- no state to migrate.
5. **Performance** -- one `git rev-parse --show-toplevel` subprocess only for Bash calls that are (a) non-worktree cwd AND (b) contain a candidate-mutating segment -- the common case (non-git or read-only git commands) never reaches the subprocess.
6. **Data integrity & migrations** -- N/A, pure Python hook change with no state or schema involved.

## Note

Per the task instructions, stopping here -- no `/review`, no merge. This hook affects every future Bash tool call across every project on this machine once merged, so it gets a separate review pass before going live.


brownm09:
## Review: feat(hooks): block git-mutating Bash commands in a canonical checkout

### Summary

Needs revision before merge. The design and test coverage are solid (19/19 new tests, ADR-024 regression suite unaffected, careful judgment calls on the `cd`/`-C` redirect-scope distinction), but two review passes over the full 1141-line diff surfaced verified, high-severity correctness bugs that undermine the hook's core promise: **the block message is discarded on every actual block** (exit(2) writes to stdout, which Claude Code ignores on that exit code — only stderr is surfaced), and **the override check uses unanchored substring matching**, so `ALLOW_CANONICAL_MUTATE=1` (or an undocumented second token, `canonical-mutate-approved`) appearing anywhere in a command — a commit message, a heredoc, an echo — silently disables the block. Both were empirically verified, not inferred. This is exactly the kind of bug worth catching before a hook that gates every future Bash call on this machine goes live. All findings below are mechanical, same-file fixes; no design rework needed.

---

### Blocking Findings

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:291`
The block path does `print(json.dumps({"reason": reason}))` to stdout, then `sys.exit(2)`. Per the official Claude Code hooks docs, exit code 2 causes stdout (and any JSON in it) to be **ignored**; only stderr is fed back to the model as an error message. The command is still blocked, but the message naming the matched command, the canonical root, and the two remedies (isolate via worktree, or override) is thrown away — the model sees a bare failure with no explanation. Both precedent hooks this PR's ADR cites as its pattern (`block-artifact-merge.py`, `block-letter-violations.py`) correctly write to `sys.stderr`; this hook (and its ADR-024 sibling, which has the same latent bug) used the wrong stream. The e2e test asserts against `proc.stdout`, which is why it passed despite the discard.
**Fix:** Write `reason` to `sys.stderr` (keep `exit(2)`), matching the two working precedent hooks. Fix the test assertion to read `proc.stderr`. Optionally file a follow-up issue for the same latent bug in `pre-tool-use-worktree-path-check.py` (ADR-024) — out of scope for this PR but a real bug sharing the same root cause.

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:266`
The override check is `if any(tok in cmd for tok in OVERRIDE_TOKENS)` — a bare substring test against the *entire* command string. Verified: `git commit -m "ALLOW_CANONICAL_MUTATE=1 was mentioned"` registers the override and the mutating commit is silently allowed. The classifier itself is carefully anchored at segment-start specifically to avoid this class of bug (the career-playbook #442 heredoc-mention lesson, cited in this PR's own docstring) — the override check re-introduces the exact hazard the rest of the hook was designed to avoid.
**Fix:** Require the override token as a genuine leading prefix on a segment (reuse the existing segment-split/env-strip helpers), not a substring match anywhere in the command.

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:122`
`OVERRIDE_TOKENS` contains a second token, `canonical-mutate-approved`, that appears in no documentation (not the ADR, not the CLAUDE.md bullet, not the docstring, not the block message). Combined with the substring bug above, this is worse than the first token: it's a lowercase phrase with no `=1`/uppercase distinctiveness, far more likely to appear incidentally in a branch name, filename, or comment — silently disabling the guard with no reader of the docs aware it's even possible.
**Fix:** Either remove the second token, or document it explicitly and give it a more collision-resistant form. Apply the anchored-prefix fix above regardless of which token(s) remain.

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py` (stash classification — see PR diff, `is_mutating_segment`)
The stash subcommand is read at a fixed token position (`tokens[1]`) rather than scanned across the segment. `git -c gc.auto=0 stash pop`, `git --no-optional-locks stash apply`, and `git stash --quiet pop` all put something other than `pop`/`apply` at that position, so the function returns `False` and a genuinely mutating `stash pop`/`apply` is allowed through with no override needed — a silent classification miss, not a documented gap. Every other verb branch already handles flags via `in tokens[1:]`; stash is the one inconsistent case.
**Fix:** Scan `tokens[1:]` for `pop`/`apply` the same way the other verb branches scan for their flags, rather than checking only `tokens[1]`.

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:209`
The `cd` carve-out regex (`^\s*cd(?:\s|$)`) runs against the raw segment, but `is_mutating_segment` strips leading env-var assignments before classifying. Verified: `FOO=1 cd /tmp && git checkout -b x` is not recognized as a `cd`-redirect (the env prefix defeats the `cd` match), so the whole-command-out-of-scope short-circuit never fires and the `git checkout -b x` — which actually runs in `/tmp`, not the canonical root — gets falsely blocked. This is the "false positive that wedges legitimate work" failure class the plan explicitly required avoiding.
**Fix:** Run the `cd` detection against the env-stripped segment too, using the same leading-env-strip helper the verb classifier already has, so env-prefixed `cd` and env-prefixed git verbs are treated consistently.

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:243-248`
Only `json.JSONDecodeError` is caught after `json.loads`. Valid-but-non-object JSON (`[]`, `"x"`, `123`, `null`) — verified with `[]` — raises an uncaught `AttributeError` on `data.get(...)`, contradicting the hook's documented "fail open on anything unparseable" contract. Practical severity is bounded (Claude Code treats a non-zero, non-2 exit as non-blocking per the docs, so this doesn't wedge Bash globally — just a spurious per-call error notice), and the real PreToolUse payload is always an object, so this is a defensive-hardening gap rather than an observed live risk.
**Fix:** Add `if not isinstance(data, dict): sys.exit(0)` immediately after `json.loads`, matching the stated fail-open contract exactly.

**[documentation]** `docs/adr/071-canonical-checkout-mutate-guard-hook.md` — Observability bullet
States the block reason is "printed as hook JSON output surfaced directly to the model." Given the stdout/stderr finding above, as shipped this is false — nothing is currently surfaced to the model on block.
**Fix:** Self-resolves once the stdout→stderr fix lands; update the wording to "written to stderr" (or whichever mechanism is chosen) in the same commit as that fix.

---

### Non-Blocking Findings

**[performance]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py` — `classify()` calls `_SEGMENT_SPLIT.split(cmd)` twice (once in the `cd`-scan loop, once in the classification loop) for every Bash call that reaches it. Doubles a real per-call cost on the hot path for no behavioral difference.
**Suggestion:** Split once into a local variable and reuse it in both loops.

**[maintainability]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:153,209` — two hot-path regexes (git-invocation detection, `cd` detection) are inline string literals passed to `re.match` rather than module-level compiled constants, unlike `_WORKTREE_RE`/`_SEGMENT_SPLIT`/`_REDIRECT_RE`/`_LEADING_ENV`. Functionally harmless (Python's `re` cache absorbs it) but inconsistent with the file's own convention.
**Suggestion:** Promote both to named `re.compile(...)` module constants alongside the others.

**[maintainability]** The mutating-verb list is spelled out independently in the code (five separate branches) and in four prose locations (module docstring, CLAUDE.md bullet, ADR, REFERENCE.md) with no single canonical source and no test asserting they agree — the concrete mechanism by which the PR's own stated drift concern will actually happen.
**Suggestion:** At minimum, comment the code branches pointing at the prose locations that must stay in lockstep; better, drive the "simple verb" set from one named constant and add a test asserting docstring-listed verbs all appear in the classifier.

**[reliability]** `claude/scripts/tests/test_canonical_mutate_guard.py` — no test exercises `_resolve_git_toplevel`'s subprocess-timeout or unexpected-output branches (the `TimeoutExpired`/`OSError` except clauses), which are the actual reliability contract of the whole hook's fail-open guarantee. A future edit that broke timeout handling would turn a hung `git` into a hard failure on every Bash call, and nothing here would catch it.
**Suggestion:** Add a unit test that injects/monkeypatches a `TimeoutExpired`-raising runner and asserts `_resolve_git_toplevel` returns `None`, matching the sibling hook's dependency-injection testing style.

**[maintainability]** `claude/scripts/tests/test_canonical_mutate_guard.py` — the throwaway `git init` repos don't neutralize `init.templateDir`/`core.hooksPath` from a contributor's global git config (matches the sibling test's existing convention; not a regression, just noted for parity-of-awareness).
**Suggestion:** Optional hardening: `git -c init.templateDir= -c core.hooksPath= init` for full portability. Not blocking.

**[performance]** The hook adds a 5th per-Bash-call `pyw -3` process spawn to the existing `Bash` matcher chain. The subprocess-avoidance design goal (no `git rev-parse` on the common path) is correctly implemented, but the dominant per-call cost is Python interpreter startup, not the work this PR optimizes — pre-existing architecture, not introduced by this PR, noted for completeness only.
**Suggestion:** No action needed in this PR. If cumulative Bash-hook latency becomes a concern later, the fix is consolidating the 5 Bash `PreToolUse` hooks into one dispatcher — a separate, larger change.

---

### Findings Disposition

To be completed before merge — every blocking and non-blocking finding above will be either fixed in this PR (same commit as applicable) or filed and linked. None left as-is.

---

PR #473, opened 2026-07-01
fix(hooks): write worktree-path-guard block reason to stderr, not stdout

### LINKED ISSUE #469: pre-tool-use-worktree-path-check.py likely has the same stdout-discard-on-exit-2 bug
## Summary

While fixing review findings on PR #468 (`claude/scripts/pre-tool-use-canonical-mutate-guard.py`, ADR-071), a reviewer noted that `claude/scripts/pre-tool-use-worktree-path-check.py` (ADR-024's hook, already merged/live, not part of that PR's diff) likely has the *same* bug: it prints its block reason JSON to **stdout** before `sys.exit(2)`, but Claude Code discards stdout on a `PreToolUse` hook's exit code 2 — only stderr is surfaced to the model.

If confirmed, this means ADR-024's worktree-escape guard has been silently blocking writes without the model ever seeing *why* — the block itself still works (the write is prevented), but the recovery guidance in the reason text never reaches the model, defeating the hook's own documented purpose of surfacing a clear reason + recovery recipe.

## Where the pattern was first caught

PR #468 (dev-env#453 / ADR-071) — the canonical-mutate-guard hook had the identical bug (block reason on stdout, discarded on exit 2). Fixed there by switching to `sys.stderr.write(...)`, matching the working pattern already used in career-playbook's `.claude/hooks/block-artifact-merge.py` and `block-letter-violations.py`.

## To verify

1. Read `claude/scripts/pre-tool-use-worktree-path-check.py` — find the block path near the end of its `main()` (or equivalent) and confirm whether the reason JSON goes to `print(...)` (stdout) or `sys.stderr.write(...)`.
2. If it's stdout: this is a live, in-production instance of the same bug.
3. Check `claude/scripts/tests/test_worktree_path_check.py` for any test that currently asserts against `proc.stdout` for the block-reason JSON — that assertion would need to move to `proc.stderr` alongside the fix, same as the one in PR #468's test suite.

## Fix (if confirmed)

Same shape as PR #468's fix #1: switch `print(json.dumps({"reason": reason}))` to `sys.stderr.write(json.dumps({"reason": reason}) + "\n")`, keep `sys.exit(2)`, and update the corresponding test assertion from stdout to stderr.

## Scope note

This issue is intentionally **not** fixed in PR #468 — that PR's diff doesn't touch `pre-tool-use-worktree-path-check.py` at all, and the reviewer explicitly scoped this out as a separate PR to keep #468 focused on the canonical-mutate-guard hook.

Related: ADR-024 (`docs/adr/024-worktree-path-guard-hook.md`), ADR-071 (`docs/adr/071-canonical-checkout-mutate-guard-hook.md`), PR #468.

brownm09:
## Summary

`pre-tool-use-worktree-path-check.py` (ADR-024's `PreToolUse` hook) had the same bug just fixed in its sibling `pre-tool-use-canonical-mutate-guard.py` (#468, finding #1, 0b103b6): the block-reason JSON was printed to **stdout** at both `sys.exit(2)` call sites, but Claude Code discards a PreToolUse hook's stdout on exit code 2 — only stderr is surfaced to the model.

The hook still worked mechanically (the write was blocked), but the model never saw why or how to fix the path, defeating the hook's documented purpose (surfacing a clear reason + recovery recipe).

## Reproduction (pre-fix)

Invoked the hook directly via subprocess with both trigger scenarios:

- **Canonical-root escape** (the primary, most-documented scenario — a live worktree, absolute `file_path` targeting the canonical root): exit 2, reason JSON on stdout, stderr empty.
- **Orphaned-worktree liveness guard** (dev-env#328 addendum): exit 2, reason JSON on stdout, stderr empty.

Both confirmed the bug at both `print(json.dumps(...))` call sites.

## Fix

Switched both call sites to `sys.stderr.write(json.dumps({"reason": reason}) + "\n")`, keeping `sys.exit(2)` — identical pattern to #468's fix and to career-playbook's `block-artifact-merge.py` / `block-letter-violations.py`.

Re-ran both reproduction scenarios post-fix: exit 2, stdout empty, reason correctly on stderr for both.

**Post-review (`710f213`):** extracted both call sites into a shared `_block(reason)` helper, and added an explicit stderr note + a References entry to ADR-024. See disposition section below.

## Test changes

`test_worktree_path_check.py`:
- Fixed `test_main_blocks_edit_from_orphaned_worktree`'s assertion from `proc.stdout` to `proc.stderr` (plus an explicit empty-stdout check) — this is the test that should have caught the bug but was itself asserting against the wrong stream.
- **Added** `test_main_blocks_write_escaping_to_canonical_root` — the canonical-root-escape block path (the hook's primary scenario per ADR-024) had **zero** main()-level test coverage before this change; only the orphan path was exercised. The new test uses a bogus (non-gitdir-link) `.git` file to make the worktree register as "live" without needing a real git repo (confirmed empirically: `git rev-parse --show-toplevel` exits non-zero against it, which `_worktree_is_live` treats as fail-open live) — keeps the test hermetic, matching this file's existing convention.
- Test suite grew from 4 to 5 passing cases.

No ADR text needed correction for the original bug: unlike #468's finding #7 (ADR-071 explicitly described the mechanism as stdout), ADR-024 and `docs/REFERENCE.md` described the hook's blocking behavior without naming a specific stream, so neither had gone stale. (ADR-024 was still updated post-review to add a positive statement of the correct mechanism — see disposition below.)

## Testing

Ran the checks required for this change per dev-env's `CLAUDE.md` → `## Testing` (item 1, since a `claude/scripts/*.py` file changed) plus the test file directly (not individually numbered in that section, same as its sibling):

```
py -3 claude/scripts/tests/test_worktree_path_check.py
```

Tests: 5 passed, 0 skipped, 0 failed. Syntax check: OK. Re-ran both, identically green, after the post-review refactor commit (`710f213`).

Reviewed at https://github.com/brownm09/dev-env/pull/473#issuecomment-4859408888 (blocking=0, non_blocking=2). Both fixed in `710f213`:

1. **[maintainability]** Duplicated block-emission comment/logic across `main()`'s two `sys.exit(2)` call sites → fixed: extracted a shared `_block(reason)` helper, both sites now call it.
2. **[documentation]** ADR-024 didn't positively state the stderr mechanism (so a future reader had nothing to check against) → fixed: added an explicit line to ADR-024's Consequences section plus a References entry for #469.

Closes #469

### DIFF
--- a/claude/scripts/pre-tool-use-worktree-path-check.py
+++ b/claude/scripts/pre-tool-use-worktree-path-check.py
@@ -154,7 +154,11 @@ def main() -> None:
             f"  git worktree add --force {worktree_root} <branch>\n"
             f"(<branch> is typically claude/<worktree-name>; confirm with `git branch -a`.)"
         )
-        print(json.dumps({"reason": reason}))
+        # Claude Code discards stdout on a PreToolUse hook exit code 2 — only
+        # stderr is surfaced to the model. Write there, matching the working
+        # pattern in career-playbook's block-artifact-merge.py /
+        # block-letter-violations.py.
+        sys.stderr.write(json.dumps({"reason": reason}) + "\n")
         sys.exit(2)
 
     file_path = data.get("tool_input", {}).get(_PATH_FIELD[tool_name], "")
@@ -191,7 +195,11 @@ def main() -> None:
         f"\n"
         f"Re-issue with the corrected path."
     )
-    print(json.dumps({"reason": reason}))
+    # Claude Code discards stdout on a PreToolUse hook exit code 2 — only
+    # stderr is surfaced to the model. Write there, matching the working
+    # pattern in career-playbook's block-artifact-merge.py /
+    # block-letter-violations.py.
+    sys.stderr.write(json.dumps({"reason": reason}) + "\n")
     sys.exit(2)
 
 
--- a/claude/scripts/pre-tool-use-worktree-path-check.py
+++ b/claude/scripts/pre-tool-use-worktree-path-check.py
@@ -112,6 +112,19 @@ def _worktree_is_live(
     return _normalize(top) == _normalize(worktree_root)
 
 
+def _block(reason: str) -> None:
+    """Emit a blocking {"reason": ...} payload and exit 2.
+
+    Claude Code discards stdout on a PreToolUse hook exit code 2 — only
+    stderr is surfaced to the model. Write there, matching the working
+    pattern in career-playbook's block-artifact-merge.py /
+    block-letter-violations.py. Centralized so main()'s two independent
+    block sites can't drift out of sync on this again (dev-env#469).
+    """
+    sys.stderr.write(json.dumps({"reason": reason}) + "\n")
+    sys.exit(2)
+
+
 def main() -> None:
     raw = sys.stdin.read().strip()
     if not raw:
@@ -154,12 +167,7 @@ def main() -> None:
             f"  git worktree add --force {worktree_root} <branch>\n"
             f"(<branch> is typically claude/<worktree-name>; confirm with `git branch -a`.)"
         )
-        # Claude Code discards stdout on a PreToolUse hook exit code 2 — only
-        # stderr is surfaced to the model. Write there, matching the working
-        # pattern in career-playbook's block-artifact-merge.py /
-        # block-letter-violations.py.
-        sys.stderr.write(json.dumps({"reason": reason}) + "\n")
-        sys.exit(2)
+        _block(reason)
 
     file_path = data.get("tool_input", {}).get(_PATH_FIELD[tool_name], "")
     if not file_path or not os.path.isabs(file_path):
@@ -195,12 +203,7 @@ def main() -> None:
         f"\n"
         f"Re-issue with the corrected path."
     )
-    # Claude Code discards stdout on a PreToolUse hook exit code 2 — only
-    # stderr is surfaced to the model. Write there, matching the working
-    # pattern in career-playbook's block-artifact-merge.py /
-    # block-letter-violations.py.
-    sys.stderr.write(json.dumps({"reason": reason}) + "\n")
-    sys.exit(2)
+    _block(reason)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


brownm09:
## Review: fix(hooks): write worktree-path-guard block reason to stderr, not stdout

### Summary
Ready to merge. This is a small, well-verified bug fix (82 lines across 2 files) that mirrors an already-validated pattern (PR #468's finding #1, merged as 0b103b6 earlier today) at both of this hook's `sys.exit(2)` call sites. The fix is empirically confirmed in the PR body (pre- and post-fix stream captures for both trigger scenarios) and the previously-uncovered canonical-root-escape code path now has a regression test. No blocking findings.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/pre-tool-use-worktree-path-check.py:157,199` (post-diff line numbers)
Both `sys.exit(2)` call sites now carry an identical 4-line comment plus an identical `sys.stderr.write(json.dumps({"reason": reason}) + "\n")` line. This duplication pre-dates this PR (the two block sites have existed separately since ADR-024's orphan-liveness addendum), and this PR is right to fix both rather than just one — but leaving the emission mechanism duplicated means a third future block site (or a future change to *how* the reason is emitted) has two places to independently get right or wrong, which is the exact drift class that produced dev-env#469 in the first place: one hook (`canonical-mutate-guard.py`) got it right, this one didn't.
**Suggestion:** Extract a small helper, e.g. `_block(reason: str) -> None: sys.stderr.write(json.dumps({"reason": reason}) + "\n"); sys.exit(2)`, and call it from both sites. Optional — the current fix is correct as-is; this only reduces the surface for a future instance of this same bug class.

**[documentation]** `docs/adr/024-worktree-path-guard-hook.md`
Neither ADR-024 nor `docs/REFERENCE.md` names a specific output stream for this hook's block reason, so nothing in either doc went stale from this fix (unlike ADR-071, which explicitly and incorrectly described the mechanism as stdout until PR #468 corrected it). Since ADR-024 also never states the *correct* mechanism (stderr) either, a future reader has no positive confirmation to check against.
**Suggestion:** Add one line to ADR-024's "Consequences" section (or a short addendum, mirroring the pattern PR #468 used for ADR-071) stating explicitly that the block reason is written to stderr, not stdout, because Claude Code discards a PreToolUse hook's stdout on exit code 2. Optional, low cost, preempts the next person from having to re-derive this from the source.

---

### Findings Disposition
Both findings above are non-blocking suggestions (optional refactor / optional doc addition) rather than defects — disposition will be recorded as fixed-in-PR or deliberately deferred-with-rationale before merge, per the repo's all-findings merge gate.

---

PR #478, opened 2026-07-01
chore: extract shared _repo_scan module for find_git_repos()

### LINKED ISSUE #471: Extract shared _repo_scan module for find_git_repos() (ADR-070 follow-up)
## Problem

`claude/scripts/prune-merged-worktrees.py`, `claude/scripts/reclaim-worktree-disk.py`, and (as of #465 / ADR-070) `claude/scripts/reconcile-project-board.py` each carry their own near-identical copy of `find_git_repos()` — a helper that scans a directory for primary git repos (has a `.git` *directory*; worktrees, whose `.git` is a file, are excluded) to support each script's `--scan-dir` mode.

This was explicitly flagged as a deferred follow-up in [ADR-070](https://github.com/brownm09/dev-env/blob/main/docs/adr/070-reconcile-project-board-scan-dir.md) (see "Alternatives Considered" and "Consequences"):

> `find_git_repos` now exists in three near-identical copies across `claude/scripts/`. Tracked as a candidate for a shared-module extraction, not done here.

## Goal

Extract the common logic into a shared module `claude/scripts/_repo_scan.py`, following this repo's existing shared-module convention (`_worktree_liveness.py`, `_worktree_topology.py`, `_hookio.py`, `_journal_shards.py`, `_hookutil.py`), then update all three scripts to import and use it instead of their own copy.

`reconcile-project-board.py`'s version (post-#465) is the most complete of the three:
- Distinguishes "scan_dir unreadable" (`None`) from "scanned, found zero repos" (`[]`) — the other two conflate both into `[]`.
- Properly closes the `os.scandir` handle via a `with` block.
- Catches both `PermissionError` and `FileNotFoundError` — `prune-merged-worktrees.py` currently catches only `PermissionError`, so a nonexistent `--scan-dir` crashes it with an unhandled traceback instead of failing gracefully.

Backport this more-complete behavior into the shared module (recommended) rather than keeping the more conservative behavior, and update all three call sites accordingly.

## Approach

- New `claude/scripts/_repo_scan.py` with `find_git_repos(scan_dir) -> list[str] | None`.
- Update `prune-merged-worktrees.py`, `reclaim-worktree-disk.py`, `reconcile-project-board.py` to import from it, deleting their local copies.
- Add `claude/scripts/tests/test_repo_scan.py` (fixture-based, no mocks, matching `test_reclaim_worktree_disk.py` / `test_worktree_topology.py` convention).
- Update dev-env's `## Testing` section with a numbered item for the new test.

brownm09:
## Summary

- Extracts the near-identical `find_git_repos(scan_dir)` helper duplicated across `prune-merged-worktrees.py`, `reclaim-worktree-disk.py`, and `reconcile-project-board.py` into a new shared module `claude/scripts/_repo_scan.py`, following the repo's existing shared-module convention (`_hookio.py`, `_worktree_liveness.py`, `_journal_shards.py`, `_hookutil.py`).
- Backports `reconcile-project-board.py`'s more-complete behavior into the shared module (recommended, per the tracking issue): catches both `PermissionError` and `FileNotFoundError`, distinguishes an unreadable `scan_dir` (`None`) from a readable-but-empty one (`[]`), and closes the `os.scandir` handle via `with`.
- **Incidental fix:** `prune-merged-worktrees.py --scan-dir <nonexistent path>` previously crashed with an unhandled `FileNotFoundError` traceback (its local copy only caught `PermissionError`). It now fails gracefully like the other two scripts always did.
- **Minor cosmetic change:** `reconcile-project-board.py`'s scan-failure warning loses its `[reconcile-board]` prefix (now identical to the other two scripts' plain `WARNING: cannot scan ...`) — not machine-parsed anywhere (routines key off the `RESULT:` stdout line), documented in ADR-072's Considered Alternatives.
- Adds `claude/scripts/tests/test_repo_scan.py` (fixture-based, no mocks).
- Adds dev-env `CLAUDE.md` Testing items #30 (`reclaim-worktree-disk.py` — backfills a pre-existing gap where that test file existed but was never cross-referenced) and #31 (`_repo_scan.py`).
- Adds [ADR-072](https://github.com/brownm09/dev-env/blob/main/docs/adr/072-shared-repo-scan-module.md) (mirrors ADR-050/057/064, the prior shared-module-extraction ADRs) and an `docs/adr/INDEX.md` row.

Closes #471. Tracked in ADR-070 ("Alternatives Considered" / "Consequences") as a deferred follow-up.

## Testing

Ran dev-env's `## Testing` section (root `CLAUDE.md`):
- Item 1 (hook-script syntax check over `claude/scripts/*.py`) — clean.
- Item 26 (`test_prune_merged_worktrees.py`) — 1 passed.
- New item 30 (`test_reclaim_worktree_disk.py`) — 6 passed.
- New item 31 (`test_repo_scan.py`) — 4 passed.
- Item 29 (`test_reconcile_project_board.py`) — 26 passed.
- Direct-invocation smoke test of all three call sites (`py -3 <script>.py --dry-run`, no `--scan-dir`) to confirm `from _repo_scan import find_git_repos` resolves correctly under Python's standard script-directory `sys.path` injection, not just the test harness's manual `sys.path.insert` — all three ran end-to-end without import errors.

**Tests: 37 passed, 0 skipped, 0 failed** (1 + 6 + 4 + 26, across the four suites above; no other test file's behavior is touched by this change).

Pre-PR suppression grep (`ts-ignore`/`eslint-disable`/`!`/`?? null` patterns) — no matches. Test-integrity grep (skip markers, deleted tests, lowered thresholds) — no matches; no test files deleted. Both patterns target JS/TS conventions (per the CLAUDE.md Scope note) — this is a pure-Python change, and this repo's `claude/scripts/tests/` suite uses a hand-rolled PASS/FAIL harness with no skip primitive to check for. Lockfile check N/A (no `package.json` changed). `baseline_test_failure_tracking` is not enabled for this repo (checked `.claude/hook-config.json`).

- **[documentation] docs/REFERENCE.md not updated** — fixed in `e6d4c49`: added an inline mention of `_repo_scan.py` near the three scripts that import it, matching how the repo's other non-invoked shared modules are documented.
- **[reliability] `find_git_repos` missed `NotADirectoryError`** — fixed in `e6d4c49`: broadened `except (PermissionError, FileNotFoundError)` to `except OSError`, added a regression test (reproduces as WinError 267 on Windows), and updated ADR-072 to describe it as a second incidental fix.

Full review: https://github.com/brownm09/dev-env/pull/478#issuecomment-4859483973


brownm09:
## Review: chore: extract shared _repo_scan module for find_git_repos()

### Summary
Clean, well-tested refactor — behavior-preserving at every call site, with two incidental robustness fixes (one already documented, one caught below). Merge with minor changes: one small reliability gap to close (a natural extension of the fix this PR already makes) and one documentation-maintenance update. Both are small enough to fix in this PR rather than file separately. Diff is appropriately scoped (399 lines, one cohesive change) — no split needed.

---

### Blocking Findings

**[documentation]** `docs/REFERENCE.md`
The PR changes `claude/scripts/_repo_scan.py` (new), `prune-merged-worktrees.py`, `reclaim-worktree-disk.py`, and `reconcile-project-board.py`, but `docs/REFERENCE.md` does not appear in the diff. The root `CLAUDE.md` Documentation Maintenance table requires a `docs/REFERENCE.md` update whenever `claude/scripts/**` changes.
Applying judgment on top of the mechanical trigger: `_repo_scan.py` itself doesn't need its own Utilities-table row (it's a non-invoked shared library module, matching the existing precedent of `_hookutil.py`/`_worktree_liveness.py`/`_journal_shards.py`/`_hookio.py`, none of which have a dedicated row either), and the three modified scripts didn't change their invocation syntax or user-visible behavior. But those same three shared modules ARE documented via a brief inline mention in the relevant prose elsewhere in `docs/REFERENCE.md` (e.g. `_hookio.py` at line 138, `_worktree_liveness.py` at line 51) — `_repo_scan.py` should get the same treatment for discoverability, since a future reader of the three Utilities-table rows has no way to learn they now share this logic.
**Fix:** Add a short inline mention of `_repo_scan.py` near the three scripts' Utilities-table rows in `docs/REFERENCE.md`, linking ADR-072.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/_repo_scan.py:36`
`find_git_repos` catches `except (PermissionError, FileNotFoundError)`. `NotADirectoryError` is a sibling `OSError` subclass, not covered by either — if `--scan-dir` is pointed at a path that exists but is a file (e.g. a typo'd argument), `os.scandir()` raises `NotADirectoryError`, which propagates uncaught and crashes the script exactly like the `FileNotFoundError` gap this same PR already fixes in `prune-merged-worktrees.py`. This is inherited from `reconcile-project-board.py`'s original implementation (not a regression introduced by this PR), but since the PR's whole premise is "back-port the more complete, crash-resistant behavior," leaving this sibling gap unaddressed right next to the fix undercuts that claim. In practice this only matters for a mistyped manual `--scan-dir` invocation — the three routines that call `--scan-dir` all pass a hardcoded, known-good directory, so it's low-likelihood — but the fix is a one-line change while the code is already open.
**Suggestion:** Broaden to `except OSError:` (the common base class for `PermissionError`/`FileNotFoundError`/`NotADirectoryError`, matching the function's documented contract of "or None if scan_dir itself could not be scanned" — for any reason, not just the two or three subtypes anyone happened to enumerate). `ValueError`/`TypeError` (a non-string argument, an embedded null byte) should keep propagating — those indicate a caller bug, not an unusable directory.

---

### Findings Disposition
Both findings above will be fixed in this PR (see the follow-up commit) rather than filed — both are small, directly adjacent to code this PR already touches, and closing them keeps the "more robust than before" claim in the PR description and ADR-072 honest.

---

PR #477, opened 2026-07-01
fix: usage-snapshot.py gates merge success on gh's marker, not exit code

brownm09:
## Summary

`usage-snapshot.py` never got the ADR-049/ADR-050 fix that moved its sibling PostToolUse merge hooks (`post-pr-merge-project.py`, `post-pr-merge-pull.py`, `post-pr-merge-reclaim.py`, `post-merge-tile-checkpoint.py`, `pr-merge-reminder.py`) off exit-code-based merge detection. It gated solely on `tool_response.exitCode != 0` — the exact proxy those ADRs identified as wrong, because a worktree merge exits non-zero on local branch cleanup ("'main' is already checked out") even though the *remote* merge succeeded (issue #275).

This reproduced on dev-env PR #466: merging from a worktree hit that exact cleanup failure, the squash-merge itself succeeded, `post-pr-merge-project.py` correctly moved the linked issue to Done — but no `### Usage Snapshot (post-merge)` block appeared, because the exit-code gate discarded the event before the hook ever read `stdout`/`stderr` or reached the credential/token logic.

- Imports `output_has_merge_marker` / `read_command_output` from `_hookio`.
- Adds a pure `merge_confirmed(command, output)` predicate, gated on gh's output success marker — mirrors `post-pr-merge-project.py`'s `merge_succeeded()`, which the header comment already claimed to "mirror" but didn't.
- Updates the module docstring's payload-shape example and "Exit 0" bullet to match.
- Adds an amendment to `docs/adr/050-shared-hookio-sibling-hook-fixes.md` documenting `usage-snapshot.py` as a sixth hook brought into the pattern.

Not a duplicate of #302 (closed 2026-06-30): that issue covered a different, already-fixed failure class in this same hook (silent skips from an expired/missing OAuth token or unreachable usage API). This is a separate gate that bails out *before* the credential/token logic ever runs.

Closes #474

## Testing

Ran from the worktree root (`claude/scripts/tests/test_usage_snapshot.py` is the change-specific suite; the others are sanity checks since this touches a shared dependency):

```
py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py
# all files parse

py -3 claude/scripts/tests/test_usage_snapshot.py
# Tests: 11 passed, 0 skipped, 0 failed

py -3 claude/scripts/tests/test_hookio.py   # sanity check of the shared dependency (unmodified)
[...]

Issue #474, opened 2026-07-01
usage-snapshot.py silently drops the post-merge snapshot on worktree merges (exit-code gate, not marker-based like sibling hooks)


brownm09:
## Symptom

While merging dev-env PR #466 from a worktree (`.claude/worktrees/routines-junction-docs`), `gh pr merge --squash --delete-branch` exited 1 with the known worktree-conflict error (`fatal: 'main' is already checked out at 'C:/Users/brown/Git/dev-env'` — see docs/REFERENCE.md's "merging a PR from a worktree" runbook; also seen on PRs #327/#329). The squash-merge itself succeeded (`gh pr view --json state,mergedAt,mergeCommit` confirmed merged), and `post-pr-merge-project.py` correctly moved the linked issue to Done despite the exit-1. But no `### Usage Snapshot (post-merge)` block appeared.

## Root cause

`claude/scripts/usage-snapshot.py` gates on the Bash tool's exit code:

```python
if data.get("tool_response", {}).get("exitCode", 0) != 0:
    sys.exit(0)
```

This is exactly the check that ADR-049/ADR-050 identified as wrong for merge detection and fixed in sibling PostToolUse hooks — `post-pr-merge-project.py`, `post-pr-merge-pull.py`, `post-pr-merge-reclaim.py`, and (later) `post-merge-tile-checkpoint.py` / `pr-merge-reminder.py` — specifically because a worktree merge exits non-zero on local branch cleanup ("'main' is already checked out") even though the *remote* merge succeeded (issue #275). Those hooks now gate on gh's output success marker (`output_has_merge_marker()` in `claude/scripts/_hookio.py`) instead of, or in addition to, the exit code.

`usage-snapshot.py` was never brought into that fix:
- It does not import `_hookio` at all.
- It never reads `tool_response.stdout`/`stderr` or checks for the merge success marker.
- Its only "did this succeed" signal is the exit-code check above, which is precisely backwards for the worktree case documented in issue #275.

This reproduces the exact scenario from PR #466: worktree merge exits 1 → `usage-snapshot.py` bails at the top of `main()` before ever loading credentials or fetching usage data → no snapshot, no diagnostic (this exit path is intentionally silent per the module docstring).

## Not a duplicate of #302

[...]

PR #484, opened 2026-07-01
docs: update stale worktree-merge usage-snapshot note

### LINKED ISSUE #483: docs/REFERENCE.md: 'no post-merge usage snapshot' worktree-merge note is now stale (fixed by #474/#477)
## Symptom

`docs/REFERENCE.md` → "Merging a PR developed in a worktree" → **Secondary effect — no post-merge usage snapshot** (lines ~637-640) documents:

> Because the `gh pr merge` invocation exits non-zero on the failed local-checkout tail, the `PostToolUse` post-merge hook does **not** emit its `### Usage Snapshot (post-merge)` block — it keys off a clean `gh pr merge`. When merging from a worktree, expect the snapshot to be absent and capture usage by other means for the journal stub.

This described real, correct behavior when it was written, but it documents the *symptom* of the exact bug fixed in [dev-env#474](https://github.com/brownm09/dev-env/issues/474) / [PR #477](https://github.com/brownm09/dev-env/pull/477): `usage-snapshot.py` gated on `tool_response.exitCode != 0` instead of gh's output success marker. Now that the fix has landed (`merge_confirmed()` gates on the marker via `_hookio`, matching `post-pr-merge-project.py`), the snapshot should fire correctly even when the worktree-merge local-cleanup step fails — validated in practice: merging PR #477 itself hit this exact `fatal: 'main' is already checked out` case, and `post-pr-merge-project.py` (same marker-based detection) correctly moved dev-env#474's linked issue to Done, confirming gh's success marker was present in the real payload.

This paragraph now tells operators to expect a permanent limitation that no longer exists — stale documentation actively misleading future sessions away from investigating a real regression if the snapshot is ever absent again for an unrelated reason.

## Fix

Update the "Secondary effect" paragraph in `docs/REFERENCE.md` to reflect that the snapshot now fires on worktree merges via the marker-based gate (linking dev-env#474/PR #477), while still noting the residual, legitimate silent-exit paths that remain in `usage-snapshot.py` by design (credentials file absent, token unrecoverable, usage API unreachable) — those are unrelated to the worktree-exit-code case and still apply.

## Acceptance criteria

- [ ] `docs/REFERENCE.md`'s worktree-merge runbook no longer claims the snapshot is unconditionally absent on worktree merges
- [ ] The paragraph links to dev-env#474 / PR #477 as the fix
- [ ] Remaining legitimate silent-exit paths (credentials/token/API) are still accurately described

brownm09:
## Summary

`docs/REFERENCE.md`'s "Merging a PR developed in a worktree" runbook documented the "no post-merge usage snapshot" effect as permanent, expected behavior. That was the symptom of the exit-code-gating bug fixed in dev-env#474 / PR #477 (`usage-snapshot.py` now gates on gh's output success marker, matching `post-pr-merge-project.py`, instead of `tool_response.exitCode`).

Discovered while merging PR #477 itself: it hit this exact worktree-cleanup failure, and `post-pr-merge-project.py` (same marker-based detection) correctly moved dev-env#474's linked issue to Done — confirming the fix's premise in practice and making the existing doc paragraph stale.

Updates the paragraph to:
- Note the fix (linking #474/#477) instead of describing the old behavior as permanent.
- List the *remaining* legitimate silent-exit paths in `usage-snapshot.py` (missing/unparseable credentials, expired refresh token, unreachable usage API) that are unrelated to this fix and still apply.
- Record a separately-observed nuance: a PostToolUse hook's stderr doesn't reliably surface to the assistant in chat when the parent `gh pr merge` call itself is shown as an error — confirm a hook ran via its side effect (e.g. board status), not the absence/presence of visible reminder text.

Closes #483

## Testing

Docs-only change (no `.py` files touched). Ran the standard pre-PR checks anyway:

```
# all files parse (unaffected by this change)
```

No automated tests apply — pure prose documentation change, no new testable behavior (coverage gate N/A). Suppression grep and test-integrity grep against `origin/main` both clean (no matches, exit 1). `baseline_test_failure_tracking` not enabled for dev-env (N/A).

## ADR-warrant check

N/A — documentation correction only; doesn't touch a rule/hook/skill/settings decision, doesn't restructure a `claude/` directory, doesn't set a new cross-CLAUDE.md rule. The underlying decision (marker-over-exit-code) was already recorded in the ADR-050 amendment landed in PR #477.

`/review` posted at https://github.com/brownm09/dev-env/pull/484#issuecomment-4859581696 (`blocking=0 non_blocking=2`).

- Both non-blocking [documentation] findings (observed-vs-inferred wording precision in the new paragraph) — fixed in `1f817e3`.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -641,13 +641,18 @@ pre-merge tooling. This is now auto-corrected: `post-pr-merge-pull.py` parks the
 failed local-checkout tail (above) discarded the snapshot even though the remote merge had
 succeeded. The hook now gates on gh's output success marker instead (`merge_confirmed()`,
 matching `post-pr-merge-project.py`'s marker-based detection) and fires correctly on worktree
-merges — validated in practice merging PR #477 itself, which hit this exact failure and still
-moved dev-env#474's linked issue to Done via the same marker. The snapshot can still be
+merges — confirmed in practice merging PR #477 itself, which hit this exact failure: the real
+payload contained gh's success marker (`post-pr-merge-project.py`, using the same `_hookio`
+dependency, correctly moved dev-env#474's linked issue to Done), and credentials/token were
+independently healthy. `usage-snapshot.py`'s own output was not directly observed in chat (see
+the next note), but it shares the identical marker-detection call. The snapshot can still be
 legitimately absent for reasons unrelated to the worktree-exit-code case: `.credentials.json`
 missing or unparseable, an expired refresh token, or the usage API unreachable after one retry —
 all intentionally silent-or-advisory per the hook's own docstring, not a regression of this fix.
-Separately: a PostToolUse hook's own stderr does not reliably surface to the assistant in chat
-when the *parent* `gh pr merge` call itself is shown as an error (non-zero exit) — confirm a hook
+Separately: a PostToolUse hook's own stderr was not observed to surface to the assistant in chat
+in either of the two instances behind this fix — worth confirming with more data points before
+treating as a firm platform rule — when the *parent* `gh pr merge` call itself is shown as an
+error (non-zero exit); until then, confirm a hook
 actually ran by checking its side effect (e.g. the linked issue's board status) rather than
 assuming absence of visible reminder text means the hook didn't fire.
 
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -636,10 +636,20 @@ pre-merge tooling. This is now auto-corrected: `post-pr-merge-pull.py` parks the
 `main`), then `git -C ~/Git/dev-env checkout main`. See
 [ADR-058](adr/058-worktree-squatting-main-detection-correction.md) (incident: dev-env#396).
 
-**Secondary effect — no post-merge usage snapshot.** Because the `gh pr merge` invocation exits
-non-zero on the failed local-checkout tail, the `PostToolUse` post-merge hook does **not** emit its
-`### Usage Snapshot (post-merge)` block — it keys off a clean `gh pr merge`. When merging from a
-worktree, expect the snapshot to be absent and capture usage by other means for the journal stub.
+**Secondary effect — no post-merge usage snapshot (fixed by dev-env#474 / PR #477).** Before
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: docs: update stale worktree-merge usage-snapshot note

### Summary
Ready to merge with a minor wording tweak. This is an 18-line, docs-only correction (`docs/REFERENCE.md` only) that accurately replaces a now-stale "permanent limitation" paragraph with the real fix history. Every factual claim checks out against the actual code and the project-board evidence gathered this session. One non-blocking finding: two sentences read as stronger claims than what was actually directly observed vs. inferred, which is worth tightening given this repo's own Error Message Diligence principle (ADR-034) — especially since this text is going into a permanent runbook.

---

### Non-Blocking Findings

**[documentation]** `docs/REFERENCE.md:642` ("validated in practice merging PR #477 itself...")
The sentence is *technically* precise (it attributes the validation to the issue moving to Done via the marker, which was directly confirmed), but a reader skimming it could easily come away thinking `usage-snapshot.py`'s own snapshot output was directly observed to render correctly this time. What was actually confirmed: the real payload contained gh's success marker (proven indirectly via `post-pr-merge-project.py`, a *different* hook using the same `_hookio` dependency) and credentials/token were independently healthy. The snapshot's own stderr was not directly observed.
**Suggestion:** Make the inference explicit, e.g.: "...confirmed the marker was present in the real payload (`post-pr-merge-project.py`, using the same `_hookio` dependency, correctly moved dev-env#474's linked issue to Done); `usage-snapshot.py`'s own output was not directly observed in chat (see the next note), but it shares the identical marker-detection call."

**[documentation]** `docs/REFERENCE.md:648` ("does not reliably surface to the assistant in chat")
"Reliably" implies a pattern established across multiple observations. The evidence for this session is one direct instance (this PR's own merge) plus one inferred prior instance (the original dev-env#474 report, where the reporter also inferred the board-move after the fact rather than watching hook output live). That's suggestive, not yet "reliable" in the sense of a confirmed repeated behavior.
**Suggestion:** Soften to something like "was not observed to surface... in the two instances behind this fix and this note — worth confirming with more data points before treating as a firm platform rule."

---

### Findings Disposition
Both findings are wording-precision suggestions in the very paragraph this PR is adding — fixing them in this PR is direct, not out of scope.

---

PR #491, opened 2026-07-01
fix: gate merge-detection hooks on gh's success marker, not exit code

### LINKED ISSUE #485: pr-merge-reminder.py and 3 sibling hooks misdetect any exit-0 'gh pr merge ...' invocation (e.g. --help) as a completed merge
## Problem

Four PostToolUse hooks share the same success-detection predicate:

```python
return exit_code == 0 or output_has_merge_marker(output)
```

- `claude/scripts/pr-merge-reminder.py` (`_is_successful_merge_call`)
- `claude/scripts/post-pr-merge-pull.py` (`is_successful_merge`)
- `claude/scripts/post-pr-merge-reclaim.py` (`is_successful_merge`)
- `claude/scripts/post-merge-tile-checkpoint.py` (`is_successful_merge`)

Because `exit_code == 0` is true for *any* successful invocation — not just a completed merge — any command matched as "gh pr merge ..." that happens to exit 0 is misdetected as a successful merge. Concretely: `gh pr merge --help` exits 0 and is misdetected, firing the journal-update reminder, the tile-checkpoint reminder, a local-`main` fast-forward attempt, and a worktree-disk reclaim spawn, for a command that never merged anything.

Reproduced live during this session: running `gh pr merge --help 2>&1 | grep ...` (per the CLAUDE.md CLI Scripting Checklist's "run `--help` first" step) fired the full `[journal-reminder]` cascade with nothing actually merged.

## Why this is the wrong gate

`output_has_merge_marker()` is the reliable signal — gh always prints its "Squashed and merged pull request #N" marker on a real merge, even from a worktree where local cleanup makes the exit code non-zero (issue #275). The `exit_code == 0` branch was added to *also* trust a clean exit, but a clean exit doesn't mean a merge happened — `--help`, a queued `--auto`, or any other non-merge-completing invocation also exits 0.

This is not a new problem in this codebase: `post-pr-merge-project.py::merge_succeeded()` was already marker-only by original design, and `usage-snapshot.py::merge_confirmed()` was just fixed the same way today (PR #477, closing #474) for the opposite-direction bug (exit-code-only gating missed real worktree merges). ADR-050 explicitly documented the OR as a deliberate, "harmless" choice for the pull/reclaim/tile-checkpoint hooks — this issue is that assumption turning out to be wrong: the false positive isn't limited to "premature" races, it fires on non-merge commands entirely.

## Fix

Converge all four hooks onto marker-only detection, matching `merge_succeeded()` / `merge_confirmed()`. Verified via `gh pr merge --help` that no quiet/dry-run flag exists, so a real merge always prints the marker — this is a strict improvement with no false-negative regression.

Also needs: updated tests in the four `test_*.py` files (replace the "exit 0 alone fires" tests that pinned the bug), updated CLAUDE.md `## Testing` prose for items #10/#15/#23/#28, and a second amendment to `docs/adr/050-shared-hookio-sibling-hook-fixes.md` (which explicitly documented the OR as deliberate).

## Context

Surfaced while checking `gh pr merge --help`'s flag syntax during PR #476.

brownm09:
## Summary

- `pr-merge-reminder.py`, `post-pr-merge-pull.py`, `post-pr-merge-reclaim.py`, and `post-merge-tile-checkpoint.py` all shared the predicate `exit_code == 0 or output_has_merge_marker(output)`. Since `exit_code == 0` is true for *any* successful invocation, an exit-0 command matched as `gh pr merge` that never actually merged anything — `gh pr merge --help`, a queued `--auto` — was misdetected as a completed merge, firing the journal-update reminder, the tile-checkpoint reminder, a local-`main` pull, and a `node_modules` reclaim spawn for nothing.
- Reproduced live in this session while checking `gh pr merge --help`'s flag syntax per the CLI Scripting Checklist's own "run `--help` first" step — the full `[journal-reminder]` cascade fired with nothing merged.
- All four hooks now gate solely on the marker, matching `post-pr-merge-project.py::merge_succeeded()` and the just-merged `usage-snapshot.py::merge_confirmed()` (PR #477). `gh pr merge --help`'s own flag list confirms there's no quiet/dry-run flag, so a real merge always prints the marker — this drops a parameter with no false-negative regression.
- Amends [ADR-050](https://github.com/brownm09/dev-env/blob/main/docs/adr/050-shared-hookio-sibling-hook-fixes.md) (Amendment 2), which had documented the OR as a deliberate, "harmless" choice for these four hooks — that framing didn't account for exit-0 *non-merge* invocations, only premature-merge races.

## Testing

Ran dev-env's `## Testing` items 1 (hook-script syntax check), 10, 15, 23, 28 (the four affected hooks' tests), plus 13 (`test_hookio.py`) and 14 (`test_post_pr_merge_project.py`) as sanity checks on unmodified sibling hooks that share `_hookio`:

```
py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py   # clean, no output
py -3 claude/scripts/tests/test_pr_merge_reminder.py            # 31 passed
py -3 claude/scripts/tests/test_post_pr_merge_pull.py            # 7 passed
py -3 claude/scripts/tests/test_post_pr_merge_reclaim.py         # 4 passed
py -3 claude/scripts/tests/test_post_merge_tile_checkpoint.py    # 4 passed
py -3 claude/scripts/tests/test_hookio.py                        # 17 passed (sanity, unmodified)
py -3 claude/scripts/tests/test_post_pr_merge_project.py         # 16 passed (sanity, unmodified)
```

**Tests: 79 passed, 0 skipped, 0 failed** (combined across all six files).

Added a `gh pr merge --help`-shaped regression test to each of the four affected test files (exit 0, no marker → must not fire), and replaced the old "exit 0 alone fires" tests (which pinned the bug) with "marker present fires" tests.

**Suppression check (Rule 3):** clean, no matches.

**Test-integrity check (Rule 3):** the grep flagged several lines, all confirmed false positives on manual review — `xit\(` (meant to catch Jasmine/Mocha's skip-test function) is a plain substring match, so it also matches inside Python's `sys.exit(0)` calls and my renamed `test_merge_call_clean_exit()` function name (removed in this diff), since both contain the literal substring `xit(`. None of the matches are actual skip markers, deleted tests, or lowered thresholds. Filed as a follow-up to fix the regex itself (see linked task below) rather than fixing inline here, since it's a pre-existing imprecision in the shared global check, not something this PR introduced.

No `package.json` touched — lockfile policy N/A. No `.claude/hook-config.json` in dev-env — baseline-test-failure policy N/A (not opted in).

## Risk audit

1. **Testing** — new regression tests per file (above).
2. **Observability** — N/A, no new output paths; a boolean-gate tightening on existing stderr/exit-code diagnostics.
3. **Security** — N/A, pure local boolean logic on already-trusted hook payload data.
4. **Resilience** — this change *is* the failure-mode correction; fully revertible via this PR.
5. **Performance** — N/A, negligible (same regex/substring checks, one fewer branch).
6. **Data integrity** — N/A, stateless hooks, no persisted state.

Per [/review](https://github.com/brownm09/dev-env/pull/491#issuecomment-4859692814):

- **Blocking [documentation]** `docs/REFERENCE.md`'s Hooks table had three stale rows describing the old "exit 0 or marker" trigger — **fixed in `6cface0`**.
- **Non-blocking [correctness]** A chained `gh pr create && gh pr merge --auto` command can suppress the create reminder if the merge sub-check fails, widened (not introduced) by this PR — **filed as [dev-env#494](https://github.com/brownm09/dev-env/issues/494)** (Impact: Low; judged out of scope for this PR's marker-vs-exit-code fix).

Closes #485


brownm09:
## Review: fix: gate merge-detection hooks on gh's success marker, not exit code

### Summary
This is a well-scoped, well-tested fix for a confirmed, live-reproduced bug (`gh pr merge --help` misdetected as a completed merge across four hooks). The fix converges all four hooks onto the marker-only pattern already established by `post-pr-merge-project.py` and the same-day `usage-snapshot.py` fix, and is backed by regression tests in all four affected test files plus two sanity runs on unmodified sibling hooks. One blocking finding: `docs/REFERENCE.md`'s Hooks table has three rows that still describe the old "exit 0 or marker" behavior as the trigger condition, and are now factually wrong. One non-blocking finding: a narrow, pre-existing edge case in `pr-merge-reminder.py` is incidentally widened by this fix. Diff is appropriately scoped (335 changed lines, single concern, no unrelated changes) — no split needed.

---

### Blocking Findings

**[documentation]** `docs/REFERENCE.md:217,219,220`
Three Hooks-table rows describe the trigger condition using the exact behavior this PR removes:
- Line 217 (`post-merge-tile-checkpoint.py`): "Command contains `gh pr merge` **and the merge succeeded (exit 0 or a stdout success marker** ...)"
- Line 219 (`post-pr-merge-pull.py`): "...and the merge succeeded (exit 0 or a stdout/stderr success marker...)"
- Line 220 (`post-pr-merge-reclaim.py`): "...and the merge succeeded (exit 0 or a stdout success marker...)"

After this PR, none of these three hooks consult the exit code at all — they gate solely on the output marker. These rows now describe code that no longer exists, which is exactly the kind of drift the CLAUDE.md doc-reconciliation checkpoint exists to catch ("does this change ... modify the behavior of a skill, hook, script, or routine?"). This PR already updated the *test-description* prose in the top-level `CLAUDE.md` `## Testing` section (items 10/15/23/28) but missed `docs/REFERENCE.md`'s Hooks table, which independently describes the same trigger condition.
**Fix:** Update the three "Trigger" column entries to say the merge succeeded when the output carries gh's success marker (regardless of exit code), matching the new `is_successful_merge()` behavior — mirroring how `post-pr-merge-project.py`'s existing REFERENCE.md row (if present) or the CLAUDE.md prose already phrases the marker-only gate.

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/pr-merge-reminder.py:229-235`
`main()`'s dispatch is `if is_merge: (early sys.exit(0) if no marker) elif exit_code != 0: sys.exit(0)`. For a single-purpose command this is fine. But for a *chained* command matching both `is_create` and `is_merge` as separate top-level statements (e.g., `gh pr create --fill && gh pr merge --auto`), the merge sub-check now runs `output_has_merge_marker()` against the *combined* output of both commands. A queued `--auto` (or `--help`, though that combination is unrealistic) prints no completion marker, so the merge check fails and the function exits immediately — before the `messages` section that would otherwise build the `is_create` reminder ever runs. The legitimate PR-create reminder is silently suppressed as a side effect.

This isn't new in kind — before this PR, the same suppression already happened whenever the chained merge sub-command failed outright (non-zero exit, no marker); this PR *widens* the trigger to also include exit-0-but-incomplete merges (`--auto`, `--help`), which is a materially larger set of chained-command shapes. The likelihood is low (chaining `gh pr create` with `gh pr merge` in one Bash call is an uncommon pattern, and CLAUDE.md's own auto-merge guidance suggests `--auto` is typically invoked as a separate, later call after the full post-PR checklist), and there's a partial backstop elsewhere (the "Incomplete journal detected" check in CLAUDE.md's journal-composition rules would eventually flag a missing stub). Fixing it properly means restructuring the message-building control flow so a failed/incomplete merge sub-check doesn't short-circuit an otherwise-valid create (or push) message — a larger, separate change from this PR's marker-vs-exit-code scope.
**Suggestion:** File a follow-up issue to decouple the three message branches (`is_create`, `is_merge`, `is_push`) so a negative result on one doesn't suppress the others; reference this PR and the specific chained-command scenario above.

---

### Findings Disposition
Per ADR-028, both findings above must be fixed-in-PR or filed-and-linked before merge; recording disposition here per the PR-body template.

---

PR #492, opened 2026-07-01
fix(hooks): exclude heredoc-command-substitution bodies from canonical-mutate-guard classification

### LINKED ISSUE #481: canonical-mutate-guard: heredoc body line starting with a mutating verb false-positives
## Summary

pre-tool-use-canonical-mutate-guard.py's heredoc protection (classify(), career-playbook
#442 lesson) only covers a mutating verb MENTIONED mid-line inside a heredoc body (e.g.
"Remember: do not run git checkout -b here."). It does not cover the case where a heredoc
body line literally BEGINS with the verb -- e.g. a markdown code-fence example inside a
`gh issue create --body "$(cat <<'EOF' ... EOF)"` invocation that itself contains a line
like `git commit -m "..." -- some/path`.

Since classify() splits the whole command on \n (_SEGMENT_SPLIT includes \n) and checks
each resulting segment against `^git(?:.exe)?\s+`, a heredoc body line that starts with
"git commit" or "git checkout" is indistinguishable from a real invoked command -- the
anchor-at-segment-start design (the correct fix for #442's mid-line-mention case) does not
distinguish "this line happens to start with git because it's example prose" from "this
line is genuinely git being invoked."

## Reproduction (2026-07-01, this session)

Ran (paraphrased -- see actual command in dev-env#480, which is what surfaced this):

  gh issue create --repo brownm09/dev-env --title "..." --body "$(cat <<'EOF'
  ## Suggested fix
  ...
  git commit -m "..." -- sessions/<project>/YYYY-MM-DD_HHMMSS.stub.md ...
  EOF
  )"

cwd was career-playbook's canonical checkout (no worktree). The hook blocked with:

  [canonical-mutate-guard] BLOCKED: ... Command : git commit -m "..." -- sessions/<project>/...
  Root : C:/Users/brown/Git/career-playbook

despite the actual invoked command being `gh issue create` (a remote API call touching no
local git state at all) -- the "Command" the hook reported was a line from inside the
quoted (`<<'EOF'`, no shell expansion) heredoc body string, not anything actually executed
as a shell command.

## Impact

False positive, not a missed block -- fails safe (blocks legitimate work, does not let a
real mutation through). But it will recur for anyone documenting example git commands
(this exact CLAUDE.md file does, in the Stub file workflow section) via `gh issue
create`/`gh pr create`/`gh pr comment --body-file` etc. from a canonical checkout. The
correct workaround (ALLOW_CANONICAL_MUTATE=1) is safe here specifically because the
override is a per-invocation, visible, human-confirmable action -- but it trains users to
reach for the override reflexively on a hook that is supposed to be a meaningful checkpoint,
which erodes its value on the cases it's actually meant to catch.

## Suggested fix

Options to consider (not prescribing one):
1. Restrict classification to segments that are plausibly real shell commands, not
   arbitrary heredoc/string-literal content -- e.g. only classify segments from the
   top-level command structure, not ones nested inside a `$(...)`, quoted heredoc
   (`<<'EOF'`), or a `--body`/`-m`/similar string-literal argument.
2. Detect when the mutating-looking segment is inside a `$(cat <<'MARKER' ... MARKER)`
   span (a quoted heredoc used as a command-substitution argument) and skip segments
   between the opening and closing MARKER lines entirely, similar to how classify()
   already special-cases `cd` and `-C`/`--git-dir` redirects.
3. At minimum, document the gap and the override as the intended recovery in the hook's
   own module docstring and ADR-071, so it reads as a known, accepted limitation rather
   than an unexplained false-positive when someone hits it.

## Scope note

Filed rather than fixed in-session -- discovered as a side effect of closing out an
unrelated task (dev-env#469), and a proper fix likely needs test coverage for the
heredoc-as-command-substitution-argument case specifically (distinct from the existing
`test_heredoc_mention_does_not_trigger` case, which covers mid-line mention, not
line-start-of-heredoc-body).

brownm09:
## Summary

Closes #481.

`classify()`'s heredoc-mention fix (career-playbook#442: a mutating verb merely *mentioned* mid-line inside a heredoc body must not trigger) does not cover the case where a heredoc body line *begins* with the verb — e.g. a markdown code-fence example line like a `git commit` invocation inside a `gh issue create --body "$(cat <<'EOF' ... EOF)"` call. After the newline-split, that line is indistinguishable from a genuinely invoked command, so the hook falsely blocked a `gh issue create` call (a remote API call touching no local git state at all) because its `--body` argument's heredoc happened to contain a commit example line.

## Changes

- Added `_strip_heredoc_command_subs()` to `claude/scripts/pre-tool-use-canonical-mutate-guard.py`, which removes every `$(cat <<[-]['"]MARKER['"] ... MARKER)` span from the command before segment classification runs — option 2 from the issue. Scoped narrowly to the `$(cat <<'MARKER'...)` idiom this repo's own `CLAUDE.md` documents extensively (commit-message / gh-body heredocs), not a general heredoc parser.
- Applied the same stripping to `_has_override()` — it has the identical false-positive shape (a heredoc body line documenting the override syntax could otherwise silently bypass the block for an unrelated real mutating segment in the same command).
- Updated the module docstring's numbered logic list (new step 3; old steps 4-7 renumbered to 4-8) and both functions' docstrings to document the new behavior.
- Added three tests to `test_canonical_mutate_guard.py`:
  - `test_heredoc_command_substitution_body_not_classified` — the exact reproduction shape (line-start-of-heredoc-body-used-as-command-substitution-argument), distinct from the pre-existing `test_heredoc_mention_does_not_trigger` (mid-line mention).
  - `test_real_command_after_heredoc_catsub_still_classified` — a real mutating command chained after the heredoc-fed call must still be caught, proving the fix stays precise (doesn't over-strip).
  - `test_override_mention_at_heredoc_body_line_start_does_not_bypass` — covers the `_has_override()` fix.

## Testing

Ran per `CLAUDE.md` → `## Testing`:

```
```
Tests: 28 passed, 0 skipped, 0 failed

```
```
No output — all hook scripts parse cleanly (exit 0).

Also hand-verified the exact reproduction shape plus several edge cases (single-quoted / double-quoted / unquoted heredoc markers, a real command correctly still caught when chained after the heredoc closes, the pre-existing mid-line-mention case unaffected) with a standalone sanity script before writing the test suite additions — all passed.

No suppressions, skip markers, deleted tests, or lowered thresholds in this diff.

- **[documentation] Missing Hooks-table entries for `pre-tool-use-canonical-mutate-guard.py`** — fixed in [518233b](https://github.com/brownm09/dev-env/commit/518233b). Added rows to `README.md`'s Hooks table and `docs/REFERENCE.md`'s `### PreToolUse` → `#### Bash hooks` table (the hook was added by PR #468/ADR-071 but never got a table entry in either file).


brownm09:
## Review: fix(hooks): exclude heredoc-command-substitution bodies from canonical-mutate-guard classification

### Summary

Well-scoped, thoroughly-tested bug fix — ready to merge after one documentation gap is closed. The core fix (`_strip_heredoc_command_subs()`, applied to both `classify()` and `_has_override()`) correctly closes dev-env#481 and stays precise: a real mutating command chained after the heredoc-fed call is still caught (`test_real_command_after_heredoc_catsub_still_classified`), and the analogous `_has_override()` gap is fixed too, which wasn't explicitly asked for in the issue but is the same root cause. The one blocking finding is a documentation gap, not a defect in this diff's own logic — `pre-tool-use-canonical-mutate-guard.py` was added by PR #468 earlier today and never got a Hooks-table entry in `README.md`/`docs/REFERENCE.md`; this PR is the first to touch that file since, so the doc-reconciliation check surfaces it here. Diff is appropriately scoped (194 lines, single-purpose, all 28 tests pass).

---

### Blocking Findings

**[documentation]** `README.md` / `docs/REFERENCE.md`
`pre-tool-use-canonical-mutate-guard.py` (added via PR #468 / ADR-071, merged earlier today) has no row in either `README.md`'s Hooks table or `docs/REFERENCE.md`'s `### PreToolUse` → `#### Bash hooks` table — confirmed by grep against `origin/main` for both files. Per dev-env's own `CLAUDE.md` Documentation Maintenance table, an event-driven hook script under `claude/scripts/` requires a Hooks-table entry in both files. This PR modifies that same file without adding the missing entries, which is what the doc-reconciliation check is designed to catch (Step 2b).
**Fix:** Add a row for `pre-tool-use-canonical-mutate-guard.py` to `README.md`'s Hooks table and to `docs/REFERENCE.md`'s `#### Bash hooks` table under `### PreToolUse`, matching the format of the neighboring `pre-merge-findings-gate.py` / `pre-tool-use-worktree-path-check.py` rows (trigger condition, what it does, ADR-071 link).

---

### Findings Disposition

Per ADR-028, will be fixed in this PR (adding the missing table rows) before merge — see the updated "Review findings disposition" section in the PR body once applied.

---

Issue #489, opened 2026-07-01
Investigate: did gh's merge-success marker actually reach the PostToolUse hooks during the PR #476 worktree-merge exit-1?


brownm09:
## Problem

Merging PR #476 via `gh pr merge https://github.com/brownm09/dev-env/pull/476 --squash --delete-branch` from a Claude-managed worktree exited 1 with (reportedly) only

    failed to run git: fatal: 'main' is already checked out at 'C:/Users/brown/Git/dev-env'

visible in the result — no `Squashed and merged pull request #476` success marker text was visible. `gh pr view` afterward confirmed the remote merge genuinely succeeded. None of the six PostToolUse hooks that gate on `_hookio.output_has_merge_marker()` (`pr-merge-reminder.py`, `post-pr-merge-pull.py`, `post-pr-merge-reclaim.py`, `post-merge-tile-checkpoint.py`, `post-pr-merge-project.py`, `usage-snapshot.py`) visibly fired — no journal reminder, no tile-checkpoint reminder, no board move, no canonical-main sync.

This is exactly the failure mode issue #275 / ADR-050 (+ its 2026-07-01 amendment for dev-env#474) already fixed: gh exits non-zero on local worktree cleanup, but the remote merge succeeded, and the shared marker-based detection is supposed to catch this. If the marker genuinely never reached the hooks, that fix has a gap for the exact failure shape it was written for.

## Research so far

`gh`'s own source (`pkg/cmd/pr/merge/merge.go`, `mergeRun`) confirms the success message is printed via `infof()` **before** `deleteLocalBranch()` (the step that fails with the checkout error) ever runs:

    if err := ctx.merge(); err != nil { return err }              // prints "✓ Squashed and merged pull request #N (Title)"
    if err := ctx.deleteLocalBranch(); err != nil { return err }  // this is what fails

So gh *should* have printed the marker before hitting the fatal error. If the marker text really was absent from what the hooks received, the gap is somewhere between gh's stdout/stderr and the hook payload — or the hooks never ran at all (the PostToolUse-inert-in-background-sessions class of issue, ADR-053/ADR-055, would explain the *total* silence across all six hooks equally well, and would need a completely different fix, if any is possible from hook code at all).

## Plan

[...]

PR #476, opened 2026-07-01
fix(hooks): scope merge-reminder repo label to explicit --repo flag

brownm09:
## Summary
- `pr-merge-reminder.py`'s `gh pr merge` reminder called `effective_merge_dir(command, cwd)` unconditionally, which only resolves a `cd <path> &&` chain prefix or falls back to cwd (ADR-067) - it had no awareness of an explicit `--repo owner/repo` flag.
- `post-pr-merge-pull.py`'s `extract_repo` (also touched by ADR-067) already checks `--repo` *first*, ahead of cd-chain/cwd resolution - this PR ports that same precedence to the reminder's `repo:` line via a new `_effective_merge_repo(command, cwd)` helper, scoped to `pr-merge-reminder.py` only.
- `effective_merge_dir` in `_hookio.py` is untouched - it stays a pure directory-resolution helper for `post-pr-merge-pull.py`'s local-clone `git -C` operations, which need an actual filesystem path, not a GitHub owner/repo label.

## Repro (now fixed)
From `C:\Users\brown\Git\dev-env`:
```
gh pr merge 110 --repo brownm09/engineering-journal --squash --delete-branch
```
Before: reminder printed `repo: C:\Users\brown\Git\dev-env`. After: `repo: brownm09/engineering-journal`.

## Testing
Ran `py -3 claude/scripts/tests/test_pr_merge_reminder.py` (item 28 in dev-env's CLAUDE.md `## Testing` section) - added 3 new cases covering `--repo`-flag precedence (over cwd, over a cd-chain, and the no-flag fallback to `effective_merge_dir`).

Tests: 32 passed, 0 skipped, 0 failed

Also ran the hook-script syntax check (item 1): `py -3 -c "import ast,sys; ..." claude/scripts/*.py` - clean, no output.

Pre-PR suppression/test-integrity greps (both required before `gh pr create`) both exited clean (no matches). No deleted test files. `baseline_test_failure_tracking` is not enabled for dev-env (no `.claude/hook-config.json`) - N/A.

## ADR-warrant
[...]

PR #493, opened 2026-07-01
fix(hooks): pull_main() falls back to ff-only pull when canonical is on main

### LINKED ISSUE #488: post-pr-merge-pull.py: pull_main() can never fast-forward dev-env's own canonical (fetch-into-ref fails when main is already checked out)
## Problem

`claude/scripts/post-pr-merge-pull.py`'s `pull_main()` fast-forwards a repo's local `main` after a successful `gh pr merge` using:

    git -C <local_path> fetch origin main:main

This technique intentionally avoids requiring `main` to be checked out — per the docstring, it "updates the local main ref even when a feature branch is currently checked out." That's correct for repos where `local_path` has a feature branch checked out. It is never correct for `brownm09/dev-env`'s own canonical checkout (`C:/Users/brown/Git/dev-env`), because dev-env's architecture rule (CLAUDE.md → Dev-Env Architecture) requires the canonical to **always** stay on `main` — its working tree is symlinked into `~/.claude/`. Git refuses to fetch into a ref that is currently checked out anywhere:

    fatal: refusing to fetch into branch 'refs/heads/main' checked out at 'C:/Users/brown/Git/dev-env'

So `pull_main()` is **guaranteed to fail every single time** a dev-env PR merges from a worktree — which, per the same architecture rule, is the *only* way dev-env PRs ever merge (the canonical never holds a feature branch). The failure is silent from the user's perspective (the hook exits 0; the `[post-merge-pull] ... git fetch failed` message is easy to miss), so the canonical silently drifts behind `origin/main` until the next prompt's `dev-env-sync` catches up — defeating the entire point of issue #275 (avoid waiting for the next prompt) for dev-env's own repo specifically.

## Confirmed by

- Merging PR #476 (2026-07-01) left the canonical `C:/Users/brown/Git/dev-env` behind `origin/main`.
- Manually running `git fetch origin main:main` against the canonical reproduced the exact `refusing to fetch into branch 'refs/heads/main' checked out` error.
- Code review of `post-pr-merge-pull.py` (`pull_main`, lines ~86-114) confirms the technique has no fallback for this case.

## Fix

Detect whether the target repo's canonical checkout is currently on `main` (reusing the existing `_worktree_topology.canonical_on_main()` helper — `post-pr-merge-pull.py` already runs `git worktree list --porcelain` against `local_path` for the squatter-parking logic, ADR-058) and branch:

- **On `main`:** `git -C <local_path> pull --ff-only origin main` (plain fast-forward; safe, refuses instead of merging on divergence).
- **Not on `main`** (a feature branch, or a squatting worktree holds it — the existing #275 case): keep `git fetch origin main:main` unchanged.

## Acceptance criteria

- [ ] `pull_main()` successfully fast-forwards dev-env's canonical after a dev-env PR merges from a worktree, observed via the `[post-merge-pull] brownm09/dev-env: local main updated` message and a fresh `git rev-parse HEAD` == `origin/main` check.
- [ ] The existing feature-branch-checked-out case (other repos) is unaffected — covered by existing + new offline tests.
- [ ] ADR-warrant check against `docs/adr/INDEX.md` (likely an ADR-058 amendment, given it already documents this exact canonical/main relationship for this file).

brownm09:
## Summary

- `post-pr-merge-pull.py`'s `pull_main()` used `git fetch origin main:main` unconditionally to fast-forward a repo's local `main` after a `gh pr merge`. Git refuses that fetch whenever `main` is checked out *anywhere*, including at the fetch's own target path — which is **always** true for dev-env's own canonical checkout (`C:/Users/brown/Git/dev-env`), since it must stay on `main` per the symlink architecture rule in `CLAUDE.md`. So `pull_main()` was guaranteed to fail on every single dev-env PR merge from a worktree (the only way dev-env PRs ever merge), silently leaving the canonical stale until the next prompt's `dev-env-sync` caught up.
- Fix: detect whether the target repo's canonical is on `main` via the existing `_worktree_topology.canonical_on_main()` helper (ADR-058), fed by the *same* `git worktree list --porcelain` call `park_worktree_off_main` already makes (consolidated into a new `list_worktrees()` — one call per merge event, not two). When on `main`, run a plain `git pull --ff-only origin main`; otherwise the original fetch-into-ref behavior (issue #275) is unchanged.
- Confirmed via code + git behavior (not just a hunch): `git -C <canonical> merge-base HEAD origin/main` showed HEAD was a strict ancestor of `origin/main` (clean fast-forward available, not diverged), and manually running `git fetch origin main:main` against the canonical reproduced the exact `refusing to fetch into branch 'refs/heads/main' checked out` error reported in #488.
- Docs: amended [ADR-058](docs/adr/058-worktree-squatting-main-detection-correction.md) (the ADR that already documents this exact canonical/main relationship for this file) with a dated amendment section, updated the `docs/adr/INDEX.md` date, and updated dev-env's own `CLAUDE.md` Testing item 15 to describe the new coverage.

Closes #488

## Testing

Ran the relevant items from dev-env's `## Testing` section:

1. Hook-script syntax check (`ast.parse` over `claude/scripts/*.py`) — OK.
2. `test_hookio.py` — 17 passed, 0 skipped, 0 failed (unchanged; regression check).
3. `test_post_pr_merge_pull.py` — 9 passed, 0 skipped, 0 failed (7 existing + 2 new `pull_command()` cases: canonical-on-main → ff-only pull, canonical-off-main → fetch-into-ref unchanged).
4. `test_worktree_topology.py` — 14 passed, 0 skipped, 0 failed (unchanged; confirms `canonical_on_main()` — already covered by `test_canonical_on_main` — needed no new coverage).

**Tests: 40 passed, 0 skipped, 0 failed** (across the four suites above).

`pull_main` / `extract_repo` / `list_worktrees` themselves shell out and are intentionally not unit-tested, per this repo's no-subprocess-mock convention — they're exercised live by this PR's own merge (see below).

### Suppression / test-integrity pre-PR greps (CLAUDE.md Code Quality)

- Suppression grep: no matches.
- Test-integrity grep: 2 matches, both `sys.exit(0)` — a **false positive** from the `xit\(` pattern (targets Mocha's `xit(` skip function) substring-matching inside Python's `exit(0)`. These are the pre-existing safe-exit-guard calls in `post-pr-merge-pull.py`'s `__main__` block and `main()`, not test-skip markers; the repo has no JS/TS test framework for this pattern to legitimately fire against (CLAUDE.md's own "Scope note" on this policy already flags JS-only patterns as not meaningfully applicable here).

## Live reproduction note (dev-env#489)

This PR's own merge — from a worktree, canonical on `main` — will hit the identical `gh pr merge` local-cleanup failure shape as PR #476. I'll capture the raw merge output at that point as the empirical test for the companion investigation issue #489 (does the success marker actually survive to the PostToolUse hooks on this failure path), and update #489 with the finding afterward.

## ADR-warrant

Amended [ADR-058](docs/adr/058-worktree-squatting-main-detection-correction.md) in this PR — it already owns the canonical/main topology concern for this exact file (its "Three wirings" section documents `post-pr-merge-pull.py`'s `park_worktree_off_main`), so this is the natural home for the missing third topology case (canonical *on* `main`) rather than a new ADR.

`/review` posted: [PR comment](https://github.com/brownm09/dev-env/pull/493#issuecomment-4859711311) — 0 blocking, 1 non-blocking.

- **[maintainability] Language-scope coverage gap** (automated test-integrity scanners target JS/TS, can't inspect Python skip idioms) — no code action needed; already manually verified and documented above under "Suppression / test-integrity pre-PR greps" (the two `xit(`/`sys.exit(0)` grep hits are confirmed false positives, not skip markers).


brownm09:
## Review: fix(hooks): pull_main() falls back to ff-only pull when canonical is on main

### Summary
Ready to merge. This is a well-scoped, correctly-generalized fix: it detects whether the target repo's canonical checkout is on `main` (reusing the existing, already-tested `canonical_on_main()` helper) and switches between `git pull --ff-only` and the original fetch-into-ref trick accordingly — rather than special-casing dev-env by name. I verified there are no other call sites for the three functions whose signatures changed (`pull_main`, `park_worktree_off_main`, new `list_worktrees`) — all three are only referenced within this same file's `main()`. No blocking findings.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/post-pr-merge-pull.py`, `claude/scripts/tests/test_post_pr_merge_pull.py`
Diff contains Python files, but this review's automated test-integrity scanners (Step 2e) target JS/TS test frameworks (Jest, Vitest, Mocha) and cannot inspect Python skip-marker idioms (`@pytest.mark.skip`, `pytest.skip(`, etc.).
**Suggestion:** Already addressed — the PR body explicitly manually-verifies this: the only two automated-grep hits (`sys.exit(0)` matching the `xit\(` pattern) are the pre-existing safe-exit-guard calls, not skip markers, and the repo has no Python-specific integrity-scan equivalent to extend. No further action needed; noting for the record since the automated gate itself can't confirm this.

---

### Findings Disposition

- **[maintainability] Language-scope coverage gap** — no action needed; already manually verified and documented in the PR body (the two `xit(`/`sys.exit(0)` grep hits are confirmed false positives, not test-skip markers).

---

### Style Notes
Clean. The new code closely follows the file's existing conventions — docstring style, the `except Exception` broad-catch safe-exit pattern used throughout every function in this file, and the `[post-merge-pull] {repo}: ...` stderr message format. `pull_command()`'s docstring clearly states the git-level invariant it works around (`refusing to fetch into branch ... checked out`), which will save the next reader a trip to `git fetch --help` or a repro session.

---

Issue #494, opened 2026-07-01
pr-merge-reminder.py: a chained gh pr create && gh pr merge command can suppress the create reminder


brownm09:
## Problem

`pr-merge-reminder.py::main()` dispatches like this:

```python
if is_merge:
    if not _is_successful_merge_call(output):
        sys.exit(0)
elif exit_code != 0:
    sys.exit(0)

messages = []
if is_create:
    ...
    messages.append(...)
if is_merge:
    ...
    messages.append(...)
```

For a single-purpose command this is fine. But a **chained** command that matches both `is_create` and `is_merge` as separate top-level statements (e.g. `gh pr create --fill && gh pr merge --auto`) runs the merge sub-check against the *combined* output of both commands. If the merge sub-command doesn't complete (a queued `--auto`, or — less realistically — `--help`), `_is_successful_merge_call` returns `False` and the function exits immediately via `sys.exit(0)` — before the `messages` section that would otherwise build the legitimate `is_create` reminder ever runs. The real PR-create event goes unremind ed.

This isn't newly introduced — before dev-env#485/PR #491, the same suppression already happened whenever the chained merge sub-command failed outright (non-zero exit, no marker). PR #491 (fixing #485) widens the trigger to also include exit-0-but-incomplete merges (`--auto`, `--help`), a larger set of chained-command shapes than before.

## Why this is lower priority than #485

- Chaining `gh pr create` with `gh pr merge` in one Bash call is an uncommon pattern (CLAUDE.md's own auto-merge guidance suggests `--auto` is typically invoked as a separate, later call after the full post-PR checklist passes in-session).
- There's a partial backstop: CLAUDE.md's "Incomplete journal detected" check would eventually flag a missing stub if the create reminder is silently skipped.

## Fix

Decouple the three message branches (`is_create`, `is_merge`, `is_push`) so a negative/incomplete result on one doesn't suppress the others — e.g., build each message independently and only skip a given message when ITS OWN condition is unmet, rather than exiting the whole function early inside the `is_merge` branch.

## Context

Surfaced as a non-blocking finding during `/review` of [dev-env#491](https://github.com/brownm09/dev-env/pull/491) (fixing #485).

PR #497, opened 2026-07-01
fix(hooks): add gh-pr-view fallback confirmation when the merge marker is lost

### LINKED ISSUE #489: Investigate: did gh's merge-success marker actually reach the PostToolUse hooks during the PR #476 worktree-merge exit-1?

brownm09:
## Summary

- `gh` prints its completed-merge success marker (`Squashed and merged pull request #N`) before attempting the local branch checkout/delete that can fail with `main is already checked out` — confirmed by reading gh's own source (`pkg/cmd/pr/merge/merge.go`, `mergeRun`: the `infof()` success print in `merge()` happens before `deleteLocalBranch()` runs). But that marker does not reliably survive to a PostToolUse hook's captured `stdout`/`stderr` when gh exits abruptly right after that failing git subprocess.
- **Confirmed via live reproduction** (not just a hunch): merging dev-env PR #493 hit the identical failure. `gh pr view` confirmed the remote merge genuinely succeeded, but **none** of the six marker-gated hooks fired — concretely verified (not just "no message seen"): the canonical `C:/Users/brown/Git/dev-env`'s `HEAD` stayed at the exact same commit it was at before the merge, proving `post-pr-merge-pull.py`'s fast-forward never ran. A control test (`echo "MARKER"; exit 1`) proved stdout generally survives a non-zero exit in this harness, ruling out a blanket "Bash tool drops stdout on failure" explanation — the loss is specific to `gh pr merge`'s output on this exact failure path (most likely gh, a Go CLI writing to a non-TTY pipe, buffering stdout and exiting abruptly before flushing it).
- Fix: two new `_hookio.py` functions — `should_confirm_via_gh(exit_code, output)` (pure: only worth a live check when the marker is absent AND the exit code is non-zero, so the common non-merge / clean-exit paths never pay a network call) and `confirm_merge_via_gh(pr_number, repo, cwd)` (shells out to `gh pr view`, returns the confirmed PR number when `state == "MERGED"`).
- Wired into `post-pr-merge-project.py` only (the board-Done move) — the one of the six marker-gated hooks with **no other eventual-consistency backstop** for a missed action (`post-pr-merge-pull.py` has `dev-env-sync` catching up next-prompt, ADR-058; `post-pr-merge-reclaim.py` has the 6-hourly reclaim routine, ADR-037; the rest are reminders/reports, not persistent state). Rolling out to the other five is tracked separately: https://github.com/brownm09/dev-env/issues/494 — avoids up to 6 redundant `gh pr view` calls for a single failed-merge event, since each hook is an independent PostToolUse process with no natural place to share one result.
- Docs: amended [ADR-050](docs/adr/050-shared-hookio-sibling-hook-fixes.md) (Amendment 3 — the ADR that already owns this hook family's shared marker-detection logic) with the full incident writeup, updated `docs/adr/INDEX.md`, and updated dev-env's own `CLAUDE.md` Testing items 13/14 to describe the new coverage.

Closes #489

## Testing

Ran the relevant items from dev-env's `## Testing` section:

1. Hook-script syntax check (`ast.parse` over `claude/scripts/*.py`) — OK.
2. `test_hookio.py` — 21 passed, 0 skipped, 0 failed (17 existing + 4 new `should_confirm_via_gh` cases: lost-marker case confirms, marker-present skips the network call, clean exit skips it, default/missing exit code still attempts).
3. `test_post_pr_merge_project.py` — 16 passed, 0 skipped, 0 failed (unchanged — `merge_succeeded`/`extract_pr_number*` bodies weren't touched, only how `main()` calls them).
4. Regression check on every other hook that imports `_hookio` (since it gained new imports): `test_post_pr_merge_pull.py` (9), `test_post_pr_merge_reclaim.py` (4), `test_pr_merge_reminder.py` (31), `test_post_merge_tile_checkpoint.py` (4), `test_usage_snapshot.py` (11) — all pass, 0 failed.

**Tests: 96 passed, 0 skipped, 0 failed** (across all suites above).

`confirm_merge_via_gh` itself shells out to `gh pr view` and is intentionally not unit-tested, per this repo's no-subprocess-mock convention.

### Suppression / test-integrity pre-PR greps (CLAUDE.md Code Quality)

- Suppression grep: no matches.
- Test-integrity grep: matches are the same `xit\(`-vs-Python-`exit(` false positive as PR #493 — this time also catching a test function named `..._on_clean_exit()` (`clean_exit()` contains `xit(`) in addition to the pre-existing `sys.exit(0)` safe-exit-guard calls. Not test-skip markers; the repo has no JS/TS test framework for this pattern to legitimately fire against.

## ADR-warrant

Amended [ADR-050](docs/adr/050-shared-hookio-sibling-hook-fixes.md) (Amendment 3) — it already owns the shared `_hookio` marker-detection logic and already has two prior 2026-07-01 amendments for closely related incidents in this same hook family, so this is the natural continuation rather than a new ADR.

`/review` posted: [PR comment](https://github.com/brownm09/dev-env/pull/497#issuecomment-4859921047) — 0 blocking, 1 non-blocking.

- **[maintainability] Language-scope coverage gap** (automated test-integrity scanners target JS/TS, can't inspect Python skip idioms) — no code action needed; already manually verified and documented above under "Suppression / test-integrity pre-PR greps" (the grep hits are confirmed false positives, not skip markers).


brownm09:
## Review: fix(hooks): add gh-pr-view fallback confirmation when the merge marker is lost

### Summary
Ready to merge. The fallback is correctly gated (no network call on the common non-merge / clean-exit paths — verified via `should_confirm_via_gh`'s tests and its own reasoning: a genuine merge always prints the marker per Amendment 2's finding, so exit-0-with-no-marker is unambiguously a non-merge), there's no command-injection surface (`subprocess.run` takes a list, never `shell=True`), and I verified via `git grep` on the remote branch that both new `_hookio` functions have exactly one call site each, correctly wired. No blocking findings.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/_hookio.py`, `claude/scripts/post-pr-merge-project.py`, two test files
Diff contains Python files; this review's automated test-integrity scanners (Step 2e) target JS/TS test frameworks and cannot inspect Python skip-marker idioms.
**Suggestion:** Already addressed — the PR body manually verifies the automated grep's hits are the same `xit\(`-vs-Python-`exit(` false positive as PR #493 (this time also matching a test function named `..._on_clean_exit()`), not skip markers. No further action needed; noting for the record since the automated gate can't confirm this itself.

---

### Findings Disposition

- **[maintainability] Language-scope coverage gap** — no action needed; already manually verified and documented in the PR body under "Suppression / test-integrity pre-PR greps."

---

### Style Notes
Clean. `confirm_merge_via_gh`'s docstring is thorough about *why* the fallback exists (the gh-source ordering, the live-reproduction incident, the specific failure mechanism hypothesis) — useful context for whoever eventually works on the #494 follow-up (rolling this out to the other five hooks) without re-deriving the rationale from scratch.

---

Issue #499, opened 2026-07-01
post-tool-use.py: gh issue/pr create detection uses unanchored substring match, false-positives on heredoc/commit-message/grep text


brownm09:
## Problem

\`post-tool-use.py::main()\` detects a \`gh issue create\` / \`gh pr create\` invocation with a raw, unanchored regex search over the *entire* Bash command string:

\`\`\`python
is_issue_create = bool(re.search(r"\bgh\s+issue\s+create\b", command))
is_pr_create = bool(re.search(r"\bgh\s+pr\s+create\b", command))
\`\`\`

This matches the substring **anywhere** in \`command\` — including inside a heredoc body, a quoted commit message, a grep pattern argument, etc. — not just an actual top-level invocation.

## Repro (from the dev-env#494 fix session, 2026-07-01)

1. \`git commit -m "$(cat <<'EOF' ... )"\` where the heredoc body of the commit message quoted an example command \`gh pr create --fill && gh pr merge --auto\`. The hook fired: \`[project-hook] PR created but no GitHub URL found in output.\`
2. A \`grep -n "gh pr create\|gh issue create\|..." claude/scripts/post-tool-use.py\` command whose *pattern argument* contained those literal strings. The hook fired again: \`[project-hook] Issue created but no GitHub URL found in output.\`

Neither command was a real \`gh issue create\` / \`gh pr create\` invocation. No project mutation happened in either case (no GitHub URL was found in the unrelated output, so \`add_to_project\` was never reached) — impact was a misleading stderr advisory, not silent corruption. But a false-positive command whose output *does* happen to contain an unrelated GitHub URL (e.g. a commit message mentioning a different PR's URL) could add the wrong item to the project board.

## Fix

\`pr-merge-reminder.py\` already solves exactly this class of problem: \`_scan_top_level()\` (a stack-based parser tracking single/double quotes, \`$()\` subshells, and heredoc bodies via \`_find_heredoc_end()\`) plus \`is_pr_create_command()\` / \`is_pr_merge_command()\` / \`is_git_push_command()\` only match a *top-level* statement, never a buried substring. \`pre-tool-use-canonical-mutate-guard.py\` has equivalent protection (the career-playbook #442 lesson, per its own test suite).

[...]

PR #500, opened 2026-07-01
fix(hooks): decouple pr-merge-reminder message branches (dev-env#494)

### LINKED ISSUE #494: pr-merge-reminder.py: a chained gh pr create && gh pr merge command can suppress the create reminder

brownm09:
## Problem

`pr-merge-reminder.py::main()` used the `is_merge` success-marker check as a shared gate for the *entire* function:

```python
if is_merge:
    if not _is_successful_merge_call(output):
        sys.exit(0)
elif exit_code != 0:
    sys.exit(0)
```

For a chained command matching both `is_create` and `is_merge` as separate top-level statements (e.g. `gh pr create --fill && gh pr merge --auto`), an incomplete merge sub-check (a queued `--auto`, or `--help` — no success marker) exited the function before the `messages` section ever built the otherwise-legitimate `is_create` reminder. The real PR-create event went unreminded.

## Fix

Extracted the dispatch into `_build_messages()`, gating each message on its own condition instead of one shared early exit:

- `is_create` message: requires `exit_ok`
- `is_merge` message: requires the success marker (unchanged)
- `is_push` message: requires `exit_ok` and `not(is_create or is_merge)` (unchanged)

`exit_ok = is_merge or exit_code == 0` — this preserves the pre-existing quirk that `exit_code` was never consulted for create/push when `is_merge` was also true (so a chained create+merge where the merge genuinely succeeds still fires both messages regardless of the chain's aggregate exit code, exactly as before). The change reproduces the original truth table exactly except for the one buggy case: an incomplete/failed merge sub-check no longer suppresses an independently-successful create.

This also makes the dispatch logic directly unit-testable without driving `main()`'s stdin, consistent with this file's existing pure-helper testing convention.

## Testing

Ran `py -3 claude/scripts/tests/test_pr_merge_reminder.py` — 40 passed, 0 skipped, 0 failed (11 new cases covering the dev-env#494 repro: chained create + queued `--auto`, chained create + `--help`-shaped merge, chained create + successful merge, and regression coverage for the unchanged single-purpose and worktree-nonzero-exit-with-marker (#275) cases).

Also ran the hook-script syntax check (`py -3 -c "import ast,sys; ..." claude/scripts/*.py`) — clean.

Pre-PR suppression grep: clean (no matches). Pre-PR test-integrity grep: the only matches are `sys.exit(0)` calls flagged by the `xit\(` pattern matching the substring "e**xit(**" — a known JS-oriented-pattern false positive in this Python repo (not a Jest skip marker), per the CLAUDE.md scope note. No deleted test files.

## Out of scope

While testing this fix, a commit message and a grep pattern (both containing the literal example text `gh pr create`/`gh issue create`) each spuriously triggered `post-tool-use.py`'s project-board-add hook — that hook does unanchored substring matching over the raw command text rather than the heredoc/quote-aware top-level-statement parsing this file already has. Filed separately as [dev-env#499](https://github.com/brownm09/dev-env/issues/499); not fixed here (different file, different hook, non-trivial scope on its own).

Closes #494

- **[correctness]** `exit_ok = is_merge or exit_code == 0` false-positives when create fails and merge never runs (chain short-circuit) — fixed in `0dd3ef9` (replaced with `create_push_ok = exit_code == 0 or merge_ok`; added 2 tests locking in the create-failure and #275-chained-worktree-merge cases).


brownm09:
## Review: fix(hooks): decouple pr-merge-reminder message branches (dev-env#494)

### Summary
Needs one revision before merge. The extraction into `_build_messages()` and the test coverage are solid, but the `exit_ok = is_merge or exit_code == 0` gate introduces a new false-positive that the pre-fix code (accidentally) avoided: a chained `gh pr create --fill && gh pr merge --auto` where **create itself fails** will now fire a "PR created" reminder even though no PR exists. This is the same bug class the PR sets out to fix, just in the opposite direction, so it should be fixed in this PR rather than filed separately.

---

### Blocking Findings

**[correctness]** `claude/scripts/pr-merge-reminder.py:361` (`_build_messages`, the `exit_ok` line)

`exit_ok = is_merge or exit_code == 0` treats `is_merge` as sufficient on its own to waive the exit-code check for the create message — but `is_merge` is a static regex match on the *command text*, not evidence that a merge actually ran. For `gh pr create --fill && gh pr merge --auto`, bash's `&&` short-circuits on create's failure: `gh pr merge` never executes, `is_merge` is still `True` (the text is present), and `exit_code` correctly reflects create's own non-zero failure. With the current gate, `exit_ok` evaluates to `True` regardless, so `is_create and exit_ok` fires the "gh pr create detected — write the journal stub" reminder for a PR that was never created. The pre-fix code avoided this by accident (the shared `if is_merge: if not marker: exit(0)` gate killed the whole function whenever there was no merge marker, including this case) — so this PR trades the reported false-negative (#494) for a new false-positive, on a command shape the PR's own description discusses.

**Fix:** Replace the gate with `exit_code == 0 or merge_ok` (computing `merge_ok = is_merge and _is_successful_merge_call(output)` once, reused by both the merge branch and this gate). A confirmed merge is independent proof that create must have already succeeded (merge cannot complete against a PR that was never opened), so it's valid evidence for the create message even when the chain's aggregate exit code is non-zero (the `#275` worktree case, chained with a preceding create). This fixes the reported #494 bug (exit_code is 0 in both the queued-`--auto` and `--help` scenarios), fixes the create-fails/merge-never-ran false positive above, and additionally handles the chained-worktree-merge-success case correctly (unlike a plain `exit_code == 0` gate, which would have to sacrifice that case). Add a test locking in the create-fails/merge-never-ran → no message behavior, and one for the chained create + worktree-merge-success (non-zero exit, marker present) → both messages fire.

---

### Findings Disposition

Per ADR-028/ADR-039, the disposition (fixed in `<sha>` or filed `#N`) for the finding above must be recorded in a "Review findings disposition" section of the PR body before merge.

---

PR #502, opened 2026-07-01
fix(claude-md): anchor xit\( test-integrity grep on a word boundary

brownm09:
## Summary

- Rule 3's pre-PR test-integrity grep (`claude/CLAUDE.md` → Code Quality → Test integrity policy) used bare `xit\(` to catch Mocha's `xit(` skip function. Because the check is a plain substring match, `xit\(` also matches inside any identifier ending in `...exit(` — Python's `exit(`/`sys.exit(`/`os._exit(` calls (this repo's standard hook-script safe-exit-guard idiom) and any `_exit(`-suffixed function name. Confirmed twice today: [#493](https://github.com/brownm09/dev-env/pull/493) (`sys.exit(0)`) and [#497](https://github.com/brownm09/dev-env/pull/497) (`sys.exit(0)` + `..._on_clean_exit()`).
- Fix: anchor the pattern on a word boundary, `\bxit\(` instead of `xit\(`, using GNU grep's `\b` extension (available in `-E` mode). There is no word boundary between `e` and `x` in `exit(`/`_exit(` (both word characters), so `\bxit\(` no longer matches Python's exit calls, while a real Mocha `xit(` — always preceded by whitespace, a quote, a paren, or line start — still does.
- Docs: amended [ADR-029](docs/adr/029-test-integrity-policy.md) with a dated `## Amendment (2026-07-01)` section documenting the false-positive history and empirical verification (leaving the original Decision code block untouched as the historical record, matching today's ADR-003/050/058 amendment pattern), and updated `docs/adr/INDEX.md`'s ADR-029 row date.

Closes #501

## Testing

This is a docs-only change to `claude/CLAUDE.md` + `docs/adr/`. No hook script (`.py`) was touched, so none of dev-env's `## Testing` items 1–33 apply beyond the docs-only guard. No automated suite targets this policy's grep pattern (confirmed — none of the 33 items exercise it; it's a manually-run documented command, not a script).

1. **Docs-only guard** (required for `claude/CLAUDE.md` changes): `grep -n 'date -u' claude/CLAUDE.md` → 0 matches. Vacuously clean.
2. **Manual empirical verification** of the fix:
   ```
   $ printf 'sys.exit(0)\nos._exit(1)\ndef clean_exit():\n  xit(%s\n' "'foo', () => {})" | grep -E '\bxit\('
     xit('foo', () => {})
   ```
   `sys.exit(0)`, `os._exit(1)`, and `clean_exit():` no longer match; a real `xit(` call still does.

[...]

PR #506, opened 2026-07-02
fix(hooks): default subprocess text-mode output to UTF-8 on Windows

### LINKED ISSUE #503: post-tool-use.py crashes with UnicodeDecodeError (cp1252) reading subprocess output
## Problem

`post-tool-use.py`'s auto-add-to-project step crashed with a `UnicodeDecodeError` immediately after `gh pr create` succeeded for PR #502, instead of silently adding the PR to project #3 as it normally does.

## Traceback (verbatim from the PostToolUse hook error)

```
Exception in thread Thread-1 (_readerthread):
Traceback (most recent call last):
  File "C:\Users\brown\AppData\Local\Programs\Python\Python312\Lib\threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "C:\Users\brown\AppData\Local\Programs\Python\Python312\Lib\threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\brown\AppData\Local\Programs\Python\Python312\Lib\subprocess.py", line 1599, in _readerthread
    buffer.append(fh.read())
                  ^^^^^^^^^
  File "C:\Users\brown\AppData\Local\Programs\Python\Python312\Lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 3055: character maps to <undefined>
```

The hook printed a fallback: `PR created but auto-add to project failed`, with the manual `gh project item-add` command as a recovery step. PR #502 itself was created fine; only the project-board automation failed.

## Likely cause

`post-tool-use.py` shells out to a subprocess (most likely `gh project item-add` or a `gh pr view`/`gh pr create` re-read) without pinning the child process's stdout encoding to UTF-8. On Windows, Python's `subprocess` defaults a text-mode pipe to the OS ANSI codepage (cp1252) unless `encoding="utf-8"` (or `errors="replace"`) is passed explicitly. Byte `0x9d` is unassigned in cp1252, so any UTF-8 multi-byte sequence landing on that byte value in the child process's output (e.g. certain Unicode punctuation echoed back from a PR title/body) crashes the reader thread.

This is the same class of Windows console/pipe encoding issue already documented elsewhere in this repo for hook stdout (see `posttooluse-inert-advisory.py`'s "ASCII/cp1252-encodable" constraint in dev-env's own `## Testing` item 18), but here it is the hook's *own outbound* subprocess read that is unguarded, not inbound hook stdout.

## Impact

Medium - recurs periodically (whenever a PR/issue title or body created in the same session contains a character outside cp1252, which is common with em-dashes, curly quotes, or other Unicode punctuation) and silently degrades correctness: the project board silently fails to gain the new item unless the operator notices the PostToolUse error and runs the manual fallback.

## Suggested fix direction

Find the `subprocess.run(...)` / `subprocess.Popen(...)` call(s) in `claude/scripts/post-tool-use.py` that read a child process's stdout/stderr as text, and pass `encoding="utf-8", errors="replace"` (or decode raw bytes explicitly as UTF-8) instead of relying on the platform default. Add a regression test consistent with this repo's existing pure-helper test convention for this file (`claude/scripts/tests/test_post_tool_use.py`, dev-env `## Testing` item 12).

brownm09:
## Summary

- `post-tool-use.py`'s `add_to_project()` crashed with `UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d` reading `gh project item-add`'s stdout, right after a successful `gh pr create` for PR #502. Root cause: `subprocess.run(..., text=True)` with no `encoding=` decodes as `locale.getpreferredencoding(False)` — cp1252 on this machine — which cannot represent every UTF-8 byte `gh`/`git` can emit (e.g. em-dashes/curly quotes echoed back from a PR/issue title). The identical gap existed in `canonical_root_via_git()` (same file) and `_hookio.py`'s `confirm_merge_via_gh()`.
- Fixed by extending `_winsubp.py`'s existing `Popen.__init__` monkey-patch — the same seam [ADR-007](docs/adr/007-hook-command-invocation.md)'s 2026-06-01 follow-up already uses to suppress console flashes — to also default `encoding="utf-8", errors="replace"` for text-mode calls that don't already specify one. This is a deliberate deviation from the issue's literal "fix the 3 call sites" suggestion: since every subprocess-using script in `claude/scripts/` already mandatorily imports `_winsubp` (enforced by `test_pyw_stdio.py`), one change here fixes the reported crash *and* the identical latent gap in ~20 other scripts, instead of touching each individually. Full rationale is recorded as a new "Decision History" entry in ADR-007.
- Also added the (previously missing) `import _winsubp` directly to `_hookio.py` itself — its one subprocess call was only protected because every current caller happens to import `_winsubp` first; this closes that latent fragility regardless of future caller diligence.

## Testing

- `claude/scripts/tests/test_winsubp.py` — pure offline tests of the extracted `_apply_windows_subprocess_defaults(kwargs)` helper (no subprocess spawn), matching this repo's pure-helper test convention: `Tests: 12 passed, 0 skipped, 0 failed`
- `claude/scripts/tests/test_pyw_stdio.py` gained a 7th check that reproduces the *exact* dev-env#503 crash end-to-end under a real `pyw -3` child (writes the exact byte `0x9d` from the traceback, asserts it decodes to U+FFFD instead of being silently lost — see the test's docstring for why the crash doesn't surface as a catchable exception): `Tests: 7 passed, 0 skipped, 0 failed`
- `claude/scripts/tests/test_post_tool_use.py`: `Tests: 22 passed, 0 skipped, 0 failed`
- `claude/scripts/tests/test_hookio.py`: `Tests: 21 passed, 0 skipped, 0 failed`
- Also re-ran the two other test files that spawn/mock real subprocesses through the `_winsubp`-patched path (highest regression risk from this change): `test_canonical_mutate_guard.py` (28 passed) and `test_prune_merged_worktrees.py` (1 passed) — both clean, no behavior change for existing callers since the defaults are additive-only (previously-unset `encoding=` gets set; nothing already working is altered).
- Hook-script syntax check (`## Testing` item 1): clean.
- Suppression grep (ADR-026) and test-integrity grep (ADR-029): both empty — no suppressions, no skip markers, no deleted tests. (JS-scoped patterns; this is a Python-only change, noted per the policy's scope note.)
- New `## Testing` item 34 added to this repo's `CLAUDE.md` documenting `test_winsubp.py`.

## ADR

[ADR-007](docs/adr/007-hook-command-invocation.md) gained a new "2026-07-02 (follow-up 3)" Decision History entry plus updated Consequences, since this changes the behavior of a shared, mandatorily-imported module. `docs/adr/INDEX.md` row updated (amended date + `subprocess`/`encoding`/`unicode`/`windows` tags for future discoverability).

`/review` pass ([comment](https://github.com/brownm09/dev-env/pull/506#issuecomment-4862936579)) found 1 blocking + 2 non-blocking findings, all fixed in `492a269`:

- **[documentation] (blocking)** — README.md/docs/REFERENCE.md's 3 existing `_winsubp.py` passages described only console-flash suppression, not the new UTF-8 decoding default. Fixed in `492a269`.
- **[correctness] (non-blocking)** — `wants_text` didn't account for `errors=` alone implying text mode (confirmed empirically against real `Popen`). No current call site affected; widened the check and added a regression test anyway. Fixed in `492a269`.
- **[reliability] (non-blocking)** — the new crash-reproduction test's own outer `pyw` capture didn't specify `encoding=`. Fixed in `492a269`.

All tests re-verified green after fixes.

Closes #503


brownm09:
## Review: fix(hooks): default subprocess text-mode output to UTF-8 on Windows

### Summary
Solid fix with strong regression coverage (an offline pure-function suite plus a live end-to-end reproduction of the exact reported crash byte) and a well-reasoned architectural deviation from the issue's literal suggestion. One blocking documentation gap and two narrow non-blocking correctness/robustness gaps found on close inspection — none affect the reported crash's fix, all cheap to close in this PR.

---

### Blocking Findings

**[documentation]** `README.md:47`, `docs/REFERENCE.md:136`, `docs/REFERENCE.md:316`
All three passages describe `_winsubp.py` as only patching `subprocess.Popen.__init__` for `CREATE_NO_WINDOW` (console-flash suppression). This PR expands the module's purpose to also default UTF-8 decoding for text-mode calls — the module's own docstring was updated thoroughly, but these three reference-doc passages (dev-env's own "quick reference" docs per `CLAUDE.md`) were not, so a reader consulting them comes away with a stale, incomplete picture of what `_winsubp.py` does.
**Fix:** Update all three passages to mention the UTF-8 defaulting behavior alongside the console-flash suppression.

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/_winsubp.py` — `_apply_windows_subprocess_defaults`
`wants_text = kwargs.get("text") or kwargs.get("universal_newlines")` doesn't account for `errors=` alone implying text mode. Confirmed empirically: `subprocess.Popen(cmd, stdout=PIPE, errors="replace")` (no `text=`, no `encoding=`) returns `str`, not `bytes` — CPython enables text mode whenever `encoding`, `errors`, `text`, or `universal_newlines` is truthy. A hypothetical future caller that passes `errors=` alone (no `text=`) would still fall through to CPython's own `locale.getpreferredencoding(False)` default for `encoding`, i.e. still hit cp1252. No current call site in this repo does this, so it doesn't affect the reported crash.
**Suggestion:** Add `kwargs.get("errors")` to the `wants_text` check so this combination is covered too, and pin it with a test.

**[reliability]** `claude/scripts/tests/test_pyw_stdio.py` — `test_winsubp_decodes_utf8_not_cp1252_under_pyw`
The new test's *outer* `subprocess.run(["pyw", "-3", "-c", program], capture_output=True, text=True)` call — which captures the middle child's JSON status and stderr — doesn't itself specify `encoding=`. Currently safe in practice (the JSON status is ASCII-only via `json.dumps`'s default `ensure_ascii=True`, and Python tracebacks are ASCII), but it's a little ironic for a test whose entire purpose is proving this exact class of bug is fixed to not dogfood the fix on its own outer capture.
**Suggestion:** Add `encoding="utf-8", errors="replace"` to the outer capture for consistency/robustness, even though it's not currently exercised.

---

### Findings Disposition
_To be filled in after fixes land in this PR._

---

PR #507, opened 2026-07-02
Dedupe worktree-canon + gh add_to_project between post-tool-use.py and reconcile-project-board.py

### LINKED ISSUE #454: [refactor] Share worktree-canonicalization + add_to_project between post-tool-use.py and reconcile-project-board.py
## Problem

`claude/scripts/reconcile-project-board.py` (added in #451 / ADR-068) duplicates two pieces of logic that already exist in `claude/scripts/post-tool-use.py`, rather than importing them from a shared module:

1. **Worktree canonicalization** — `post-tool-use.py`'s `_WORKTREE_RE` + `canonical_root_from_worktree()` vs. `reconcile-project-board.py`'s near-identical `_WORKTREE_RE` + `canonical_repo_root()`. The regex is copy-pasted verbatim; the functions differ subtly in contract — `canonical_root_from_worktree()` returns `None` on no-match, `canonical_repo_root()` passes the input through unchanged.
2. **`add_to_project()`** — both scripts have their own `gh project item-add` wrapper that differs subtly in exception-handling scope (one wraps only the `subprocess.run` call, the other wraps the whole function body).

## Why this matters

ADR-068's own rationale for reading a *shared config file* (`hook-config.json`) is explicitly anti-drift — "the project number/owner/node-id and the required-field IDs never drift between the add-hook and this reconciler." But the canonicalization regex and the add-to-project logic, which must **also** stay in lockstep for the two scripts to agree on where the config lives and how items get added, are duplicated rather than shared — undercutting that same anti-drift argument.

This is the same class of problem the repo has already solved elsewhere by extracting a shared `_`-prefixed module: `_hookio.py`, `_journal_shards.py`, `_worktree_topology.py`, `_hookutil.py` all exist specifically to prevent sibling scripts from silently diverging on shared logic.

## Proposed fix

Extract the worktree-canonicalization regex + helper into a shared module (either a new `_worktree_canon.py` or an addition to the existing `_worktree_topology.py`, which already owns worktree-topology logic). Reconcile the two contracts (None-on-no-match vs. pass-through-unchanged) explicitly rather than silently picking one. Share one `add_to_project()` implementation between the two call sites.

## Scope note

Deliberately **not** bundled into #451 — fixing this means touching a second, already-shipped, already-tested hook (`post-tool-use.py`) and its existing test (`claude/scripts/tests/test_post_tool_use.py`), which is a larger and riskier change than adding a new reconciler script. Flagged as a `[maintainability]` finding during #451's review; filed per the fix-or-file review-findings policy rather than fixed inline.

## References

- PR: https://github.com/brownm09/dev-env/pull/451
- ADR-068 (not yet merged as of filing — see #451; renumbered from ADR-067 during rebase due to a numbering collision with #448)

brownm09:
## Summary

Closes #454. `reconcile-project-board.py` (PR #451, ADR-068) duplicated two pieces of logic that already existed in `post-tool-use.py`:

1. **Worktree canonicalization** — an identical `_WORKTREE_RE` regex, wrapped by two functions with different no-match contracts (`None` vs. passthrough). Extracted to new `claude/scripts/_worktree_canon.py`, which exposes both contracts as thin wrappers over one shared match rather than picking one and breaking a caller.
2. **`add_to_project()`** — two independent `gh project item-add` subprocess wrappers with different signatures/return shapes/timeouts/encoding. Extracted to new `claude/scripts/_gh_project.py`, reconciled onto the superset `(item_id, stderr)` shape.

Both `post-tool-use.py` and `reconcile-project-board.py` now import and delegate; local definitions removed. Full rationale, the explicit contract reconciliation, and why the two pieces weren't bundled into one module: [ADR-073](docs/adr/073-shared-worktree-canon-gh-project-modules.md).

**Note on the branch:** this was planned against an earlier `reconcile-project-board.py`. By the time the branch was cut, `--scan-dir` (ADR-070) and a new shared `_repo_scan.py` (ADR-072) had both merged concurrently, consuming ADR numbers 070–072 and moving `add_to_project`'s call site into a new `_reconcile_repo()` helper. The two functions extracted here are byte-identical to what was originally read — only surrounding context shifted, and this PR (ADR-073) was re-verified against the current file before implementing.

## Behavior-equivalence notes (both intentional, called out per repo convention rather than left implicit)

- **`post-tool-use.py`'s `add_to_project` call now decodes with `encoding="utf-8"`** (previously the OS default locale — cp1252 on Windows). This is a deliberate correctness fix: `gh` emits UTF-8 JSON, and a non-ASCII issue title could previously raise `UnicodeDecodeError`, silently swallowed by the old bare `except Exception: return None`.
- **`config["project_number"]`/`config["project_owner"]` dict access moved from inside `add_to_project`'s own try/except to `main()`'s own body**, since the shared function now takes explicit args instead of the whole config dict. A malformed config's `KeyError` now propagates to the module-level `if __name__ == "__main__": try: main() except Exception: sys.exit(0)` wrapper instead of the function's own catch — same ultimate outcome (silent `sys.exit(0)`), so behavior-equivalent, not behavior-identical. **Independently re-verified during `/review`** (see disposition below): the OLD code's `else:` fallback branch already did an unguarded `config['project_number']` subscript in its own f-string, so this was already the outcome for a malformed config before this PR — the change only moves where the latent `KeyError` fires.

No other behavior changes — `timeout` is preserved exactly for each call site (20 for post-tool-use.py via the new default, 30 for reconcile-project-board.py via an explicit keyword).

## Testing

Ran dev-env's `## Testing` items #1 (hook-script syntax check — covers the two new modules via its `claude/scripts/*.py` glob), #12 (`test_post_tool_use.py`), #29 (`test_reconcile_project_board.py`), and the new #35 (`test_worktree_canon.py`, added in this PR):

```
Tests: 22 passed, 0 skipped, 0 failed   (test_post_tool_use.py — unchanged, no edits needed)
Tests: 26 passed, 0 skipped, 0 failed   (test_reconcile_project_board.py — unchanged, no edits needed)
Tests: 8 passed, 0 skipped, 0 failed    (test_worktree_canon.py — new)
```

56/56 passed, 0 skipped, 0 failed. Neither existing test file needed edits: both access the moved functions via `module_attr` on an `exec_module`-loaded copy of the hyphenated script, and `from X import Y` binds the same function object into the importing module's namespace — confirmed by running both unchanged.

Also ran the pre-PR suppression grep, test-integrity grep, and deleted-test-file check — all clean (no matches).

## Documentation

- `docs/REFERENCE.md`: new paragraph (mirroring `_repo_scan.py`'s) noting both modules as non-invoked shared library modules; updated the `post-tool-use.py` PostToolUse-hooks row to cite them.
- `CLAUDE.md`: new `## Testing` item #35 for `test_worktree_canon.py`.
- `docs/adr/073-shared-worktree-canon-gh-project-modules.md` + `docs/adr/INDEX.md` row.

`/review` complete (2 parallel subagents: correctness/security — 0 findings; reliability/maintainability — 2 non-blocking findings). Full report: https://github.com/brownm09/dev-env/pull/507#issuecomment-4863135706

- **[maintainability]** Two more un-consolidated `_WORKTREE_RE` copies exist in `pre-tool-use-worktree-path-check.py` and `pre-tool-use-canonical-mutate-guard.py` — filed [#510](https://github.com/brownm09/dev-env/issues/510) (Impact: Low). Out of scope for this PR (different `PreToolUse` hook family, own ADR-024/ADR-071 test suites).
- **[maintainability]** `_gh_project.py`'s `str(project_number)` coercion was undocumented in the reconciliation docstring — fixed in `d14c1a2`.

Correctness/security subagent also independently verified (deep trace, not just review) that the `config["project_number"]`/`config["project_owner"]` dict-access change at the `post-tool-use.py` call site produces byte-identical observable behavior to the old code on a malformed config (both silently exit 0 via the pre-existing top-level catch-all) — not a regression.


brownm09:
## Review: Dedupe worktree-canon + gh add_to_project between post-tool-use.py and reconcile-project-board.py

### Summary
Ready to merge. This is a clean, behavior-preserving extraction: the two duplicated pieces of logic (`_WORKTREE_RE` canonicalization and `add_to_project`) are consolidated into two new shared modules while explicitly preserving each caller's own divergent contract rather than silently picking one and breaking the other. All 56 tests pass (26 existing + 8 new, all unchanged/added), no correctness or security issues were found (including a deep trace confirming the `config["project_number"]`/`config["project_owner"]` access at the new `post-tool-use.py` call site produces byte-for-byte identical observable behavior to the old code on a malformed config — both silently exit 0), and the diff is appropriately scoped to the two files issue #454 named. Two small non-blocking maintainability items surfaced below, both now disposed.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/_worktree_canon.py` (new module)
Two more copies of essentially the same worktree-path regex exist outside this PR's scope: `pre-tool-use-worktree-path-check.py` (byte-identical anchored-capture regex) and `pre-tool-use-canonical-mutate-guard.py` (a `.search`-variant). The new module's own docstring and ADR-073 frame `_WORKTREE_RE` as "the shared regex," and the comment this PR deleted from `post-tool-use.py` explicitly credited `pre-tool-use-worktree-path-check.py` as the pattern's origin — so the natural home for those two remaining copies is now this module, and a future path-convention change still has to be applied in 3 places, not the 1 the new module implies.
**Suggestion:** Migrate both onto `_worktree_canon.py` in a follow-up (the mutate-guard likely needs a new `is_worktree_path()`-shaped export rather than reusing `canonical_root_from_worktree` directly, since it wants a boolean check, not a captured root). Both files carry their own ADR-024/ADR-071 test suites that a migration must keep green.

**[maintainability]** `claude/scripts/_gh_project.py:52`
The shared `add_to_project` wraps `project_number` in `str()` before use — `reconcile-project-board.py`'s original already did this; `post-tool-use.py`'s original passed the config value straight through, unwrapped. A no-op today (the JSON config schema always stores it as a string), but the module docstring's bullet list otherwise meticulously documents every reconciled delta (return shape, encoding, timeout) and had omitted this one.
**Suggestion:** Documented — see disposition below.

---

### Findings Disposition

- **`_WORKTREE_RE` copies in `pre-tool-use-worktree-path-check.py` / `pre-tool-use-canonical-mutate-guard.py`** — filed [dev-env#510](https://github.com/brownm09/dev-env/issues/510) (Impact: Low). Genuinely out of scope for this PR: different hook family (`PreToolUse` path/mutate guards vs. the `PostToolUse` project-board pair), each with its own ADR and test suite; pulling them in here would have expanded #454's stated scope well past the project-board pair it named.
- **Undocumented `str(project_number)` coercion** — fixed in `_gh_project.py`'s docstring in this PR (added a bullet documenting the normalization and that it's carried over from reconcile-project-board.py's original contract).

---

*Reviewed by `claude-opus-4-8` (2 parallel subagents: correctness/security, reliability/maintainability) via Claude Code*

PR #508, opened 2026-07-02
fix(hooks): anchor gh issue/pr create detection to top-level statements

### LINKED ISSUE #499: post-tool-use.py: gh issue/pr create detection uses unanchored substring match, false-positives on heredoc/commit-message/grep text

brownm09:
## Summary

- `post-tool-use.py` (the project-board-add hook) detected `gh issue create` / `gh pr create` with an **unanchored** `re.search` over the entire raw Bash command string — matching the pattern anywhere, including inside a heredoc body, a quoted commit message, a grep pattern argument, or a `--text` field value.
- Extracts `pr-merge-reminder.py`'s existing stack-based statement parser (`scan_top_level` / `_find_heredoc_end`) into the shared `_hookio.py` module so both hooks detect only genuine top-level invocations, instead of duplicating or inventing a new parser.
- `post-tool-use.py` gains its own `is_issue_create_command()` / `is_pr_create_command()` wrappers (mirroring `pr-merge-reminder.py`'s style) built on the shared engine — the wrapper functions themselves are intentionally **not** hoisted/shared (the two hooks' create-detection needs already diverge: issue-vs-PR union here, PR-only there), only the ~130-line parser is.
- Documented as ADR-050 Amendment 5, including an explicit note that `pre-tool-use-canonical-mutate-guard.py`'s heredoc/quote handling is narrower and not equivalent protection (a candidate follow-up, out of scope here).

## Problem

Reproduced four times in the dev-env#494 fix session alone: a `git commit -m "..."` message, a `grep -E '...'` pattern argument, a `gh project item-edit --text "..."` value, and a `gh pr comment --body "..."` all quoted the example command `gh pr create --fill && gh pr merge --auto` while describing that fix, and each spuriously matched the detector. No harmful mutation occurred (no GitHub URL was present in the unrelated output), but a false positive whose output *does* contain an unrelated GitHub URL could misfile the wrong item onto the project board.

**Live confirmation during this very session:** the still-unmerged (old) hook fired on this branch's own commit message, which necessarily quotes "gh issue create" / "gh pr create" while describing the bug — a real-time repro of exactly the false positive this PR fixes. No project mutation occurred (no GitHub URL in `git commit`'s output), matching the documented harmless-advisory impact.

## Fix

See the full design and rationale in [ADR-050 Amendment 5](https://github.com/brownm09/dev-env/blob/fix/499-unanchored-create-detection/docs/adr/050-shared-hookio-sibling-hook-fixes.md#amendment-5-2026-07-02--extracting-scan_top_level-into-_hookiopy-fixes-post-tool-usepys-unanchored-create-detection).

## Testing

Ran the three required `## Testing` suites for the files touched (dev-env `CLAUDE.md` items 1, 12, 13, 28), plus the full hook-script syntax check:

```
-> SYNTAX_OK

Tests: 31 passed, 0 skipped, 0 failed

py -3 claude/scripts/tests/test_post_tool_use.py
Tests: 41 passed, 0 skipped, 0 failed

py -3 claude/scripts/tests/test_pr_merge_reminder.py   (regression check -- no functional change)
Tests: 42 passed, 0 skipped, 0 failed
```

New coverage added: direct engine-level tests for `scan_top_level` in `test_hookio.py` (anchored-match semantics, non-splitting inside quotes/subshells/heredocs, correct splitting on `&&`/`;`/`||`/newline); `test_post_tool_use.py` gains tests for the two new wrapper functions covering all four dev-env#499 false-positive reproductions for **both** `gh pr create` and `gh issue create`, subshell/quote/cd-prefix/chained-with-merge cases, and an end-to-end `subprocess`-driven pair (mirroring `test_worktree_path_check.py`'s pattern) pinning that the pre-existing `exit_code != 0` gate still short-circuits correctly immediately after the detection swap — without either branch invoking a live `gh` call.

Suppression grep and test-integrity grep both ran clean (no matches — Python-only change, no `!`/`ts-ignore`/skip-marker patterns). No `package.json` touched. dev-env has no `baseline_test_failure_tracking` opt-in, so the pre-existing-failure baseline diff does not apply.

Per `/review` ([comment](https://github.com/brownm09/dev-env/pull/508#issuecomment-4863130180)): zero blocking findings. Three non-blocking items:

- ADR-050 Amendment 5 length/narrative duplication — partially addressed in `7aece64` (dropped a hardcoded test count that would have rotted); full trim deferred, not urgent enough to block.
- `scan_top_level` module docstring frames itself around its origin hook rather than its current shared contract — deferred to the PR that closes dev-env#509.
- `usage-snapshot.py` / `post-pr-merge-project.py` still carry their own independent (behaviorally identical) copies of `_find_heredoc_end`/`_scan_top_level`, predating this PR — filed [dev-env#509](https://github.com/brownm09/dev-env/issues/509).

Closes #499


brownm09:
## Review: fix(hooks): anchor gh issue/pr create detection to top-level statements

### Summary

**Ready to merge.** Two parallel deep-dive passes (correctness/security; reliability/performance/maintainability) plus my own mechanical checks found **zero blocking findings**. The parser relocation from `pr-merge-reminder.py` into `_hookio.py` was verified byte-for-byte behavior-preserving via AST-level comparison, the new `post-tool-use.py` detection wrappers correctly mirror the established anchoring pattern, and the new end-to-end tests are hermetic (no live `gh`/network dependency). Three non-blocking maintainability notes and one filed-and-linked follow-up round out the findings — all already dispositioned below. Diff size (~770 lines across 8 files) is appropriately scoped: one cohesive extraction + fix + tests + docs, not something that should be split further.

---

### Non-Blocking Findings

**[maintainability]** `docs/adr/050-shared-hookio-sibling-hook-fixes.md` (Amendment 5)
The amendment restates content that also lives in `_hookio.py`'s module docstring, `post-tool-use.py`'s inline comment, and the PR description (the false-positive class, the four repro shapes, the "engine shared, not wrappers" rationale). Multiple copies of the same narrative is a drift risk — updating one doesn't guarantee the others follow.
**Suggestion:** Next time this ADR is touched, trim to the load-bearing, non-duplicated content (the scope decision + the canonical-mutate-guard out-of-scope note); the rest is already stated closer to the code.

**[maintainability]** `claude/scripts/_hookio.py:17-23` (module docstring)
Frames `scan_top_level` around its origin hook (`pr-merge-reminder.py`'s merge/create/push detection) rather than its current shared contract. Will read as stale once a third caller lands.
**Suggestion:** Reword to lead with the generic contract; mention origin as history. Natural to fold into whichever PR closes the follow-up below.

**[maintainability]** `usage-snapshot.py` / `post-pr-merge-project.py` still carry independent copies of `_find_heredoc_end` and a merge-only `_scan_top_level`, pre-dating this PR. Verified behaviorally identical to the now-shared `scan_top_level` (same quote/subshell/heredoc/`;`/`&&`/`||` handling) — a safe, low-risk consolidation, just not one to fold into this bug-fix PR (touches two more production files plus their own test suites).
**Fix:** Filed as [dev-env#509](https://github.com/brownm09/dev-env/issues/509).

---

### Verification notes (explicit non-findings, checked because the review prompts asked)

- **Parser relocation correctness** — AST-level comparison confirms `scan_top_level`/`_find_heredoc_end` in `_hookio.py` are behaviorally identical to the deleted `pr-merge-reminder.py` originals; only the name and docstring changed.
- **Anchoring semantics** — the new `_ISSUE_CREATE_RE`/`_PR_CREATE_RE` + check functions in `post-tool-use.py` correctly mirror `pr-merge-reminder.py`'s established `.match()`-on-`.lstrip()` pattern; no both/neither-fire edge case.
- **`Callable` import removal** — confirmed safe, zero remaining references in `pr-merge-reminder.py`.
- **Performance** — `scan_top_level` is a single-pass O(n) scan with no backtracking; called on every Bash tool invocation, but cost is negligible even on pathological input.
- **New end-to-end tests** — hermetic: the temp-dir config path resolves before `canonical_root_via_git`'s subprocess branch is ever reached, so no live `git`/`gh` dependency. Matches the established `test_worktree_path_check.py`/`test_canonical_mutate_guard.py` pattern exactly.
- **Doc-reconciliation (Step 2b)** — `claude/scripts/**` changed but README.md/docs/REFERENCE.md aren't in the diff; investigated and confirmed not a gap: `_hookio.py` is explicitly documented as a non-tabulated shared library in both docs, and the Hooks table entries for `post-tool-use.py`/`pr-merge-reminder.py` remain accurate (trigger/exit-code/behavior summary unchanged by this PR).
- **Doc coverage (Step 2c)** — no ancestor READMEs affected; the equivalent indices (`CLAUDE.md` Testing items 12/13/28, `docs/adr/INDEX.md`) were already updated in this PR.
- **Test coverage gate (Step 2d)** — new behavior (bug fix) has substantial new test coverage in the diff. Satisfied.
- **Test integrity gate (Step 2e)** — diff is Python-only; the skill's JS/TS scanners don't apply. Manually verified: registered-test counts match exactly across all three suites (confirmed by running each), no skip/threshold/bypass mechanism exists in this repo's custom test-runner convention, and the PR body's `Tests: N passed, 0 skipped, 0 failed` line is present for all three.
- **Security** — no injection surface introduced; all subprocess calls use argument-list form, and this change only decides *whether* a hook fires, never constructs a command. If anything, the fix reduces risk by eliminating false-positive detections.

---

### Findings Disposition

- ADR-050 Amendment 5 length/duplication — non-blocking; partially addressed in `7aece64` (dropped a hardcoded test count that would have rotted); full trim deferred, not urgent enough to block.
- `scan_top_level` docstring origin-framing — non-blocking; deferred to the PR that closes dev-env#509.
- `usage-snapshot.py`/`post-pr-merge-project.py` duplicate parser copies — filed [dev-env#509](https://github.com/brownm09/dev-env/issues/509).

---

*Reviewed by `claude-opus-4-8` (parallel deep-dive subagents) with `claude-sonnet-5` orchestration, via Claude Code*

PR #512, opened 2026-07-02
refactor(hooks): consolidate usage-snapshot.py and post-pr-merge-project.py onto shared scan_top_level

brownm09:
## Summary

- `claude/scripts/usage-snapshot.py` and `claude/scripts/post-pr-merge-project.py` each carried their own independent, near-identical copy of `_find_heredoc_end` + a merge-only `_scan_top_level(command: str) -> bool` (hardcoded to `_check_merge_stmt`, not parameterized), predating dev-env#499 / PR #508's extraction of that exact parser into a shared, parameterized `scan_top_level(command, check_fn)` in `claude/scripts/_hookio.py` (ADR-050 Amendment 5).
- Both files now import `scan_top_level` from `_hookio` and call it as `scan_top_level(command, _check_merge_stmt)`, using the `_check_merge_stmt` predicate each file already defined. The local `_find_heredoc_end`/`_scan_top_level` duplicates are removed, along with `post-pr-merge-project.py`'s now-inaccurate "duplicated from pr-merge-reminder.py" comment.
- Behavior-preserving by construction — verified during #508's review that both local scanners handle `;`, `\n`, `&&`, `||`, single/double quotes, `$()` subshells, and heredoc bodies identically to the shared engine.
- Adds ADR-050 Amendment 6 documenting completion of the sweep Amendment 5 started (`docs/adr/050-shared-hookio-sibling-hook-fixes.md`).

Closes #509

## Testing

Ran the dev-env `## Testing` items relevant to this change (hook-script syntax check, plus the three test suites covering the touched code):

```
py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py
py -3 claude/scripts/tests/test_usage_snapshot.py
py -3 claude/scripts/tests/test_post_pr_merge_project.py
py -3 claude/scripts/tests/test_hookio.py
```

- Syntax check: clean, no errors.
- `test_usage_snapshot.py` — Tests: 11 passed, 0 skipped, 0 failed
- `test_post_pr_merge_project.py` — Tests: 16 passed, 0 skipped, 0 failed
- `test_hookio.py` — Tests: 31 passed, 0 skipped, 0 failed

[...]

PR #513, opened 2026-07-02
fix(hooks): converge canonical-mutate-guard onto shared split_top_level engine

### LINKED ISSUE #511: canonical-mutate-guard: converge heredoc/quote handling onto _hookio's shared segmenting engine
## Problem

During `/review` of PR #508 (fix for #499, merged as 65a5f84), ADR-050 Amendment 5 flagged that `claude/scripts/pre-tool-use-canonical-mutate-guard.py` has its own narrow heredoc/quote-adjacent handling that is **not equivalent protection** to the shared `scan_top_level` engine in `claude/scripts/_hookio.py`:

- `_strip_heredoc_command_subs()` only strips one exact shape: `$(cat <<[-]['"]MARKER['"] ... MARKER)`. A plain heredoc (no command-substitution wrapper) isn't touched by it at all.
- `_SEGMENT_SPLIT` is a plain regex (`&&|\|\||;|\n|\|`) with **no quote-tracking**. A command like `git commit -m "text && rm -rf /"` is mis-split inside the quoted string.

The ADR flagged this as "a real API mismatch, not simple reuse" since `scan_top_level(command, check_fn)` is a boolean reducer — it can't give a caller back *which* segment matched, which the guard's `classify()` needs (for the block-reason message and for per-segment cd-scope / `-C`-redirect-skip logic).

## Concrete impact — two false positives verified by hand-tracing the current code

1. **Quote-mis-split false positive.** `git log --grep="foo && git checkout -b evil"` — the unquoted-aware split produces a fake segment `git checkout -b evil"` that gets flagged as mutating. A harmless `git log` search gets blocked in a canonical checkout.
2. **Bare-heredoc-body false positive.** `git status <<EOF` / `git commit --amend` (body) / `EOF` — `_HEREDOC_CATSUB_RE` only covers the `$(cat <<MARKER...)` idiom (the #481 fix), not a bare heredoc. The body line `git commit --amend`, after the `\n`-split, becomes its own segment and is misclassified as a real invocation — same failure class as #481 (dev-env#481), just the non-command-substitution variant, currently untested and currently broken.

## Proposed fix

Extract the segmenting core of `_hookio.scan_top_level` into a new function `split_top_level(command: str, *, split_pipe: bool = False) -> list[str]` that returns the quote/subshell/heredoc-aware top-level segments instead of just reducing to a boolean. Rewrite `scan_top_level` as a thin wrapper: `any(check_fn(seg) for seg in split_top_level(command))` — behavior-preserving for its existing two consumers (`pr-merge-reminder.py`, `post-tool-use.py`).

`pre-tool-use-canonical-mutate-guard.py` then imports `split_top_level(cmd, split_pipe=True)` (pipe-splitting is opt-in and guard-only, since `git commit -F -` can read a piped heredoc and the other two consumers have no need for it) and deletes its own `_HEREDOC_CATSUB_RE` / `_strip_heredoc_command_subs()` / `_SEGMENT_SPLIT` entirely — the shared engine's `$()`-subshell and heredoc opacity subsumes the narrow regex for free, fixing both false positives above as a side effect of convergence.

This matches the repo's established shared-pure-helper pattern (ADR-050's five amendments, ADR-057, ADR-072, ADR-073) rather than letting a sixth hook re-derive its own copy of quote/heredoc handling.

## Scope

Single PR (not decomposed): the `_hookio.py` extraction and the guard's convergence are one atomic change — introducing `split_top_level` with no consumer, or converting the guard before the function exists, are each incomplete on their own.

Touches a shared module (`_hookio.py`) two other hooks depend on (`pr-merge-reminder.py`, `post-tool-use.py`) — their test suites (`test_pr_merge_reminder.py`, `test_post_tool_use.py`) must be run as a regression check even though those files aren't edited directly.

## References

- `claude/scripts/pre-tool-use-canonical-mutate-guard.py`
- `claude/scripts/_hookio.py`
- `claude/scripts/tests/test_canonical_mutate_guard.py`
- [docs/adr/050-shared-hookio-sibling-hook-fixes.md](https://github.com/brownm09/dev-env/blob/main/docs/adr/050-shared-hookio-sibling-hook-fixes.md) — Amendment 5, "A related but narrower implementation, left out of scope"
- [docs/adr/071-canonical-checkout-mutate-guard-hook.md](https://github.com/brownm09/dev-env/blob/main/docs/adr/071-canonical-checkout-mutate-guard-hook.md)
- dev-env#481 — the original command-substitution heredoc false-positive fix this generalizes

brownm09:
## Summary

- Extracts `split_top_level(command, *, split_pipe=False) -> list[str]` from `_hookio.scan_top_level()`'s internals — the same quote/subshell/heredoc-aware stack parser, now returning the segment list instead of reducing it to a boolean. `scan_top_level` becomes a thin `any(check_fn(seg) for seg in split_top_level(command))` wrapper, behavior-preserving for its existing two callers (`pr-merge-reminder.py`, `post-tool-use.py`).
- `pre-tool-use-canonical-mutate-guard.py` now imports `split_top_level(cmd, split_pipe=True)` in both `classify()` and `_has_override()`, replacing its own `_SEGMENT_SPLIT` regex splitter. `_HEREDOC_CATSUB_RE`/`_strip_heredoc_command_subs()` are deleted entirely — the shared engine's `$()`-subshell and heredoc opacity subsumes that narrow regex.
- `split_pipe` is opt-in (default `False`) so the two existing `scan_top_level` consumers see zero behavior change; only the guard, which needs pipe-awareness (`git commit -F -` reading piped stdin), opts in.
- **Update (post-`/review`):** fixed a false-negative regression the convergence itself introduced — see Review findings disposition below.
- **Update (post-rebase):** PR #512 merged concurrently and also claimed "ADR-050 Amendment 6" for a different, unrelated consolidation — rebased onto the new `main` and renumbered this PR's amendment to **Amendment 7** throughout (code comments, docstrings, both ADR files). No functional conflict — PR #512 converges `usage-snapshot.py`/`post-pr-merge-project.py` onto the *existing* `scan_top_level` (unaffected by this PR's internal refactor of it), verified by re-running its own test suites post-rebase.

## Why

ADR-050 Amendment 5 (from PR #508 / dev-env#499) flagged that the guard's own heredoc/quote handling was **not equivalent protection** to `scan_top_level`: its heredoc stripper only covered the `$(cat <<MARKER...)` idiom, and its segment splitter had no quote-tracking at all. That amendment called converging it onto `scan_top_level` "a real API mismatch, not simple reuse" and left it as a flagged follow-up. This PR resolves the mismatch by extracting the missing primitive (segment list, not just a boolean) rather than accepting the gap permanently.

Hand-tracing the pre-fix code turned up two concrete, currently-live false positives (not just theoretical divergence):

1. **Quote-mis-split.** `git log --grep="foo && git checkout -b evil"` — the old regex splitter had no quote-tracking, so it split *inside* the quoted `--grep` value and misclassified a harmless `git log` as a `checkout`.
2. **Bare-heredoc-body.** `git status <<EOF` / `git commit --amend` (body) / `EOF` — the old heredoc stripper only recognized the `$(cat <<MARKER...)` command-substitution shape (dev-env#481); a *bare* heredoc was untouched, so the body line became its own segment and was misclassified as a real invocation.

Both are fixed as a side effect of convergence, verified with new regression tests (including one end-to-end `main()` subprocess test proving the fix reaches the real hook process).

Closes #511

## Testing

Ran (all from repo root):

```
py -3 claude/scripts/tests/test_pr_merge_reminder.py   # regression: other scan_top_level consumer
py -3 claude/scripts/tests/test_post_tool_use.py       # regression: other scan_top_level consumer
py -3 claude/scripts/tests/test_usage_snapshot.py           # regression: PR #512's scan_top_level consumer
py -3 claude/scripts/tests/test_post_pr_merge_project.py    # regression: PR #512's scan_top_level consumer
```

Results (final, after rebase onto `main` post-PR #512, commit `8b39c1f`):
- Hook-script syntax check: clean (0.4s)
- `test_hookio.py`: **Tests: 44 passed, 0 skipped, 0 failed** (0.2s) — 31 pre-existing `scan_top_level` tests unchanged + 13 new `split_top_level` tests
- `test_canonical_mutate_guard.py`: **Tests: 38 passed, 0 skipped, 0 failed** (3.0s) — 28 original pre-existing tests unchanged + 10 new tests (2 quote/heredoc false-positive fixes, 2 pipe-splitting, 4 review-fix regressions, 1 review-fix e2e proof, 1 DOTALL-alternative-would-be-wrong guard)
- `test_pr_merge_reminder.py` (regression, not edited): **Tests: 42 passed, 0 skipped, 0 failed** (0.2s)
- `test_post_tool_use.py` (regression, not edited): **Tests: 41 passed, 0 skipped, 0 failed** (0.4s)
- `test_pyw_stdio.py`: **Tests: 7 passed, 0 skipped, 0 failed** (0.7s)
- `test_usage_snapshot.py` (regression, not edited, PR #512's own suite): **Tests: 11 passed, 0 skipped, 0 failed**
- `test_post_pr_merge_project.py` (regression, not edited, PR #512's own suite): **Tests: 16 passed, 0 skipped, 0 failed**

199 tests passed, 0 skipped, 0 failed across all seven files. Suppression grep (`git diff origin/main`) and test-integrity grep both clean — no new suppressions, no skip markers, no deleted tests, no lowered thresholds. No `package.json` touched (lockfile policy N/A).

## Risk-dimension audit (Pass 3)

1. **Testing** — covered above.
2. **Observability** — N/A change; block reason still written to stderr with exit 2, unchanged.
3. **Security** — this *is* the security-relevant surface (the concurrency-collision guard from ADR-071/dev-env#453). `/review` caught a real false-negative regression in the initial version (see disposition below), now fixed and independently verified — the merged state is strictly-or-neutral versus pre-PR: quote-tracking eliminates the two original false positives without weakening detection (every pre-existing mutating-verb test still passes, plus 5 new regression tests specifically targeting the multi-line-segment failure mode), and the opt-in `split_pipe` flag scopes the only behavior change to the one caller that needs it.
4. **Resilience** — `split_top_level` and `_first_line` are pure functions (no I/O/subprocess); the hook's existing fail-open guarantees (`_resolve_git_toplevel`, malformed-JSON, non-git-cwd) are untouched. Rollback = revert.
5. **Performance** — negligible; command strings are short single Bash-tool-call strings. The post-review fix also removed a minor double-parse (`classify()`/`_has_override()` now share one `segments` list computed once in `main()`).
6. **Data integrity** — N/A, no schema/data changes, stateless hook.

## Docs

- [ADR-050](docs/adr/050-shared-hookio-sibling-hook-fixes.md) **Amendment 7** — the convergence decision, the two original false positives, coverage, and (added post-`/review`) a subsection documenting the false-negative regression `/review` caught and its fix. Numbered 7, not 6, because PR #512 concurrently claimed Amendment 6 for its own consolidation — see the rebase note above.
- [ADR-071](docs/adr/071-canonical-checkout-mutate-guard-hook.md) — cross-referenced (Testing bullet + References).
- `docs/REFERENCE.md` — hooks-table entry updated to describe the shared-engine mechanism instead of the retired regex.
- `docs/adr/INDEX.md` — tags updated.

`/review` (two parallel Opus subagents — correctness/security, reliability/maintainability — plus independent verification) ran on the initial version and found 2 blocking + 5 non-blocking findings. Full review: https://github.com/brownm09/dev-env/pull/513#issuecomment-4867084991

- **Multi-line-segment false negative in `_GIT_INVOCATION_RE`** (blocking correctness/security) — a real `git commit -m "$(cat <<'EOF' ...)"` (this repo's own commit-message idiom) was silently classified as non-mutating. **Fixed in `8b39c1f`** (new `_first_line()` helper).
- **Multi-line-segment false negative in `_REDIRECT_RE`** (blocking correctness/security, found during my own verification of the first finding) — a commit whose heredoc body merely mentioned `git -C /somewhere` as prose was wrongly treated as redirected to another repo and skipped. **Fixed in `8b39c1f`** (same `_first_line()` helper).
- **Double-parse in `main()`** (non-blocking performance) — `classify()`/`_has_override()` each re-ran `split_top_level` independently. **Fixed in `8b39c1f`** (optional `segments` parameter, computed once).
- **Docstring duplication across 3 locations** (non-blocking maintainability) — the engine's quote/heredoc semantics were re-explained near-verbatim in the module docstring, `classify()`'s docstring, and `split_top_level`'s own docstring. **Fixed in `8b39c1f`** (trimmed `classify()`'s copy to a pointer).
- **Pre-existing mutating-verb-list duplication across 4 locations** (non-blocking maintainability, not introduced by this PR) — **filed [dev-env#514](https://github.com/brownm09/dev-env/issues/514)**, out of scope for this PR per the reviewing subagent's own recommendation.
- **Trailing empty segment from a bare `|` with `split_pipe=True`** (non-blocking reliability) — confirmed harmless, matches pre-existing splitter behavior. **No fix needed.**
- **Unterminated-quote tail-drop narrows prior coverage** (non-blocking correctness) — intentional fail-permissive contract (same philosophy as the hook's other fail-open paths), already covered by a test. **No fix needed.**


brownm09:
## Review: fix(hooks): converge canonical-mutate-guard onto shared split_top_level engine

### Summary

Ready to merge. The convergence itself is sound and well-tested, but the parallel review pass (two Opus subagents, correctness/security + reliability/maintainability, plus independent verification) caught a **serious false-negative regression** the convergence introduced: two of the guard's classification regexes were never updated for the fact that `split_top_level`'s heredoc/subshell opacity lets a single segment span multiple physical lines now. A real `git commit -m "$(cat <<'EOF' ...)"` — this repo's own documented commit-message idiom — was silently classified as non-mutating, and a commit whose heredoc body merely mentioned `git -C /somewhere` as prose was wrongly treated as redirected to another repo and skipped. Both are now fixed (commit `b74f4aa`), independently verified, and covered by 5 new regression tests. All other findings are non-blocking and have been addressed or explicitly deferred with rationale below.

---

### Blocking Findings

**[correctness/security]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py` (`is_mutating_segment`, was line ~204) / `claude/scripts/_hookio.py` (`split_top_level`)
`_GIT_INVOCATION_RE = re.compile(r"^git(?:\.exe)?\s+(.*)$", re.IGNORECASE)` has no `re.DOTALL`. `split_top_level`'s heredoc/`$(...)` opacity means a segment can now legitimately span multiple physical lines (that's the point — it keeps the whole span inside one segment instead of splitting it out). On such a segment, `.*` cannot cross the first embedded `\n` and `$` (not `re.MULTILINE`) requires the true end of string, so the match fails outright and `is_mutating_segment` returns `False` — not "safe", just unmatched.
This defeats the hook's purpose for its single most common trigger. `git commit -m "$(cat <<'EOF' ... EOF)"` is this repo's own documented commit-message convention; a second concurrent session committing that way in a shared canonical checkout — exactly the dev-env#453 collision class this hook exists to prevent — would have proceeded silently unguarded. Independently verified before trusting the report: `classify()` returned `None` (allow) for `git commit -m "$(cat <<'EOF'\n...)"`, `git commit -F - <<EOF...EOF`, `git merge -m "$(cat <<EOF...)" feature`, and `git reset --hard $(cat <<EOF...)`.
**Fix:** Added `_first_line(segment)` — the git invocation and its flags only ever appear on a segment's own first physical line; a heredoc/command-substitution body is data, not more command syntax. `is_mutating_segment` now matches against `_first_line(stripped)` instead of the whole (possibly multi-line) string. Rejected the `re.DOTALL` alternative (see maintainability note below on why). Fixed in `b74f4aa`.

**[correctness/security]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py` (`classify`, `_REDIRECT_RE.search(seg)`)
Found during my own verification of the finding above, not by either subagent. `_REDIRECT_RE = re.compile(r"(?:^|\s)git\s+(?:-C\s+\S|--git-dir=\S)")` is used with `.search()` (unanchored) over the *whole* segment, and its `(?:^|\s)` alternation treats an embedded `\n` exactly like a space. A commit whose heredoc body merely mentions `git -C /somewhere` as prose text — e.g. `git commit -m "$(cat <<'EOF'\nSee also: git -C /tmp status for context\nEOF\n)"` — matches `_REDIRECT_RE` on that body text and gets `continue`d past in `classify()`'s loop, even though the segment's actual top-level invocation (`git commit -m ...`) was never redirected to another repo at all. Reproduced directly: `classify()` returned `None` for this exact case before the fix.
**Fix:** Same `_first_line()` helper — `classify()` now checks `_REDIRECT_RE.search(_first_line(seg))` instead of the whole segment, so body text after an embedded newline can no longer masquerade as a `-C`/`--git-dir` redirect on the segment's own invocation. Fixed in `b74f4aa`.

---

### Non-Blocking Findings

**[performance]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py` (`main`, `classify`, `_has_override`)
`main()` called `classify(cmd)` then `_has_override(cmd)` in sequence, each independently re-running the O(n) `split_top_level` parse over the same command string — small in absolute terms (short shell command strings), but the hook is explicitly architected "cheapest checks first," and paying the parse twice cuts against that stated intent.
**Fix:** `classify()` and `_has_override()` now accept an optional `segments: list = None` parameter; `main()` computes `segments` once and passes it to both. Every existing direct caller (all of `test_canonical_mutate_guard.py`) keeps calling either function with just `cmd`, unchanged. Fixed in `b74f4aa`.

**[maintainability]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py` (module docstring, `classify()` docstring) vs `claude/scripts/_hookio.py` (`split_top_level` docstring)
The quote/subshell/heredoc-opacity semantics of `split_top_level` were re-explained in near-identical multi-paragraph prose in three places (the guard's module docstring, `classify()`'s own docstring, and the engine's own docstring) — exactly the kind of drift risk the convergence itself was meant to reduce (a future change to the engine's semantics would need updating in three places, easy to miss two of them).
**Fix:** Trimmed `classify()`'s docstring to a one-line pointer at `split_top_level`'s own docstring and the module docstring's step 3, keeping only the `classify()`-specific consequences (segments-param reuse, cd-scope, `-C`-skip, the new first-line rationale). Fixed in `b74f4aa`.

**[maintainability]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py` (`is_mutating_segment`, mutating-verb list)
Pre-existing, not introduced or worsened by this PR: the mutating-verb list is spread across the branch chain in `is_mutating_segment` with no single canonical source, independently re-spelled in four prose locations (this file's module docstring, `claude/CLAUDE.md`, ADR-071, `docs/REFERENCE.md`) that must stay in sync by hand. This PR's whole thesis is "stop re-deriving parsing logic per hook," so a reviewer landing here will notice the tension, but extracting the verb set to a single module constant is a separate, unrelated refactor.
**Disposition:** Deferred — filed as [dev-env#514](https://github.com/brownm09/dev-env/issues/514) rather than expanded in this diff, per the subagent's own recommendation ("Out of scope for this PR... No action required in this PR").

**[reliability]** `claude/scripts/_hookio.py` (`split_top_level`, trailing `|` with `split_pipe=True`)
A trailing bare `|` (no content after it) yields a final whitespace-only empty-ish segment. Confirmed harmless: such a segment is never a `cd`, never matches `_REDIRECT_RE`, never a mutating verb, never the override token — and the pre-existing regex splitter emitted the same kind of empty segments, so this is not a new behavior.
**Disposition:** No fix needed — documented for the record only, matching the subagent's own conclusion.

**[correctness]** `claude/scripts/_hookio.py` (`split_top_level`, unterminated-quote tail-drop)
Lower severity: a mutating command after an unterminated quote/subshell (e.g. `git log --grep="foo && git checkout -b evil` with no closing quote) is dropped by `split_top_level`'s documented fail-permissive contract, where the old regex splitter would have still caught the trailing `git checkout` segment. An unterminated quote is a shell syntax error that would never actually execute, so the practical risk is negligible, and this is the same "fail open on ambiguous input" philosophy the rest of the hook already follows (malformed JSON, non-git cwd, etc.).
**Disposition:** No fix needed — intentional, documented, and covered by `test_split_top_level_unterminated_quote_drops_trailing_segment`, matching the subagent's own "not blocking" assessment.

---

### Findings Disposition

- Multi-line-segment false negative in `_GIT_INVOCATION_RE` (blocking correctness/security) — **fixed in `b74f4aa`**
- Multi-line-segment false negative in `_REDIRECT_RE` (blocking correctness/security) — **fixed in `b74f4aa`**
- Double-parse in `main()` (non-blocking performance) — **fixed in `b74f4aa`**
- Docstring duplication across 3 locations (non-blocking maintainability) — **fixed in `b74f4aa`**
- Pre-existing mutating-verb-list duplication across 4 locations (non-blocking maintainability) — **filed [dev-env#514](https://github.com/brownm09/dev-env/issues/514)**, out of scope for this PR
- Trailing empty segment from a bare `\|` with `split_pipe=True` (non-blocking reliability) — **no fix needed**, documented, harmless, matches pre-existing splitter behavior
- Unterminated-quote tail-drop narrows prior coverage (non-blocking correctness) — **no fix needed**, intentional fail-permissive contract, covered by an existing test

---

*Reviewed by `claude-opus-4-8` (parallel subagents) via Claude Code, synthesized and independently verified by `claude-sonnet-5`*

PR #517, opened 2026-07-02
fix(tests): narrow winsubp underscore-skip exemption to _winsubp.py only

### LINKED ISSUE #515: Narrow test_every_subprocess_using_hook_imports_winsubp's underscore skip to _winsubp.py only
## Problem

`claude/scripts/tests/test_pyw_stdio.py`'s `test_every_subprocess_using_hook_imports_winsubp` is supposed to statically verify that every hook script in `claude/scripts/` which spawns a subprocess also imports `_winsubp` (the module suppressing Windows console-flash and defaulting subprocess text-mode decoding to UTF-8 — see ADR-007). Its directory scan currently skips **every** underscore-prefixed file (`if path.name.startswith("_"): continue`), not just `_winsubp.py` itself (the one module that legitimately cannot import itself).

This blanket exemption meant the guard never actually scanned any of the repo's underscore-prefixed shared modules — including `_hookio.py` and `_gh_project.py`, both of which contain real `subprocess.run` call sites. PR #506 (closing dev-env#503) found and fixed `_hookio.py` missing its own `import _winsubp` (it was only accidentally protected because every current caller happened to import `_winsubp` first), and [ADR-007](docs/adr/007-hook-command-invocation.md)'s "2026-07-02 (follow-up 3)" entry explicitly flagged the blanket-skip as "tracked as a separate follow-up rather than fixed here."

## Goal

- Narrow the skip so only `_winsubp.py` is exempted; every other underscore-prefixed module is scanned like any other hook script.
- Audit all underscore-prefixed modules in `claude/scripts/` for subprocess use and add `import _winsubp` to any that need it but don't already have it.
- Add test coverage (a negative control) that would catch a regression back to the blanket skip.
- Amend ADR-007 with a follow-up entry documenting the fix, per its own explicit forward-reference at line 93.

## References

- [ADR-007](docs/adr/007-hook-command-invocation.md), line 93
- dev-env#503, PR #506

brownm09:
## Summary

- `test_every_subprocess_using_hook_imports_winsubp` (`claude/scripts/tests/test_pyw_stdio.py`) skipped **every** underscore-prefixed file in `claude/scripts/` via `path.name.startswith("_")`, not just `_winsubp.py` itself (the one module that legitimately cannot import itself). This let subprocess-using shared modules bypass the guard entirely — PR #506 (dev-env#503) found and fixed `_hookio.py` missing its own `import _winsubp` reactively, and [ADR-007](docs/adr/007-hook-command-invocation.md) line 93 explicitly flagged the blanket skip as "tracked as a separate follow-up rather than fixed here." This PR is that follow-up.
- Replaced the blanket skip with a named `_is_winsubp_module(name: str) -> bool` predicate (matches this repo's `_is_<noun>(...) -> bool` convention, e.g. `_is_windows()`, `_is_devenv_cwd()`) and added inline negative-control assertions pinning the narrowed behavior — a future regression back to a blanket underscore-prefix skip now fails the test immediately.
- **Audited all 9 underscore-prefixed modules in `claude/scripts/` for subprocess use** (`_winsubp.py`, `_hookio.py`, `_gh_project.py`, `_hookutil.py`, `_journal_shards.py`, `_worktree_liveness.py`, `_worktree_topology.py`, `_repo_scan.py`, `_worktree_canon.py`). Only `_hookio.py` and `_gh_project.py` contain a real subprocess use-site, and both already import `_winsubp` (`_hookio.py` since PR #506; `_gh_project.py` since its creation under ADR-073). The other 6 either don't use subprocess at all, or only mention the word "subprocess" in docstring prose (e.g. "no subprocess"), which the test's own `_SUBPROCESS_USE_RE` correctly does not match. **Conclusion: zero source-code changes were needed to any hook script — this is a test-only fix.**
- `test_no_hook_spawns_python_via_py_launcher` (the sibling test) was checked and needs **no change** — its scan loop has no underscore-prefix filter of any kind and already covers every underscore-prefixed file unconditionally.
- Amended [ADR-007](docs/adr/007-hook-command-invocation.md) with a new "2026-07-02 (follow-up 4)" Decision History entry documenting the gap, audit, and fix, per its own forward-reference.

## Testing

- `py -3 claude/scripts/tests/test_pyw_stdio.py`: `Tests: 7 passed, 0 skipped, 0 failed`. `test_every_subprocess_using_hook_imports_winsubp`'s `checked` count went from **26 → 28** (empirically confirmed by stashing the fix and re-running) — exactly the `+2` the audit predicted (`_hookio.py` + `_gh_project.py`, previously silently excluded).
- Hook-script syntax check (`## Testing` item 1, `ast.parse` over `claude/scripts/*.py`): clean.
- Suppression grep (ADR-026) and test-integrity grep (ADR-029): both empty — no suppressions, skip markers, or deleted tests in this diff.
- `baseline_test_failure_tracking` is not enabled for this repo (no such key in `.claude/hook-config.json`), so ADR-030's baseline-diff step is N/A.

Closes #515


brownm09:
## Review: fix(tests): narrow winsubp underscore-skip exemption to _winsubp.py only

### Summary

Ready to merge as-is. This is a small (37-line), single-purpose fix that does exactly what it claims: it replaces a blanket `startswith("_")` skip with a named `_is_winsubp_module()` predicate that exempts only `_winsubp.py`, and the accompanying ADR-007 amendment accurately documents the gap, the audit, and the fix. I independently re-traced the loop logic and the sibling `test_no_hook_spawns_python_via_py_launcher` function against the PR branch (not just the diff hunks) and confirmed both match the PR's narrative precisely — the `checked` 26→28 delta, the "no source changes needed" audit conclusion, and the "sibling test needs no change" claim all check out against the actual code. No blocking or non-blocking findings. Diff also contains a non-JS (Python) test file; I manually checked it against the pytest skip idioms (`@pytest.mark.skip`, `pytest.skip(`, `pytest.xfail(`, `unittest.skip`) since Step 2e's automated patterns are JS/TS-scoped — none present, so no language-scope gap to flag.

---

PR #526, opened 2026-07-02
feat: opt-in ephemeral-diff prunability signal for prune-merged-worktrees.py

### LINKED ISSUE #525: prune-merged-worktrees.py can't detect journal-compose's superseded scaffolding as merged
## Problem

\`prune-merged-worktrees.py\`'s \`is_merged()\` only recognizes two kinds of "merged": a branch
that's an ancestor of \`origin/main\`, or a branch with its own merged GitHub PR. Neither
condition holds for a real, verified-safe class of worktrees in \`brownm09/engineering-journal\`.

That repo's \`/journal-compose\` workflow spawns background-agent worktrees on throwaway
\`claude/<slug>\` branches that commit per-session \`*.stub.md\`/\`*.manifest.jsonl\`/\`open-prs/*.json\`
scaffolding files. That content later gets folded into a canonical composed doc on \`main\` by
journal-compose's Step 9 ("Delete stub files and release lock"), **which then deletes the raw
stub files from main**. So the throwaway branch's own commits never become an ancestor of
\`main\` (the files were removed, not carried forward) and never get their own PR (only the
shared \`draft/YYYY-MM-DD\` branch does).

Result: even though the daily \`prune-stale-worktrees\` routine (\`--scan-dir C:/Users/brown/Git\`)
already scans engineering-journal, these worktrees are permanently invisible to its merge check
and accumulate indefinitely. A manual audit found 47 stale worktrees (all individually
content-verified safe — every non-scaffolding file in every branch's diff already exists at the
same path on \`origin/main\`) plus 7 additional orphaned local branches, cleaned up manually as a
one-time pass. Without a code fix, the same backlog will reaccumulate.

## Proposed direction

Add an opt-in, per-repo additional prunability signal: a branch counts as prunable if its entire
diff vs \`origin/main\` matches a configurable set of "ephemeral" filename patterns (e.g.
\`.claude/hook-config.json\`'s \`prune_ephemeral_patterns\`), checked only as a fallback when the
existing ancestor/PR checks already say "not merged" — stays off by default for every repo that
doesn't opt in, never tightens the existing \`is_merged()\` behavior.

Full design (function signatures, exact insertion points, tests, ADR) already worked out in a
plan document from the investigating session — implementing in the immediate follow-up PR.

brownm09:
## Summary
- `is_merged()` only recognizes ancestor-of-main or a merged PR; engineering-journal's `/journal-compose` workflow produces `claude/<slug>` branches whose content is folded into a composed doc and then **deleted** from main by compose Step 9, so those branches are permanently invisible to either check. A manual audit found 47 stale worktrees + 7 orphaned branches accumulated from this gap (all individually content-verified safe, cleaned up by hand in the same session).
- Adds `diff_files()` / `files_are_all_ephemeral()` / `load_ephemeral_patterns()`: a branch not otherwise merged is still treated as merged if its entire diff vs `origin/main` matches a per-repo opt-in `prune_ephemeral_patterns` regex list in `.claude/hook-config.json`. Off by default, additive only, lazy (loaded once per repo, only when a branch isn't already merged the normal way) — never tightens `is_merged()` itself, same precedent as ADR-051's liveness guard.
- Engineering-journal's own opt-in config lands in a **separate** follow-up commit on its `draft/YYYY-MM-DD` branch only after this PR merges (that repo has no feature-branch/PR flow of its own for repo files).

## Test plan
- [x] `py -3 claude/scripts/tests/test_prune_merged_worktrees.py` — 14 passed, 0 skipped, 0 failed (13 new tests covering `files_are_all_ephemeral` and `load_ephemeral_patterns`, including the OSError-handling fix from self-review; `diff_files()` and the `prune_one()` integration have no dedicated test, matching this file's existing convention that merge-detection/worktree-list steps are exercised via `--dry-run`, not unit tests — see the two other untested helpers `is_merged`/`is_dirty`)
- [x] Hook-script syntax check (`ast.parse` over `claude/scripts/*.py`) — clean
- [x] `--dry-run --repo-path C:/Users/brown/Git/dev-env` smoke test — completes cleanly, existing skip reasons unaffected
- [x] Suppression grep — clean
- [x] Test-integrity grep — clean

Tests: 14 passed, 0 skipped, 0 failed

## ADR
[ADR-075](docs/adr/075-ephemeral-diff-worktree-pruning.md) (`docs/adr/075-ephemeral-diff-worktree-pruning.md`) — written in this PR per the ADR-warrant check (touches a documented script's behavior, sets a new per-repo opt-in convention). Note: originally slotted as ADR-073 during planning; re-checked the filesystem (not just `INDEX.md`) immediately before writing and found two other same-day PRs had already landed 073 and 074 — renumbered to 075 before this PR was opened, so no collision was ever pushed.

## Known scope limitations (documented in the ADR's Consequences)
- Filename-pattern-only: a branch whose diff includes a non-scaffolding file is never auto-pruned by this feature even if that file is *also* safe (e.g. a byte-identical composed doc next to a merely-stale README.md). 4/33 of engineering-journal's `claude/*` branches fell in this residual and were handled by the manual cleanup, not this code path.
- Worktree-scoped, not branch-scoped: `prune_one()` only iterates registered worktrees, so a local branch with no worktree is invisible to this script regardless of this change. Engineering-journal's separate ~89-branch orphaned-local-branch backlog (no worktree attached to any of them) is **not** addressed here — flagged as its own follow-up.

`/review` was run before this PR was opened for outside review (self-review during authoring); all three findings it surfaced were fixed in commit `17270e6` on this branch, pushed before the review comment was posted. None deferred or filed.

- **[reliability]** `load_ephemeral_patterns` caught only `FileNotFoundError`, not the broader `OSError` family (e.g. a transient Windows `PermissionError`) — fixed in `17270e6`.
- **[correctness]** Stale `ADR-073` reference in the test file's own docstring (should be `ADR-075`) — fixed in `17270e6`.
- **[maintainability]** `load_ephemeral_patterns(repo)` called once per worktree instead of once per repo — fixed in `17270e6`.

Full review: https://github.com/brownm09/dev-env/pull/526#issuecomment-4868598302

Closes #525


brownm09:
## Review: feat: opt-in ephemeral-diff prunability signal for prune-merged-worktrees.py

### Summary
Small, well-scoped, appropriately additive PR (362 lines). Found three issues during
self-review — one genuine reliability gap, one stale cross-reference, one maintainability
hoist — all fixed before this comment was posted (see commits `17270e6`). No outstanding
findings. Ready to merge.

---

### Blocking Findings (found and fixed pre-merge)

**[reliability]** `claude/scripts/prune-merged-worktrees.py` — `load_ephemeral_patterns`
Originally caught only `except (FileNotFoundError, json.JSONDecodeError)` around the config
read. On Windows, a transient file lock (antivirus/indexer scanning the file, or a momentary
handle held by another process) raises `PermissionError` — a subclass of `OSError`, not
`FileNotFoundError` — which would have propagated uncaught out of `prune_one()`. In
`--scan-dir` mode, `main()`'s `for repo in repos: prune_one(repo, ...)` loop has no
try/except around that call, so one repo's momentary config-read hiccup would abort the
entire daily scan across every remaining repo — not just skip the ephemeral-diff feature
for that one repo. This exact class of Windows file-handle flakiness was independently
observed in this same session (a `git worktree remove` on an unrelated worktree hit "not a
working tree" after an apparent lingering handle), so it isn't hypothetical for this
environment.
**Fix:** Broadened to `except (OSError, json.JSONDecodeError)` (`FileNotFoundError` is a
subclass of `OSError`, so this is a strict widening, not a behavior change for the already-
covered case). Added `test_load_ephemeral_patterns_unreadable_file_returns_empty`, which
mocks `open` to raise `PermissionError` and asserts the function still returns `[]` rather
than propagating.

**[correctness]** `claude/scripts/tests/test_prune_merged_worktrees.py:8`
The module docstring said "Also covers the ADR-073 ephemeral-diff prunability signal" — a
leftover from before the ADR was renumbered to 075 (two concurrent same-day PRs had already
claimed 073 and 074). A repo-wide grep confirmed every *other* `ADR-073` reference in the
tree correctly points at the real, different ADR-073 (`_worktree_canon`/`_gh_project`
modules) from that concurrent PR — this was the one leftover from my own change.
**Fix:** Corrected to `ADR-075`.

---

### Non-Blocking Findings (found and fixed pre-merge)

**[maintainability]** `claude/scripts/prune-merged-worktrees.py` — `prune_one()`
`load_ephemeral_patterns(repo)` was called inside the per-worktree loop, once for every
not-yet-merged `claude/*` worktree — meaning a repo with many stale worktrees (exactly the
scenario this feature targets; engineering-journal had 30+) would re-read and re-validate
the identical config file and re-compile the same regexes redundantly on every iteration.
**Fix:** Hoisted to a single `ephemeral_patterns = load_ephemeral_patterns(repo)` call
right after `gh_repo = detect_gh_repo(repo)`, mirroring the existing once-per-repo
`detect_gh_repo` pattern; the loop now reuses the hoisted value.

**[maintainability]** Test-language scope note (Step 2e of the review process)
The diff is 100% Python; the codebase's standard pre-PR suppression/test-integrity greps
target JS/TS idioms and cannot inspect Python skip markers. Manually verified: no
`@pytest.mark.skip`/`pytest.skip(` equivalents exist (this repo's tests use a bespoke
offline `assert`-based runner, not pytest), no tests were deleted, and the full suite
(14 tests) passes clean. No action needed beyond this note.

---

### Findings Disposition
All three blocking/non-blocking findings above were fixed in commit `17270e6` on this
branch, pushed before this review was posted. No filed-and-linked deferrals.

---

PR #528, opened 2026-07-02
fix(hooks): roll out confirm_merge_via_gh fallback to the 5 remaining marker-gated hooks

### LINKED ISSUE #504: Roll out confirm_merge_via_gh gh-pr-view fallback to the other five marker-gated hooks
## Problem

PR #497 (closing #489) added a live `gh pr view` confirmation fallback (`should_confirm_via_gh` / `confirm_merge_via_gh` in `claude/scripts/_hookio.py`) for when gh's merge-success marker doesn't survive to a PostToolUse hook's captured `stdout`/`stderr` — a real, reproduced bug on the worktree-cleanup-failure path (`'main' is already checked out`). The fallback was deliberately wired into `post-pr-merge-project.py` **only**, per PR #497's stated rationale: it's the one hook of the six with no other eventual-consistency backstop for a missed action.

PR #497's body claims this rollout is "tracked separately: dev-env#494" — that's a mislink. #494 is an unrelated, already-filed bug (a chained `gh pr create && gh pr merge` command suppressing the create reminder). No issue currently tracks the actual rollout. This issue is that tracker.

## Evidence this isn't hypothetical

Confirmed recurring, not a one-off, all in the same day:
- PR #466 → #474 (marker survived; a *different* bug — `usage-snapshot.py`'s then exit-code-only gate — missed it anyway)
- PR #493 → #489 (marker lost; root-caused and partially fixed by PR #497)
- PR #491 → #496 (marker lost again, 3 sibling hooks silent — same root cause as #489, occurring 32 minutes *before* PR #497's fix landed; see #496 for the full investigation)

Two confirmed marker-loss occurrences of the same worktree-cleanup failure shape in one day. The five hooks without the fallback — `pr-merge-reminder.py`, `post-merge-tile-checkpoint.py`, `usage-snapshot.py`, `post-pr-merge-pull.py`, `post-pr-merge-reclaim.py` — silently drop their action (a journal reminder, a tile-checkpoint reminder, a usage snapshot, a local-`main` fast-forward, or a `node_modules` reclaim) every time this recurs.

## Scope

Per PR #497's own reasoning, a naive per-hook rollout means up to six redundant `gh pr view` calls for a single failed-merge event, since each hook is an independent PostToolUse process with no natural place to share one result. Worth deciding explicitly rather than defaulting to "just copy the post-pr-merge-project.py pattern five times":

- Share one `gh pr view` result across all six hook invocations for the same merge event (would need some kind of short-lived cache keyed on PR/commit — added complexity), vs.
- Accept the redundant calls (simpler, and `gh pr view` is cheap/read-only) and wire `should_confirm_via_gh` / `confirm_merge_via_gh` into the other five directly, matching `post-pr-merge-project.py`'s pattern.
- For `post-pr-merge-pull.py` / `post-pr-merge-reclaim.py` specifically, PR #497 argued they already have eventual-consistency backstops (`dev-env-sync` on next prompt, ADR-058; the 6-hourly reclaim routine, ADR-037) — worth re-examining whether that's sufficient justification to leave them out of scope, or whether closing the immediate gap is still worth it.

## Related

- #489 (original bug + fix, `post-pr-merge-project.py` only)
- #496 (this failure shape recurring for PR #491, before #489's fix even landed — full root-cause writeup)
- #498 (open question: did #489's own fix actually fire correctly on its first live trigger, or did native GitHub Projects automation mask the result the same way ADR-050 documents for #369)
- [ADR-050](https://github.com/brownm09/dev-env/blob/main/docs/adr/050-shared-hookio-sibling-hook-fixes.md), Amendment 3

brownm09:
## Summary

- Wires `should_confirm_via_gh` / `confirm_merge_via_gh` (dev-env#489's live `gh pr view` fallback, wired into `post-pr-merge-project.py` only) into the remaining five marker-gated PostToolUse hooks: `usage-snapshot.py`, `post-merge-tile-checkpoint.py`, `post-pr-merge-pull.py`, `post-pr-merge-reclaim.py`, `pr-merge-reminder.py`.
- Motivated by career-playbook PR #635 hitting the identical worktree-cleanup failure shape (`fatal: 'main' is already checked out`) with zero hook output — the first occurrence outside dev-env itself, confirming the gap is a property of the shared global hook architecture (these hooks fire for every repo), not dev-env-repo-specific.
- `pr-merge-reminder.py`'s `_build_messages()` is directly unit-tested with ~20 synthetic inputs, several of which have a non-zero exit code and no marker. A naive wiring would have made those existing tests shell out to a real `gh pr view` call (one stayed accidentally safe only because its synthetic cwd doesn't exist on disk). Instead, the live check is resolved in `main()` and passed in via a new `live_confirmed: bool | None = None` parameter — default `None` reproduces every pre-existing test's behavior exactly; the function itself never shells out.
- Per dev-env#504's own "Scope" section, this takes the simpler of the two named options: accepting up to six independent (redundant) `gh pr view` calls per failed-merge event rather than building a cross-hook shared-result cache.
- Adds [ADR-050 Amendment 8](docs/adr/050-shared-hookio-sibling-hook-fixes.md) documenting the rollout, and updates the `docs/REFERENCE.md` "Secondary effect" note to reflect the completed coverage (dev-env#498's separate "did the original fallback actually fire" question remains open and is untouched by this PR).

Closes #504

## Test plan

- [x] `py -3 -c "import ast,sys; ..."` syntax check on all 6 changed `.py` files
- [x] `py -3 claude/scripts/tests/test_usage_snapshot.py` — 11 passed
- [x] `py -3 claude/scripts/tests/test_post_merge_tile_checkpoint.py` — 4 passed
- [x] `py -3 claude/scripts/tests/test_post_pr_merge_pull.py` — 9 passed
- [x] `py -3 claude/scripts/tests/test_post_pr_merge_reclaim.py` — 4 passed
- [x] `py -3 claude/scripts/tests/test_pr_merge_reminder.py` — 45 passed (3 new: `live_confirmed=True/False/omitted`)
- [x] `py -3 claude/scripts/tests/test_hookio.py` — 44 passed (sanity check, unchanged file)
- [x] `py -3 claude/scripts/tests/test_post_pr_merge_project.py` — 16 passed (sanity check, unchanged file)
- [x] Suppression + test-integrity pre-PR greps — clean, no matches

Tests: 133 passed, 0 skipped, 0 failed (across all 8 files run)

The live `confirm_merge_via_gh` calls themselves remain untested in all 6 hooks, per this repo's established convention (it shells out to `gh pr view`; the repo avoids subprocess mocks) — same boundary `post-pr-merge-project.py`'s own test suite already leaves uncovered.

- **[maintainability] Dead `merge_pr_number_from_output` call in `usage-snapshot.py`** — fixed in `4b656cc`.
- **[reliability] Cumulative timeout exposure across up to 6 `gh pr view` calls** — accepted, no code change. Deliberate tradeoff named in dev-env#504's own "Scope" section (redundant calls vs. a shared-result cache) and re-confirmed explicitly this session. Worth revisiting only if cumulative latency becomes visible in practice.
- **[reliability] Crude substring merge-detection gate in 3 hooks widens a pre-existing false-positive's cost** — filed [dev-env#529](https://github.com/brownm09/dev-env/issues/529), out of scope for this PR (touches each file's existing tested predicate; needs its own test coverage).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -389,8 +389,13 @@ the colliding item(s) to the next free number, and re-run `gh pr merge`.
     `cd <repo> && git push` target — so a cross-repo push is evaluated against THAT repo, not the
     session cwd — and falls back to cwd for a bare push; and that `_is_successful_merge_call` gates
     solely on gh's success marker, not the exit code, so an exit-0 non-merge invocation like
-    `gh pr merge --help` no longer fires the reminder (dev-env#485). The live `_open_pr_for_cwd`
-    subprocess boundary is not covered (the repo avoids subprocess mocks).
+    `gh pr merge --help` no longer fires the reminder (dev-env#485). Also exercises `_build_messages`'s
+    `live_confirmed: bool | None` parameter (dev-env#504 — the live `gh pr view` fallback for when the
+    marker itself is lost, dev-env#489): `True` overrides a marker-less `merge_ok` to fire the merge
+    reminder, `False` leaves it unfired, and the default `None` (every pre-existing call in this suite)
+    reproduces the exact pre-#504 marker-only behavior — `main()`, not `_build_messages`, decides
+    whether to attempt the live call, so this suite never shells out. The live `_open_pr_for_cwd` and
+    `confirm_merge_via_gh` subprocess boundaries are not covered (the repo avoids subprocess mocks).
 
     ```bash
     py -3 claude/scripts/tests/test_pr_merge_reminder.py
--- a/claude/scripts/post-merge-tile-checkpoint.py
+++ b/claude/scripts/post-merge-tile-checkpoint.py
@@ -21,13 +21,20 @@ Stdin JSON shape (PostToolUse):
 Output is read via _hookio.read_command_output (stdout+stderr, legacy output
 fallback) per ADR-049/ADR-050 — do NOT read tool_response.output directly.
 
-Exit 0 — not a successful `gh pr merge` call; no action.
+Exit 0 — not a successful `gh pr merge` call (including a genuinely-failed
+         merge the live `gh pr view` fallback also could not confirm); no action.
 Exit 2 — successful merge detected; tile-checkpoint reminder emitted via stderr.
 """
 import json
 import sys
 
-from _hookio import output_has_merge_marker, read_command_output
+from _hookio import (
+    confirm_merge_via_gh,
+    effective_merge_dir,
+    output_has_merge_marker,
+    read_command_output,
+    should_confirm_via_gh,
+)
 
 
 def is_successful_merge(command: str, output: str) -> bool:
@@ -60,9 +67,20 @@ def main() -> None:
 
     command = data.get("tool_input", {}).get("command", "")
     output = read_command_output(data)
+    cwd = data.get("cwd", "")
 
     if not is_successful_merge(command, output):
-        sys.exit(0)
+        # gh's marker does not always survive to this hook's captured output
+        # when gh exits abruptly right after a worktree's local-cleanup
+        # failure (dev-env#489) — fall back to a live `gh pr view` confirmation
+        # rather than silently dropping the tile-checkpoint reminder (dev-env#504).
+        if "gh pr merge" not in command:
+            sys.exit(0)
+        exit_code = data.get("tool_response", {}).get("exitCode", -1)
+        if not should_confirm_via_gh(exit_code, output):
+            sys.exit(0)
+        if confirm_merge_via_gh(None, "", effective_merge_dir(command, cwd)) is None:
+            sys.exit(0)
 
     print(
         "[tile-checkpoint] PR merged — spawn follow-up tiles now via spawn_task for "
--- a/claude/scripts/post-pr-merge-pull.py
+++ b/claude/scripts/post-pr-merge-pull.py
@@ -28,7 +28,13 @@ import re
 import subprocess
 import sys
 
-from _hookio import effective_merge_dir, output_has_merge_marker, read_command_output
+from _hookio import (
+    confirm_merge_via_gh,
+    effective_merge_dir,
+    output_has_merge_marker,
+    read_command_output,
+    should_confirm_via_gh,
+)
 from _worktree_topology import canonical_on_main, merge_park_target, parse_worktree_porcelain
 
 # Map GitHub repo slugs to local clone paths.
@@ -231,7 +237,17 @@ def main() -> None:
     cwd = data.get("cwd", "")
 
     if not is_successful_merge(command, output):
-        sys.exit(0)
+        # gh's marker does not always survive to this hook's captured output
+        # when gh exits abruptly right after a worktree's local-cleanup
+        # failure (dev-env#489) — fall back to a live `gh pr view` confirmation
+        # rather than silently skipping the local-main fast-forward (dev-env#504).
+        if "gh pr merge" not in command:
+            sys.exit(0)
+        exit_code = data.get("tool_response", {}).get("exitCode", -1)
+        if not should_confirm_via_gh(exit_code, output):
+            sys.exit(0)
+        if confirm_merge_via_gh(None, "", effective_merge_dir(command, cwd)) is None:
+            sys.exit(0)
 
     repo = extract_repo(command, cwd)
     if not repo:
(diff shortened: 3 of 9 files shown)


brownm09:
## Review: fix(hooks): roll out confirm_merge_via_gh fallback to the 5 remaining marker-gated hooks

### Summary
Ready to merge with minor changes. The rollout itself is sound — four of the five hooks add the live-confirmation fallback only in `main()`, never inside their existing directly-tested marker predicates, and `pr-merge-reminder.py`'s new `live_confirmed` parameter is a careful, well-justified design to keep its `_build_messages()` unit tests from accidentally shelling out to a real `gh pr view` call. The one real finding is a small dead-code issue in `usage-snapshot.py`: it extracts a PR number from `output` in a branch where that extraction is structurally guaranteed to return `None`. Two further non-blocking observations are worth recording but don't need code changes in this PR.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/usage-snapshot.py` (main(), the new fallback block)
`pr_number = merge_pr_number_from_output(output)` is called in the branch where `merge_confirmed(command, output)` has already returned `False` due to a missing marker (the `scan_top_level` check above it already excludes the "not a merge command" case). `merge_confirmed` and `merge_pr_number_from_output` both scan `output` against the identical `_MERGE_MARKER_RE` regex in `_hookio.py` — if one finds no match on any line, the other cannot either. So this line always evaluates to `None` here; the subsequent `confirm_merge_via_gh(pr_number, ...)` call is equivalent to always passing `None` explicitly, same as the other four hooks. Not a bug (passing `None` is exactly the correct fallback — `gh pr view` infers from cwd), but a future reader will reasonably assume this extraction does something.
**Suggestion:** Replace with a literal `None` (matching `post-merge-tile-checkpoint.py`/`post-pr-merge-pull.py`/`post-pr-merge-reclaim.py`/`pr-merge-reminder.py`'s pattern) and drop the now-unused `merge_pr_number_from_output` import.

**[reliability]** All six changed hook files, `confirm_merge_via_gh` call sites
Per dev-env#504's own "Scope" section, this PR deliberately accepts up to six independent `gh pr view` calls (each with a 15s subprocess timeout in `_hookio.py`) for the same failed-merge event, rather than a shared-result cache. That tradeoff was explicitly re-confirmed this session. Worth a quick sanity check once this is live: if PostToolUse hooks for the same event execute sequentially rather than concurrently, a slow/hanging `gh` could compound across all six hooks. Not a reason to hold this PR — just worth watching in practice, since this PR is what makes the six-call scenario real for the first time (previously only one hook paid it).
**Suggestion:** If cumulative latency becomes visible in practice, revisit the shared-cache option #504 named as the alternative.

**[reliability]** `claude/scripts/post-merge-tile-checkpoint.py`, `claude/scripts/post-pr-merge-pull.py`, `claude/scripts/post-pr-merge-reclaim.py` — `is_successful_merge()`'s pre-existing `"gh pr merge" not in command` substring gate
This crude substring check (not `scan_top_level`-anchored, unlike `usage-snapshot.py`/`pr-merge-reminder.py`) predates this PR and isn't changed here. Before this PR, a false substring match (e.g., "gh pr merge" appearing in a heredoc body or quoted string) combined with a non-zero exit code and no marker was harmless — the hook just exited 0. After this PR, that same false match now pays a live `gh pr view` subprocess call before exiting 0. The false-positive frequency is unchanged; only the cost of a false positive went from free to a network round-trip. These three files were apparently missed by the `scan_top_level` convergence sweeps in ADR-050 Amendments 5/6 (which covered `pr-merge-reminder.py`, `post-tool-use.py`, `usage-snapshot.py`, `post-pr-merge-project.py`) — consistent with Amendment 7's own "a sweep is only as complete as the list it started from" lesson.
**Suggestion:** File a follow-up issue to converge these three onto `scan_top_level`, matching the pattern already established for the other three hooks — out of scope for this PR (it would touch each file's existing tested predicate function, which is a separately-scoped change).

---

### Findings Disposition
To be completed in the PR body before merge:
- Dead-code `merge_pr_number_from_output` call in `usage-snapshot.py` — fix in this PR.
- Cumulative timeout exposure — accept with rationale (deliberate tradeoff, already chosen twice: in dev-env#504's own scope discussion and re-confirmed this session).
- Crude substring gate in 3 hooks — file a follow-up issue, out of scope for this PR.

---

PR #531, opened 2026-07-03
fix(hooks): live-fetch project-hook single-select options, correct gitignored-cache docs

### LINKED ISSUE #527: [bug] project-hook's gitignored hook-config.json epic_options silently drifts after Epic field mutations
## Problem

\`~/.claude/scripts/post-tool-use.py\` (the \`gh issue create\`/\`gh pr create\` project-auto-add hook) reads its Epic-field option IDs from a **gitignored, machine-local** \`.claude/hook-config.json\` in the target project repo (see docstring at the top of the script). It does not query the GitHub Projects GraphQL API live — it trusts whatever is cached in that file.

Found in lifting-logbook: after \`gh issue create\` for #637, the hook printed 6 Epic options with IDs that matched *none* of the live GraphQL field state, and was missing 4 known epics entirely (Observability, API Implementation, Client Applications, Operations). Direct \`gh api graphql\` query against the live project confirmed the actual field has all 10 options with IDs matching lifting-logbook's \`CLAUDE.md\` table exactly — the *live* project data was never corrupted. Only the local hook-config.json cache was stale.

## Root cause

The dates line up with lifting-logbook's documented 2026-05-08 Epic-field mutation incident (CLAUDE.md: \"an Observability epic addition wiped assignments on all 89 project items\", recovery precedent #203, backup files dated 2026-05-10 in \`.claude/backups/\`). The project-level **Backup-and-restore procedure** for single-select option mutations (documented in both the global \`CLAUDE.md\` → \"GitHub Projects — single-select option mutation hazard\" section and lifting-logbook's own \`CLAUDE.md\`) has a step to update the tracked \`CLAUDE.md\` option-ID table in the recovery PR — but no step to refresh the **gitignored** \`.claude/hook-config.json\`, since it isn't part of any PR. It silently kept serving pre-mutation IDs and a stale, truncated option list from that point forward, with no git history to signal the drift.

## Impact

Any session that trusted the hook's printed \`<option-id>\` values for a field-edit command would either get an outright GraphQL error (wrong ID) or, worse, could silently mis-tag an item if a stale ID happened to still validate against some other field/option. In this instance it was caught only because the target epic ("API Implementation") wasn't in the stale list at all, forcing a live-state check before acting — a differently-scoped edit could have gone through silently wrong.

## Suggested fix

One or both of:
1. Add a step to the Backup-and-restore procedure (both global and per-project CLAUDE.md copies) to regenerate every onboarded project's \`.claude/hook-config.json\` \`epic_options\` (and any other cached option-ID maps) immediately after a single-select field mutation, alongside the CLAUDE.md table update.
2. Better: change \`post-tool-use.py\` to fetch the field's current options live via \`gh api graphql\` (or at least detect option-ID validity) instead of trusting a cache that has no mechanism to invalidate itself. This removes the drift vector entirely rather than relying on procedure discipline.

## Immediate remediation

lifting-logbook's \`.claude/hook-config.json\` \`epic_options\` has been manually corrected to match the live GraphQL state (all 10 current option IDs) as part of the session that found this. Other onboarded projects using this hook pattern have not been checked and may have the same drift.

brownm09:
## Summary

- `post-tool-use.py`'s `single_select` field options (Epic, Impact, etc.) were rendered entirely from cached `hook-config.json`, with no live-fetch and no drift detection. Every GitHub Projects single-select option mutation regenerates all option IDs for that field, and this cache had no way to notice. It drifted silently at least twice before this fix (lifting-logbook#637 discovery, lifting-logbook#628's local fix, and an independent rediscovery noted in this issue).
- `post-tool-use.py` now live-fetches a `single_select` field's current options via `gh api graphql` at reminder time, using the cached value as a fallback **only** when the live call fails — and labels the reminder `(live)` or `(cached — live fetch failed; may be stale)` so staleness is visible instead of silent. See [ADR-076](docs/adr/076-live-fetch-project-hook-single-select-options.md) for the full design rationale, including why fallback-on-failure (not blocking) and why replace-not-just-warn.
- The global `claude/CLAUDE.md` single-select mutation procedure now explicitly names `.claude/hook-config.json` / `.claude/propose.json` as caches to refresh after any option mutation — lifting-logbook#628 only patched lifting-logbook's own local CLAUDE.md copy; this generalizes it so every project inherits the safeguard.
- Corrects the "hook-config.json is gitignored and machine-local" claim, found repeated in 8 places (`post-tool-use.py`'s docstring, ADR-023, `_worktree_canon.py`, `reconcile-project-board.py`, `test_post_tool_use.py`, `docs/REFERENCE.md` ×2, dev-env's own `CLAUDE.md`) plus an `## Amendment` on ADR-052. That's dev-env's own `.gitignore` convention, not universal — lifting-logbook deliberately tracks the file in git.

Closes #527

## Design notes

- `field_id` is passed to `gh api graphql` as a `-f` variable, never string-interpolated into the query — no injection surface.
- Live fetch only runs after `add_to_project` already returned a real `item_id` (a failed create never pays for it), and only for `single_select` fields with a `field_id` — `milestone`-type fields and `.claude/propose.json`'s `epics` are explicitly out of scope (no live-fetch equivalent exists for either; the CLAUDE.md procedure remains their only safeguard).
- `format_reminder` gained an optional `live_options` kwarg defaulting to `None`, which reproduces the exact pre-fix output — every existing call site is unaffected.
- Verified the real subprocess path by hand against dev-env's own live Impact field during development (`fetch_live_field_options('PVTSSF_lAHOAjEKvM4BWKFezhRgkNc')` returned the live `{High, Medium, Low}` IDs, matching the cached CLAUDE.md table exactly — no drift currently present on dev-env's own project — and an invalid field ID correctly degraded to `None`).

## Testing

Tests: 56 passed, 0 skipped, 0 failed (`test_post_tool_use.py`, includes 17 new cases after the /review fix commit) — 3.2s
Tests: 8 passed, 0 skipped, 0 failed (`test_worktree_canon.py`, sanity re-run, docstring-only touch) — 0.4s
Tests: 26 passed, 0 skipped, 0 failed (`test_reconcile_project_board.py`, sanity re-run, docstring-only touch) — 1.2s

- [x] `py -3 -c "import ast,sys; ..." claude/scripts/*.py` — all scripts parse
- [x] `py -3 claude/scripts/tests/test_post_tool_use.py` — required (item 12), primary coverage for this change
- [x] `py -3 claude/scripts/tests/test_worktree_canon.py` — sanity (docstring-only edit)
- [x] `py -3 claude/scripts/tests/test_reconcile_project_board.py` — sanity (docstring-only edit)
- [x] Suppression grep (`git diff origin/main | grep -E '...'`) — 1 match, **false positive**: `+query($id: ID!) {` is GraphQL's non-null type syntax inside the new `_FIELD_OPTIONS_QUERY` string literal (`ID!` followed by `)`), not a TS/JS non-null assertion, `?? null`, `ts-ignore`, or `eslint-disable`. No actual suppression was added.
- [x] Test-integrity grep — clean, no matches.
- [x] `baseline_test_failure_tracking` — not applicable; this worktree has no `.claude/hook-config.json` (gitignored, as expected — dev-env's own project doesn't opt into the Jest-only baseline tracking anyway).

## ADR

New: [ADR-076](docs/adr/076-live-fetch-project-hook-single-select-options.md) — this change touches hook behavior (`post-tool-use.py`) and sets a new global `claude/CLAUDE.md` procedure step, both qualifying per the ADR-warrant check. Amended: [ADR-052](docs/adr/052-worktree-config-canonical-fallback.md) (terminology correction, doesn't change its Decision).

`/review` posted: [comment](https://github.com/brownm09/dev-env/pull/531#issuecomment-4871791591). 0 blocking, 3 non-blocking — all fixed in [f96942e](https://github.com/brownm09/dev-env/commit/f96942ec418bca7357da6e9d1a1bbe9cfde43923):

- **[performance]** Serial per-field live-fetch timeout is multiplicative as `required_fields` grows — fixed: documented the multiplicative cost in `fetch_live_required_field_options`'s docstring (no restructure needed; both onboarded projects have exactly one single_select field today).
- **[reliability]** `fetch_live_field_options` promised "never raises" but its except tuple missed `UnicodeDecodeError`, reachable because the explicit `encoding="utf-8"` opts out of `_winsubp`'s `errors="replace"` default — fixed: added `UnicodeDecodeError` to the except tuple.
- **[maintainability]** `_resolve_required_fields(config)` was called twice per reminder (`main()` and again inside `format_reminder`) — fixed: `format_reminder` gained an optional `required_fields` param (default `None` re-derives from config, unaffected); `main()` now resolves once and threads the same list into both `fetch_live_required_field_options` and `format_reminder`.

Also dismissed as false positives during review (not fixed, nothing to fix): a claim that `subprocess`/`json` might not be imported (verified both are pre-existing top-of-file imports), and a stdin-handling question (verified consistent with the sibling `add_to_project`/`canonical_root_via_git` calls, neither of which sets `stdin=DEVNULL` either).


brownm09:
## Review: fix(hooks): live-fetch project-hook single-select options, correct gitignored-cache docs

### Summary
Ready to merge with minor changes. No blocking findings — no correctness or security defects. This PR is well-scoped, well-tested (16 new tests, 55/55 passing, plus two sanity re-runs on adjacent test suites for docstring-only touches), and the design (fallback-on-failure, replace-not-just-warn, byte-identical default behavior) is sound and documented in ADR-076. Three non-blocking findings, all hardening/maintainability items rather than defects: a serial-timeout scaling note, a narrow exception-handling gap relative to the new function's own "never raises" docstring contract, and a redundant double-resolution of `_resolve_required_fields` per reminder. Two independent review passes converged on the same double-resolution observation, which is a good signal it's worth addressing. The diff (455 lines, 11 files, 9 of them docs/comment-only) is appropriately scoped for a fix this size — no split needed.

One correctness claim from automated analysis was checked and dismissed as a false positive: a concern that `subprocess`/`json` might not be imported (would cause `NameError` on every hook fire) — verified false by direct inspection (`import json` / `import subprocess` are pre-existing, unchanged lines 49/52) and by the fact that the full test suite already executes this code successfully. A stdin-handling concern was also checked and dismissed: `fetch_live_field_options` doesn't set `stdin=DEVNULL`, but neither do the sibling `add_to_project` or `canonical_root_via_git` calls it's modeled on — consistent with existing convention, not a new risk.

---

### Non-Blocking Findings

**[performance]** `claude/scripts/post-tool-use.py` (`fetch_live_required_field_options`)
The live-options fetch loop is serial — one synchronous `gh api graphql` call per `single_select` field, each with `timeout=10`. Both onboarded projects today have exactly one such field, so worst-case added latency is ~10s, matching the ADR's "one more bounded call" framing. But the cost is multiplicative as a project's `required_fields` grows: a hypothetical project with N single-select fields and a hung `gh` call blocks the session for up to N×10s before the reminder prints, since this is a synchronous PostToolUse hook.
**Suggestion:** Not worth restructuring now (no project has N>1 today), but consider noting the multiplicative-per-field cost explicitly in `fetch_live_required_field_options`'s docstring so a future config author (or a future contributor extending this function) doesn't add several single-select fields without realizing the serial cost compounds.

**[reliability]** `claude/scripts/post-tool-use.py` (`fetch_live_field_options`)
The function's docstring promises "Never raises," but its `except (FileNotFoundError, subprocess.TimeoutExpired, OSError)` doesn't include `UnicodeDecodeError`. The call passes `encoding="utf-8"` explicitly without `errors=`, which — per this repo's own `_winsubp` contract ("an explicit `encoding=`... from the caller is never overridden") — means the `_winsubp` patch's `errors="replace"` safety net does *not* apply here, since the caller already supplied its own `encoding=`. If `gh` ever emitted a non-UTF-8 byte on stdout (GitHub's API is UTF-8 JSON in practice, so this is a narrow/low-probability path), the `UnicodeDecodeError` would propagate uncaught out of this function, then out of `fetch_live_required_field_options`'s loop (which has no per-field try/except of its own), aborting the *entire* live-fetch pass for every remaining field in that reminder — caught only by `main()`'s outer safe-exit guard, which silently drops the reminder. That's exactly the "silently dropped reminder" failure class dev-env#527 exists to close, just relocated rather than eliminated. Note: the pre-existing `canonical_root_via_git` has an identical narrow gap (out of scope for this PR, not touched by this diff, and its docstring doesn't make as strong a "never raises" claim), and `add_to_project` in `_gh_project.py` is already more defensive here (bare `except Exception`).
**Suggestion:** Add `UnicodeDecodeError` to the except tuple (or pass `errors="replace"` explicitly alongside `encoding="utf-8"`) so the function's own documented "never raises" contract holds unconditionally, and one field's decode failure can't drop every other field's live data in the same reminder.

**[maintainability]** `claude/scripts/post-tool-use.py` (`main()` / `format_reminder`)
`_resolve_required_fields(config)` is called twice per reminder: once in `main()` (to discover live-fetch targets) and again inside `format_reminder` (to render). Both are pure and deterministic today, so this isn't a live bug, but the invariant "both call sites see identically-resolved fields" is maintained by re-derivation rather than by construction — a future change to `_resolve_required_fields` (e.g., any source of non-determinism, or a change gated on external state) could make the two calls silently diverge, and nothing in the code would catch it. Independently flagged by both review passes.
**Suggestion:** Resolve once in `main()` and thread the already-resolved `required_fields` list into `format_reminder` as a parameter (or have `format_reminder` accept the pre-resolved list instead of the raw `config`), so there's a single resolution per reminder by construction.

---

### Findings Disposition
All three non-blocking findings above are being fixed in this PR before merge (see follow-up commit).

---

PR #533, opened 2026-07-03
fix(hooks): converge stub-push-archive-reminder.py onto scan_top_level

brownm09:
## Summary

- Converges `stub-push-archive-reminder.py`'s inline "is this a git push?" check in `main()` from the crude `"git push" not in command` substring test onto `scan_top_level`-anchored detection — the pattern already established by `usage-snapshot.py` / `pr-merge-reminder.py` / `post-pr-merge-project.py` / `post-merge-tile-checkpoint.py` / `post-pr-merge-pull.py` / `post-pr-merge-reclaim.py`. This is the file PR #530's own description and ADR-050 Amendment 9's "General lesson" both explicitly named as the one remaining out-of-scope case for #529.
- Adds a local `_PUSH_RE` / `_check_push_stmt` pair, identical to the one `pr-merge-reminder.py` already defines for its own `git push` detection — a port, not new regex design.
- Unlike the three files #530 fixed, this file's check had no separate pre-existing testable predicate (it lived inline in `main()`). Extracts a new `is_git_push_command(command)` pure function — mirroring `is_successful_merge()`'s role in the other hooks — so the new heredoc/quote/subshell coverage has something pure to call, consistent with the repo's `main()`-I/O-is-untested convention.
- Closes the false-positive gap: `git push` text inside a heredoc body, a quoted argument, or a `$()` subshell no longer counts as an invocation.
- Documents the convergence as **ADR-050 Amendment 10**, including a repo-wide grep confirming the pre-fix state (only this file's check remained; the `"gh pr create"` and `"gh pr merge"` analogs were already at zero matches).
- Updates the `## Testing` item #16 description in `CLAUDE.md` to note the new `is_git_push_command()` coverage, mirroring #530's item #10/15/23 update style.

## Test plan

- [x] `py -3 -c "import ast,sys; ..."` syntax check over every `claude/scripts/*.py` — clean
- [x] `py -3 claude/scripts/tests/test_stub_push_archive_reminder.py` — **Tests: 10 passed, 0 skipped, 0 failed** (5 pre-existing `has_push_error()` cases + 5 `is_git_push_command()` cases: bare-push and cd-prefixed-push sanity baselines plus the 3 heredoc/quote/subshell false-positive shapes)
[...]

PR #530, opened 2026-07-02
fix(hooks): converge post-merge-tile-checkpoint/pull/reclaim onto scan_top_level

brownm09:
## Summary

- Converges `is_successful_merge()`'s command-shape check in `post-merge-tile-checkpoint.py`, `post-pr-merge-pull.py`, and `post-pr-merge-reclaim.py` from the crude `"gh pr merge" not in command` substring test onto `scan_top_level(command, _check_merge_stmt)` — the anchored predicate already used by `usage-snapshot.py` / `pr-merge-reminder.py` / `post-pr-merge-project.py` (ADR-050 Amendments 5/6). These 3 files were missed by both prior convergence sweeps, per #529.
- Adds a local `_MERGE_RE` / `_check_merge_stmt` pair to each file, identical to the established pattern. Both occurrences of the old check — inside `is_successful_merge()` and inside `main()`'s live-`gh pr view`-fallback gate — are replaced.
- Closes the false-positive gap: `gh pr merge` text inside a heredoc body, a quoted argument, or a `$()` subshell no longer counts as an invocation. Since PR #528 wired `confirm_merge_via_gh` into these hooks, a false match on this shape paid a real `gh pr view` network round-trip before exiting 0.
- Documents the convergence as **ADR-050 Amendment 9**, including a false positive I live-reproduced during this session's own testing: my scratch verification script's heredoc fixture data (containing literal `"gh pr merge --squash --delete-branch"` text as test input) false-triggered the *canonical*, not-yet-fixed `post-merge-tile-checkpoint.py` hook mid-session.
- Lightly updates the 3 corresponding `## Testing` item descriptions in `CLAUDE.md` (items #10, #15, #23) to note the anchored check, mirroring item #24's existing wording style.

## Test plan

- [x] `py -3 -c "import ast,sys; ..."` syntax check over every `claude/scripts/*.py` — clean
- [x] All 32 test files under `claude/scripts/tests/` run individually — **Tests: 511 passed, 0 skipped, 0 failed**
- [x] 9 new heredoc/quote/subshell regression tests added (3 per file) to the 3 target test suites, each verified directly against the extracted pre-fix predicate logic to confirm they reproduce the old bug (`True`) — not vacuously-true assertions
- [x] Repo-wide grep confirms zero remaining `"gh pr merge" not in command` occurrences in `claude/scripts/*.py` (the one remaining similarly-shaped check, `stub-push-archive-reminder.py`'s unrelated `git push` check, is out of scope for #529 and left untouched)

[...]

PR #536, opened 2026-07-03
fix(hooks): automate ADR-050 Amendment 9's crude-command-substring-check grep into a regression test

### LINKED ISSUE #534: Automate ADR-050 Amendment 9's crude-command-substring-check grep into a regression test
## Problem

ADR-050 (`docs/adr/050-shared-hookio-sibling-hook-fixes.md`) documents a recurring bug class: a crude, unanchored substring check like `if "gh pr merge" not in command: return False` used to detect whether a raw Bash command string invoked a specific CLI command. It false-positives when the literal text appears inside a heredoc body, a quoted argument, or a `$()` subshell rather than as a real top-level statement. The fix (`_hookio.scan_top_level`) was converged across Amendments 5, 6, and 9 — each sweep missed at least one sibling hook because it scoped itself conceptually ("which hooks need this kind of detection") rather than textually.

Amendment 9's closing "General lesson" (dev-env#529 / PR #530) proposes a durable mechanical proxy:

> grep the whole `claude/scripts/` tree for the literal string `"gh pr merge" not in command` (and its `"gh pr create"` / `"git push"` analogs...) before declaring any future `scan_top_level` convergence sweep complete

This is currently just prose in the ADR — a manual step someone has to remember to run. It has already been the actual gap three sweeps running.

## Fix

Add an automated test (`claude/scripts/tests/test_no_crude_command_substring_checks.py`) that AST-parses every top-level `claude/scripts/*.py` file (excluding `tests/`) and fails on any live `<string literal> {in|not in} command`-shaped `Compare` node — generalized beyond the ADR's 3 named literals to the general shape, since an AST walk costs no more to make generic. AST is used specifically because it makes comments and docstrings *structurally* invisible (not just filtered), which is the exact "live check vs. explanatory comment" distinction PR #530's own diff had to navigate by hand via careful grep phrasing.

## Complications found while scoping (before writing any code)

1. `claude/scripts/stub-push-archive-reminder.py` currently has **three** live matches, not the one the ADR names: the documented `"git push" not in command` (line 82), plus **two previously-untracked** checks of the identical shape — `"engineering-journal"` / `"engineering_journal" not in command` (line 86), checking for a repo-name reference rather than a CLI invocation. Neither has been named by any prior ADR or issue. A hard zero-tolerance test would fail on introduction, so this needs a small, auditable, two-sided-checked exceptions list (fails on both a new unlisted offense and a stale listed one whose underlying code has since been fixed).
2. **PR #533** (closes #532) is already in flight, converging exactly the line-82 `git push` check onto `scan_top_level`, and already claims **ADR-050 Amendment 10**. This issue's fix will *not* duplicate that work or its behavior change — it is test-and-documentation only. The `git push` exception is transitional (removed once #533 merges); any ADR addition from this issue is numbered Amendment 11, not 10.

## Scope

Test-only / tooling change — no hook runtime behavior is modified. The two `engineering-journal` checks are tracked as a follow-up (tile + issue) at merge time, not fixed here.

brownm09:
## Summary

- Adds `claude/scripts/tests/test_no_crude_command_substring_checks.py` — an AST-based repo-wide test that automates ADR-050 Amendment 9's closing "General lesson," which proposed a manual grep as a "durable mechanical proxy" before declaring any future `scan_top_level` convergence sweep complete. That proxy was never more than prose in the ADR; Amendment 10 (dev-env#532/PR #533) needed its own separate issue to close the one instance Amendment 9 had already named by hand, which is direct evidence the reminder alone wasn't durable enough.
- AST-based rather than grep-based specifically because a naive grep for the crude pattern's own text false-positives on the fix PRs' own explanatory comments (PR #530's diff quotes the literal pattern to describe the bug it replaced). `ast.parse` sidesteps this categorically: comments are stripped by the tokenizer before the parser ever sees them, and a docstring's string constant is never itself a `Compare` operand unless it's live code. One of the 11 self-tests (`test_ignores_pr530_style_explanatory_comment`) proves this directly against PR #530's own verbatim comment text.
- Generalizes beyond the ADR's three originally-named literals (`gh pr merge` / `gh pr create` / `git push`) to flag *any* string-literal `in`/`not in` check against a variable named `command` — an AST walk costs no more to make generic. Running the generalized detector surfaced **two previously-untracked instances** of the identical shape in `stub-push-archive-reminder.py`: `"engineering-journal"` / `"engineering_journal" not in command` (checking for a repo-name reference, not a CLI invocation). Neither has been named by any prior ADR or issue — not something a grep scoped to the three historical literals, including Amendment 10's own repo-wide grep, would have found.
- Both are carried in a small, two-sided-checked `_KNOWN_EXCEPTIONS` allowlist (matched on filename + literal text, not line number): the test fails on **either** a new unlisted offense **or** a stale listed exception whose underlying code has since been fixed, so the allowlist can't silently rot. This isn't hypothetical — dev-env#532/PR #533 converged `stub-push-archive-reminder.py`'s `"git push" not in command` check while this PR's test was being authored, which is exactly why that exception is *not* in the final allowlist (it would have gone stale within this PR's own lifetime had it been hardcoded).
- Converging the two `engineering-journal` checks onto `scan_top_level` is a hook-behavior change and is intentionally out of scope here — this PR is test-and-documentation only. Tracked as a follow-up tile + issue at merge time.
- Adds CLAUDE.md `## Testing` item 39 and documents the work as **ADR-050 Amendment 11**.

## Scope note

No hook runtime behavior is modified by this PR.

## Test plan

- [x] `py -3 -c "import ast,sys; ..."` syntax check over every `claude/scripts/*.py` — clean
- [x] `py -3 claude/scripts/tests/test_no_crude_command_substring_checks.py` — **Tests: 13 passed, 0 skipped, 0 failed** (10 detector self-tests + 2 allowlist-diff self-tests added during review + the repo-wide gate)
- [x] All 33 test files under `claude/scripts/tests/` run individually — **all pass, 0 failures**
- [x] Repo-wide detector run confirms exactly 2 live offenses today (both in the documented allowlist), 0 unexpected, 0 stale, 0 duplicated
- [x] Two-sided allowlist check independently verified to actually fire in both directions (an unlisted offense, and a stale exception), not just pass today by construction

No automated tests were skipped or degraded; no suppressions were added.

Closes #534

`/review` posted 0 blocking / 3 non-blocking findings (all narrow, in-scope precision/robustness gaps in the new test file itself — none affect the current, clean scan result). All 3 fixed in [`370d04d`](https://github.com/brownm09/dev-env/pull/536/commits/370d04dcc3bba4d39d71494ea5cb847e262499b4):

1. **[maintainability]** `_KNOWN_EXCEPTIONS` matched on `(filename, literal)` only, so a second, distinct offending call site reusing an already-allowlisted literal in the same file would silently absorb into the existing exception instead of surfacing as a new offense — **fixed in `370d04d`**, extracted `_diff_against_known_exceptions()` with occurrence-count tracking; an allowlist entry with more than one live occurrence now fails the gate.
2. **[reliability]** `_scan_real_repo()` had no error handling around `read_text()`/`ast.parse()` — `ast.parse`'s `filename=` kwarg already attributes a `SyntaxError` to the right file (verified empirically), but a `UnicodeDecodeError` from `read_text()` carries no filename of its own — **fixed in `370d04d`**, wrapped both in a narrow try/except that re-raises with the filename attached.
3. **[maintainability]** The module-level "Known non-goal" docstring paragraph named only the f-string exclusion, not the (already function-docstring-documented) bare-`Name`/literal-first-only detection scope — **fixed in `370d04d`**, extended the paragraph so a reader skimming just the module docstring doesn't read a clean run as "no crude substring checks of any shape."

Two new self-tests added alongside the fix (`test_diff_detects_duplicated_offense_behind_allowlist_entry`, `test_diff_single_occurrence_is_not_duplicated`), exercising the extracted helper directly via synthetic fixtures rather than depending on the real repo state.


brownm09:
## Review: fix(hooks): automate ADR-050 Amendment 9's crude-command-substring-check grep into a regression test

### Summary

Ready to merge. This converts ADR-050 Amendment 9's manual grep reminder into an enforced AST-based regression test — a genuinely stronger fix than what the ADR literally asked for (it generalizes beyond the three named literals to any `<literal> in/not in command` shape, which is what caught two previously-untracked `stub-push-archive-reminder.py` checks the ADR's own historical greps missed). I independently verified the diff: the hook-script syntax check is clean, the new test's 11 self-tests pass, and all 33 pre-existing test suites under `claude/scripts/tests/` pass unchanged. No blocking findings. Three non-blocking findings below, all narrow and already fixed in a follow-up commit on this branch (see disposition).

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/tests/test_no_crude_command_substring_checks.py:139-147` (`test_repo_has_no_unexpected_or_stale_command_substring_checks`)

`_KNOWN_EXCEPTIONS` matches on `(filename, literal)` only, not line number — a deliberate choice (per the code's own comment) so a line shift doesn't stop covering the real exception. But it also means if a *second, distinct* call site in the same file happens to reuse an already-allowlisted literal string (e.g. a new, unrelated `"engineering-journal" not in command` check added elsewhere in `stub-push-archive-reminder.py`), the set-based diff silently absorbs it into the existing exception entry — the two-sided unexpected/stale check can't distinguish "still exactly the one audited offense" from "now two offenses, one audited and one new."

**Suggestion:** Track occurrence counts per `(filename, literal)` and flag an allowlist entry whose live count exceeds 1, in addition to the existing unexpected/stale checks.

**[reliability]** `claude/scripts/tests/test_no_crude_command_substring_checks.py:110-120` (`_scan_real_repo`)

`path.read_text(encoding="utf-8")` has no error handling. `ast.parse`'s `filename=` kwarg already attributes a `SyntaxError` to the right file (verified: `str(e)` includes `(filename, line N)`), but a `UnicodeDecodeError` raised earlier, during `read_text()` itself, carries no filename — so a scan failure on a bad-encoding file would surface as a bare codec error with no indication of which of the 47 files in the glob caused it.

**Suggestion:** Wrap the read+parse in a narrow try/except that re-raises with the filename attached.

**[maintainability]** `claude/scripts/tests/test_no_crude_command_substring_checks.py:78-90` (`find_command_substring_checks` docstring)

The detector is correctly and deliberately scoped to one AST shape — a bare `Name` spelled exactly `command`, literal-first operand order — and this is already documented in the function's own docstring. But the module-level "Known non-goal" paragraph only calls out the f-string exclusion, not this (larger) scope boundary, so a reader skimming just the module docstring could read a clean run as "no crude substring checks of any shape" rather than "no crude substring checks in this specific shape" (a future hook using `self.command`, a differently-named alias, or `command.find(literal) == -1` would evade detection silently).

**Suggestion:** Extend the module-level "Known non-goal" paragraph to cross-reference this scope boundary, not just the f-string one.

---

### Findings Disposition

All three non-blocking findings above are fixed in this branch's follow-up commit (see PR body). No blocking findings, no filed issues needed.

---

PR #543, opened 2026-07-03
fix(hooks): converge stub-push-archive-reminder.py's engineering-journal checks onto scan_top_level

### LINKED ISSUE #539: Converge stub-push-archive-reminder.py's engineering-journal reference checks onto scan_top_level
## Problem

`claude/scripts/stub-push-archive-reminder.py`'s `main()` gates the archive-reminder sentinel on two crude, unanchored substring checks against the raw command string:

```python
# Must reference engineering-journal
if "engineering-journal" not in command and "engineering_journal" not in command:
    sys.exit(0)
```

This is the exact same bug class ADR-050 documents and Amendments 5/6/9/10 converged onto `scan_top_level`: the literal text `"engineering-journal"` (or `"engineering_journal"`) can appear inside a heredoc body, a quoted argument, or a `$()` subshell without the command actually being a push that targets that repo — the check would then incorrectly treat an unrelated push as engineering-journal-bound (or vice versa, depending on which side of the check misfires).

Unlike the three literals Amendment 9 named (`gh pr merge` / `gh pr create` / `git push`), these two checks aren't detecting a *CLI invocation* — they're checking whether the already-confirmed `git push` command's target references the engineering-journal repo by name. Same false-positive shape, different purpose.

## Discovery

Found by [dev-env#534](https://github.com/brownm09/dev-env/issues/534)'s AST-based regression test ([`test_no_crude_command_substring_checks.py`](https://github.com/brownm09/dev-env/blob/main/claude/scripts/tests/test_no_crude_command_substring_checks.py)), which generalizes beyond the ADR's three originally-named literals to flag *any* string-literal `in`/`not in` check against a variable named `command`. Neither check was named by any prior ADR text or issue — not something a grep scoped to the three historical literals (including Amendment 10's own repo-wide grep) would have found. Currently carried in that test's `_KNOWN_EXCEPTIONS` allowlist so the regression test passes against this known, not-yet-fixed debt:

```python
_KNOWN_EXCEPTIONS = {
    ("stub-push-archive-reminder.py", "engineering-journal"),
    ("stub-push-archive-reminder.py", "engineering_journal"),
}
```

Promised as a follow-up in [ADR-050 Amendment 11](https://github.com/brownm09/dev-env/blob/main/docs/adr/050-shared-hookio-sibling-hook-fixes.md#amendment-11-2026-07-02--automating-amendment-9s-durable-mechanical-grep-proxy-into-a-regression-test-dev-env534) ("The two `engineering-journal` checks are tracked via a `spawn_task` tile and a linked follow-up issue at merge time... not converged in this PR").

## Fix

Converge both checks onto `_hookio.scan_top_level`, mirroring the pattern Amendments 9/10 already established for the other six hooks in this family. Concretely:

- Add a local `_check_engineering_journal_ref` predicate (or extend the existing `is_git_push_command`-style helper) anchored via `scan_top_level`, replacing the inline `if "engineering-journal" not in command and "engineering_journal" not in command:` in `main()`.
- Add heredoc/quote/subshell regression tests to `test_stub_push_archive_reminder.py`, mirroring the three dev-env#499 false-positive shapes Amendments 9/10 already cover for the sibling checks in this same file.
- Once fixed, remove both entries from `_KNOWN_EXCEPTIONS` in `test_no_crude_command_substring_checks.py` — the test's two-sided check will fail loudly (as "stale") if the entries are left behind after the underlying code is fixed, so this step is self-enforcing.
- Document as ADR-050 Amendment 12.

## Scope note

Hook-behavior change (unlike #534/#536, which were test-and-documentation only) — needs its own PR and test coverage, not bundled into a docs/test-only change.

Prior art: #534, [PR #536](https://github.com/brownm09/dev-env/pull/536), ADR-050 Amendments 9/10/11.

brownm09:
## Summary

- `stub-push-archive-reminder.py`'s `main()` gated the journal archive-reminder sentinel on two crude, unanchored substring checks (`"engineering-journal" not in command and "engineering_journal" not in command`) — the same false-positive bug class ADR-050 Amendments 9/10 already fixed for this file's `git push` detection, where the literal text can appear inside a heredoc body, a quoted argument, or a `$()` subshell without the push actually targeting that repo.
- Adds `_EJ_REF_RE` / `_check_engineering_journal_ref` / `references_engineering_journal()`, anchored via the shared `_hookio.scan_top_level`. Unlike `is_git_push_command` (which anchors a CLI verb to a segment's start), this predicate anchors to a `cd <path>` / `git -C <path>` directory-argument **prefix**, since the target text is an argument *value* and never begins its own statement — documented in ADR-050 Amendment 12, including empirical proof that a naive "just scope the substring check to `scan_top_level` segments" fix would **not** have closed the gap.
- Converging both checks empties `test_no_crude_command_substring_checks.py`'s `_KNOWN_EXCEPTIONS` allowlist — its own two-sided gate confirms 0 unexpected / 0 stale / 0 duplicated across all 47 scanned files, so no crude command-substring check of this shape remains anywhere in `claude/scripts/*.py`.

## Test plan

- [x] `py -3 claude/scripts/tests/test_stub_push_archive_reminder.py` — 17 passed (10 pre-existing + 7 new: `cd`-prefix / `git -C`-prefix / underscore-spelling positive baselines, an unrelated-repo negative baseline, and the three dev-env#499 heredoc/quote/subshell false-positive shapes applied to this check)
- [x] `py -3 claude/scripts/tests/test_no_crude_command_substring_checks.py` — 13 passed; repo-wide gate reports "47 files ... 0 known exception(s), 0 unexpected, 0 stale, 0 duplicated"
- [x] Full `claude/scripts/tests/` suite — all 34 Python test files + 5 shell test files pass (hundreds of individual assertions; two shell tests self-skip as documented: `run-shellcheck.sh` when `shellcheck` isn't installed)
- [x] Hook-script syntax check (`ast.parse` over every `claude/scripts/*.py`) — clean

Tests: all passed, 0 skipped (except the documented self-skip in `run-shellcheck.sh`), 0 failed.

Closes #539

Reviewed: https://github.com/brownm09/dev-env/pull/543#issuecomment-4872265619

- **[documentation, blocking]** `docs/REFERENCE.md:225`'s Hooks-table entry named only `has_push_error` as a unit-tested helper, missing `is_git_push_command` (Amendment 10) and the new `references_engineering_journal` (Amendment 12). **Fixed in `d1e6a74`** — entry now names all three tested predicates, matching the `pr-merge-reminder.py` row's convention.
- **[maintainability, non-blocking]** `_EJ_REF_RE`'s `\S*` can't match a quoted `cd`/`git -C` path containing a space (unlike `pr-merge-reminder.py`'s quote-aware `_CD_CHAIN_RE`). **No code change** — not a regression (matches this file's existing `_PUSH_RE`/`_MERGE_RE`/`_CREATE_RE` limitation) and not reachable with this environment's actual no-space engineering-journal path. Noted for future reference if this file's `cd`-handling is ever generalized further.
- **[maintainability, non-blocking]** Language-scope note: Step 2e's automated skip-marker/deletion/threshold scanners target JS/TS, not Python. **No code change** — manually verified during review (this repo has no pytest, only a hand-rolled test runner; no test deleted, no threshold/bypass config touched) — this PR-body note is the record of that manual extension.


brownm09:
## Review: fix(hooks): converge stub-push-archive-reminder.py's engineering-journal checks onto scan_top_level

### Summary

Merge with a small in-PR fix. The core change is correct and well-verified: `references_engineering_journal()` correctly anchors detection to a `cd`/`git -C` directory-argument prefix rather than reusing the CLI-invocation anchoring pattern verbatim, and the ADR documents (with an empirical repro) exactly why the naive "just rescope the substring check to `scan_top_level` segments" alternative would not have worked. All 17 tests in `test_stub_push_archive_reminder.py` and all 13 in `test_no_crude_command_substring_checks.py` pass, and the repo-wide gate confirms 0 unexpected/stale/duplicated crude checks across all 47 scanned files. One blocking documentation gap: `docs/REFERENCE.md`'s Hooks-table entry for this file doesn't name either of its two tested pure predicates. One non-blocking robustness note on the new regex's handling of quoted paths with spaces (pre-existing pattern limitation elsewhere in the file, not a regression). DIFF_SIZE (230 lines) is appropriately scoped — single, focused fix plus its direct test/doc footprint, no unrelated changes.

One thing worth stating explicitly since it could misread on a fast diff pass: `test_no_crude_command_substring_checks.py`'s `_KNOWN_EXCEPTIONS` set shrinking from 2 entries to empty is the allowlist getting **stricter**, not a test being weakened — an allowlist entry represents a *permitted* known violation, so removing one means that violation is no longer tolerated. This is the opposite shape of an ADR-029 test-integrity violation, not an instance of one.

---

### Blocking Findings

**[documentation]** `docs/REFERENCE.md:225`
The Hooks-table entry for `stub-push-archive-reminder.py` says: *"The push-failure guard (`has_push_error`, reading output via the shared `_hookio.read_command_output`) is unit-tested by `tests/test_stub_push_archive_reminder.py`."* It names only `has_push_error` as a tested pure helper. `is_git_push_command()` (added by Amendment 10, already merged) and now `references_engineering_journal()` (this PR, Amendment 12) are also unit-tested pure predicates in the same file, but neither is mentioned — this entry has been stale since Amendment 10 and this PR makes it more so. Neighboring rows in the same table (e.g. `pr-merge-reminder.py`'s row: *"the pure `_effective_push_dir` and `effective_merge_dir` helpers are unit-tested by..."*) establish the convention of naming every tested pure helper, which this row doesn't follow. This falls under the CLAUDE.md doc-reconciliation checkpoint's "modify the behavior of ... a script" trigger (not just add/remove/rename) — README.md's own entry for this hook (line 85) is a high-level behavior summary that remains accurate and needs no change; only the more granular `docs/REFERENCE.md` row is affected.
**Fix:** Extend the `docs/REFERENCE.md:225` "What it does" cell to also name `is_git_push_command()` and `references_engineering_journal()` as unit-tested pure predicates, matching the `pr-merge-reminder.py` row's convention.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/stub-push-archive-reminder.py:68`
`_EJ_REF_RE = re.compile(r"(?:cd|git\s+-C)\s+\S*engineering[-_]journal\S*")` uses `\S*` for the path portion, which cannot match across a space — so a quoted path containing a space (e.g. `cd "C:/Users/My Name/Git/engineering-journal"`) would fail to match even though it's a genuine reference. This isn't a regression (the file's existing `_PUSH_RE`/`_MERGE_RE`/`_CREATE_RE` all have the identical `\S+` limitation for their own `cd`-prefix branches, so this is consistent with the surrounding code, not a new gap), and it's not reachable in practice today (this environment's actual engineering-journal path, `C:/Users/brown/Git/engineering-journal`, has no spaces). Worth noting because a more robust pattern already exists in this codebase — `pr-merge-reminder.py`'s `_CD_CHAIN_RE = re.compile(r"""cd\s+("[^"]+"|'[^']+'|[^\s;&|]+)\s*(?:&&|;)""")` — handles quoted paths with embedded spaces correctly.
**Suggestion:** No action needed now given the non-reachability in this environment; if this file's `cd`-prefix handling is ever generalized further, consider porting the quote-aware alternation from `_CD_CHAIN_RE` rather than `\S*`.

**[maintainability]** Language-scope coverage note (Step 2e)
This diff touches only `.py` files, but Step 2e's automated skip-marker/deletion/threshold scanners target JS/TS test frameworks (Jest/Vitest/Mocha) and can't inspect Python. Manually verified during this review: no `@pytest.mark.skip`/`pytest.skip(`/equivalent pattern was introduced (this repo's test files use a hand-rolled `def main() -> int: tests = [...]` runner, not pytest, so the pytest-specific idioms don't even apply here); no test file or `describe`/`it`/`test` block was deleted — only additions; no coverage-threshold or CI bypass config was touched. Confirmed no test-integrity violation.
**Suggestion:** No code change needed. Recorded here for the record per the language-scope preamble's own guidance to note when the integrity check was extended manually.

---

### Style Notes

None. The new `_EJ_REF_RE` / `_check_engineering_journal_ref` / `references_engineering_journal()` trio mirrors the existing `_PUSH_RE` / `_check_push_stmt` / `is_git_push_command` trio's structure, comment style, and docstring format closely — consistent with the file's established conventions.

---

### Findings Disposition

To be completed before merge: the blocking `docs/REFERENCE.md` finding will be fixed in-PR; the quoted-path-with-space note and the language-scope note are non-blocking with no code change required (the latter's disposition is this review comment itself serving as the manual-verification record).

---

Issue #537, opened 2026-07-03
test_pr_merge_reminder.py: no direct test coverage for is_git_push_command's cd-prefix branch


brownm09:
## Problem

`_PUSH_RE` in `claude/scripts/pr-merge-reminder.py` is:

```python
_PUSH_RE = re.compile(r"(?:cd\s+\S+\s+&&\s+)?git\s+push\b")
```

`is_git_push_command()` is built on this regex via the shared `scan_top_level` engine. The existing test suite (`claude/scripts/tests/test_pr_merge_reminder.py`) covers only a bare `git push` (`test_push_simple`) and a `$()` subshell negative case (`test_push_inside_subshell_not_matched`) — there is no test exercising the optional `(?:cd\s+\S+\s+&&\s+)?` prefix branch for `is_git_push_command` specifically, i.e. a `cd <dir> && git push` command.

## Context

Flagged as a non-blocking finding during `/review` of PR #533 (https://github.com/brownm09/dev-env/pull/533), which ported this same `_PUSH_RE`/`is_git_push_command` pair into `stub-push-archive-reminder.py` and added the equivalent cd-prefix test there (`test_push_with_cd_prefix_matched`). The behavior itself was verified correct by hand during that review (`is_git_push_command("cd /Git/engineering-journal && git push")` returns `True`) — this is purely a test-coverage gap, not a bug.

## Goal

Add a test to `test_pr_merge_reminder.py` (e.g. `test_push_with_cd_prefix_matched`) mirroring the one PR #533 added, exercising `is_git_push_command` with a `cd <dir> && git push` command.

Issue #532, opened 2026-07-03
fix(hooks): converge stub-push-archive-reminder.py onto scan_top_level


brownm09:
## Problem

`claude/scripts/stub-push-archive-reminder.py` gates its "is this a git push?" check in `main()` on a crude unanchored substring check: `if "git push" not in command:`. This is the exact false-positive class ADR-050 Amendment 9 (PR #530) just fixed for `post-merge-tile-checkpoint.py`, `post-pr-merge-pull.py`, and `post-pr-merge-reclaim.py`: a literal `"git push"` substring inside a heredoc body, a quoted argument (e.g. a commit message containing the text "git push"), or a `$()` subshell would false-trigger the check without a real invocation.

PR #530's own description explicitly named this file as the one remaining out-of-scope case:

> the one remaining similarly-shaped check, `stub-push-archive-reminder.py`'s unrelated `git push` check, is out of scope for #529 and left untouched

And ADR-050 Amendment 9's "General lesson" flagged it as the next mechanical-grep target:

> A durable mechanical proxy going forward: grep the whole `claude/scripts/` tree for the literal string `"gh pr merge" not in command` (and its `"gh pr create"` / `"git push"` analogs, still live in `stub-push-archive-reminder.py`'s unrelated `git push`-only check, left untouched here as out of scope for this issue) before declaring any future `scan_top_level` convergence sweep complete

## Fix

Converge `stub-push-archive-reminder.py` onto the same `scan_top_level(command, _check_push_stmt)` pattern already established in `usage-snapshot.py` / `pr-merge-reminder.py` / `post-pr-merge-project.py` / `post-merge-tile-checkpoint.py` / `post-pr-merge-pull.py` / `post-pr-merge-reclaim.py`. `pr-merge-reminder.py` already defines the exact predicate needed — `_PUSH_RE = re.compile(r"(?:cd\s+\S+\s+&&\s+)?git\s+push\b")`, `_check_push_stmt`, and a wrapper `is_git_push_command(command)` — so this is a same-shape port, not new regex design.

Concretely: add the identical `_PUSH_RE` / `_check_push_stmt` / `is_git_push_command()` trio to `stub-push-archive-reminder.py`, and replace the inline `if "git push" not in command: sys.exit(0)` in `main()` with `if not is_git_push_command(command): sys.exit(0)`.

## Scope note

[...]

PR #544, opened 2026-07-03
fix(hooks): resolve cross-repo hook-config.json for gh issue/pr create --repo

### LINKED ISSUE #542: post-tool-use.py project-add hook silently no-ops when gh issue/pr create --repo targets a different repo than cwd
## Problem

`post-tool-use.py`'s project-add hook silently does not fire when `gh issue create --repo <other-repo>` (or `gh pr create --repo <other-repo>`) is run from a session whose **cwd belongs to a different project**. Observed twice in one calendar day:

- The session that filed #532 noted directly in its journal stub: *"My session's primary directory is career-playbook, not dev-env... The repo's auto-project-add PostToolUse hook didn't fire after `gh issue create`... ran the documented fallback manually."*
- The session that filed #537 (also from a non-dev-env cwd) hit the same silent non-fire.

Both incidents are reproducible **right now**, from this very issue's own creation: this issue is being filed via `gh issue create --repo brownm09/dev-env ...` from a career-playbook session, and the hook will not fire for the same reason documented below.

## Root cause

`load_config(cwd)` in `claude/scripts/post-tool-use.py` resolves `.claude/hook-config.json` **only** from the invoking session's own cwd (plus two worktree-fallback branches that stay within the *same* repo lineage — `canonical_root_from_worktree` and `canonical_root_via_git`). It has no awareness of the `--repo owner/name` flag on the `gh issue create`/`gh pr create` command itself:

```python
config = load_config(cwd)      # cwd = the SESSION's cwd, never the --repo target
if config is None:
    sys.exit(0)                # silent skip
```

Confirmed directly:
- `C:/Users/brown/Git/dev-env/.claude/hook-config.json` exists (`"repo": "brownm09/dev-env"`, `"project_number": "3"`).
- `C:/Users/brown/Git/career-playbook/.claude/hook-config.json` does **not** exist.
- So `load_config("C:/Users/brown/Git/career-playbook")` returns `None` regardless of what `--repo` the `gh` command names, and the hook exits 0 before ever reading the command's `--repo` flag.

This is architecturally stable, longstanding behavior (`load_config`'s cwd-only resolution model predates today), not a recent regression.

### What this is *not*

An initial pass (background research agent) proposed that both incidents ran the pre-#508 unanchored regex (`re.search(r"\bgh\s+issue\s+create\b", command)`) and that today's anchoring fix (commit `65a5f84`, closing #499/#508) was the culprit. **This is disproven by timestamps**: `65a5f84` merged `2026-07-02 07:12:34 UTC`; #532 was created `2026-07-03 01:07:39 UTC` and #537 at `2026-07-03 01:54:50 UTC` — both ~18 hours *after* the anchoring fix, using the current code. Mechanically it also doesn't hold up: an *unanchored* `re.search` is strictly more permissive than the new anchored `.match()`, so it would have matched a plain top-level `gh issue create --repo ... --title ... --body-file ...` just as easily — the #499/#508 fix addressed false *positives* (over-matching inside heredocs/quotes), not false negatives on a plain invocation. `65a5f84`'s diff touches only the detection predicate, never `load_config`.

The bottom-of-file `except Exception: sys.exit(0)` catch-all is also not implicated — the actual path taken is the deliberate, documented `if config is None: sys.exit(0)` branch, not a swallowed exception.

## Mitigating factor

`reconcile-project-board.py --scan-dir C:/Users/brown/Git` runs nightly (`0 6 * * *`, ADR-068/ADR-070) and catches orphaned issues across every repo with a `.claude/hook-config.json` within ~24h. So neither incident lost data permanently — but the live hook still misses the synchronous "run these commands now, before any file edits" workflow the `## GitHub Project` section documents, forcing the manual-fallback recovery both incidents required.

## Proposed fix

Extend `load_config` to attempt one more resolution when the existing cwd/worktree branches miss **and** the command names an explicit `--repo owner/name`: look for `name` as a sibling checkout under the same parent directory this whole fleet already uses (the same `C:/Users/brown/Git/<repo>` convention `reconcile-project-board.py --scan-dir` and `_repo_scan.find_git_repos` already rely on), and only trust it when the sibling's own `hook-config.json` self-reports `"repo"` matching exactly — never a directory-name guess alone. Mirrors the precedent already established for this exact class of bug in ADR-065 (scope push reminder to push-target repo) and ADR-067 (scope merge-keyed hooks to merge-target repo). Bounded risk: when the sibling resolution can't confirm a match, behavior is byte-for-byte identical to today (silent skip, caught later by the nightly backstop).

## Impact
High — causes manual recovery work on every occurrence (twice today alone), for a workflow pattern (filing a dev-env issue from another project's session) that is common, not edge-case.

## Why
Silently drops the live in-session project-add reminder for any cross-repo `gh issue create`/`gh pr create`, forcing manual recovery each time until the nightly backstop catches up.

brownm09:
## Summary
- `load_config(cwd)` in `post-tool-use.py` only ever resolved `.claude/hook-config.json` from the invoking session's own cwd/worktree lineage, never from the `gh` command itself — so `gh issue create --repo owner/name` filed from a different project's session always silently no-oped. Reproduced twice in one day (#532, #537), and a third time live while filing #542.
- An initial investigation pass mis-attributed this to the pre-#508 unanchored detection regex; disproven by timestamps (both incidents ran ~18h after that fix merged) and confirmed directly by the #532 session's own journal stub: cwd was career-playbook, not dev-env.
- `load_config` now takes an optional `command` string and, after the existing branches miss, resolves an explicit `--repo`/`-R` target as a sibling checkout — trusting it only when the sibling's own config self-reports a matching `repo` field, never a directory-name guess alone.
- Bounded risk: when cross-repo resolution can't confirm a match, behavior is byte-identical to before (silent skip, still caught by the nightly `reconcile-project-board --scan-dir` backstop).

## Test plan
- [x] `py -3 -c "import ast,sys; ..." claude/scripts/*.py claude/scripts/tests/*.py` — syntax check, clean
- [x] `py -3 claude/scripts/tests/test_post_tool_use.py` — **77 passed, 0 skipped, 0 failed** (21 new total: 6 for `extract_repo_flag`, 8 for `_sibling_repo_config`/`load_config` cross-repo resolution, 7 more from the review-fix follow-up commit covering non-dict JSON, malformed paths, the pre-fix crash repro, trailing-separator `own_root`, and URL-form/host-prefixed `--repo`)
- [x] Live-reproduced the bug while filing #542 itself (no hook reminder fired), then ran the documented manual fallback to board it
- [x] Docs updated: `docs/REFERENCE.md` hooks table row, new `docs/adr/077-cross-repo-config-resolution-for-issue-pr-create.md`, `docs/adr/INDEX.md`

Tests: 77 passed, 0 skipped, 0 failed

Closes #542

`/review` ran a parallel two-subagent pass (diff was 419 lines). All 4 findings fixed in commit `972a8c4`:
- **[correctness, blocking]** `_read_config` unhandled `OSError`/`ValueError` on a malformed derived path (reproduced directly) — fixed, now catches both and rejects non-dict JSON.
- **[correctness, non-blocking]** URL-form/host-prefixed `--repo` not normalized before comparison — fixed.
- **[reliability, non-blocking]** trailing-separator on `own_root` searches one level too deep — fixed.
- **[maintainability, non-blocking]** cross-repo branch's trigger condition undocumented — fixed with a clarifying comment.

All verified by 7 new regression tests (77 total, 0 failed). Full review: https://github.com/brownm09/dev-env/pull/544#issuecomment-4872304774

**Not merged pending Mike's explicit confirmation** — this modifies core harness automation (fires on every Bash tool call, every session); held for a sanity check despite the repo's normal in-session merge convention, given the blast radius.


brownm09:
## Review: fix(hooks): resolve cross-repo hook-config.json for gh issue/pr create --repo

### Summary
Fixes a confirmed, twice-reproduced silent failure in `post-tool-use.py`'s cross-repo issue/PR project-add path (`load_config` never inspected the `gh` command's `--repo` flag). The parallel two-subagent review found and reproduced a real crash bug plus two lower-severity gaps, all fixed in a follow-up commit with 7 additional regression tests. Ready to merge pending final confirmation from the PR author.

---

### Blocking Findings

**[correctness]** `claude/scripts/post-tool-use.py` (`_read_config`)
`_read_config` only caught `FileNotFoundError`/`json.JSONDecodeError`. A misparsed `--repo` value (e.g. literal `-R foo:bar` text inside an earlier quoted `--title`) derives a path with a Windows-reserved character, which raised an unhandled `OSError` instead of returning `None` — reproduced directly against the pre-fix code.
This runs inside a global PostToolUse hook firing on every Bash tool call in every session; the outer `try/except Exception: sys.exit(0)` swallows the exception, so the visible symptom is "the hook mysteriously stopped firing" for an unrelated, legitimate same-repo project-add — reintroducing the exact silent-failure class this PR set out to fix, via a different path.
**Fix:** `_read_config` now catches `(OSError, ValueError)` (subsumes the original two exception types) and rejects non-dict JSON, retroactively hardening the pre-existing cwd/worktree read paths too. Fixed in `972a8c4`.

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/post-tool-use.py` (`extract_repo_flag`)
A URL-form (`--repo https://github.com/owner/name`) or bare host-prefixed (`--repo github.com/owner/name`) value — both valid `gh` syntax — wasn't normalized before the sibling config's canonical `owner/name` comparison, so the cross-repo resolution silently failed to fire for these forms even when the sibling correctly existed. Bounded (no wrong add) but a real completeness gap for a common way of writing `--repo`.
**Fix:** Normalized to `owner/name` before comparison. Fixed in `972a8c4`.

**[reliability]** `claude/scripts/post-tool-use.py` (`_sibling_repo_config`)
`own_root` can be the unnormalized raw `cwd` (the `root or cwd` fallback when `canonical_root_via_git` fails); a trailing separator made `os.path.dirname` return `own_root` itself instead of its parent, silently searching one level too deep. Fails safe, but the feature silently doesn't work for that session.
**Fix:** `own_root` normalized via `os.path.normpath` at entry. Fixed in `972a8c4`.

**[maintainability]** `claude/scripts/post-tool-use.py` (`load_config`)
The cross-repo branch is only reached when the invoking session's own cwd (and worktree fallbacks) resolve no config — non-obvious from reading the branch alone.
**Suggestion:** Addressed with a clarifying comment at the `if command:` branch. Fixed in `972a8c4`.

---

### Style Notes
None beyond the above — reviewed with style checks on; no separate style-only findings surfaced.

---

### Findings Disposition
All four findings fixed in commit `972a8c4` (pushed to this branch), verified by 7 new regression tests (77 total, 0 failed) including a direct repro of the pre-fix crash.

---

PR #560, opened 2026-07-04
[fix] Recognize gh pr merge -d/--delete-branch as canonical-checkout mutation

### LINKED ISSUE #558: canonical-mutate-guard doesn't recognize gh pr merge --delete-branch as local-mutating
## Context

While investigating a canonical-checkout race incident (lifting-logbook, 2026-07-03: two Claude Code sessions concurrently in `C:/Users/brown/Git/lifting-logbook`, one session's `git push`/`gh pr merge` silently raced the other's), I checked whether `pre-tool-use-canonical-mutate-guard.py` (ADR-071) should extend to `git push`, `gh pr create`, and `gh pr merge`.

Conclusion for plain push/create/merge: **no** — ADR-071 blocks because a mutating command silently corrupts *local* state a concurrent session depends on. `git push` (rejects non-fast-forward / no-ops identical-SHA), `gh pr create` (GitHub rejects a duplicate head+base PR), and a bare `gh pr merge` (GitHub rejects double-merge) all already fail loudly and unambiguously at the git/GitHub layer — exactly what happened in the incident. Blocking these would violate ADR-071's own zero-friction-for-safe-operations design for operations that can't actually corrupt anything.

## The actual gap

`gh pr merge --help` output confirms:

```
-d, --delete-branch          Delete the local and remote branch after merge
```

`gh pr merge -d`/`--delete-branch`, run from the branch it's merging, must check out the base branch and delete the local branch to fulfil `--delete-branch` (a checked-out branch can't be deleted) — this is the **exact same silent local-HEAD-thrash harm model** ADR-071 already blocks for `git checkout`/`branch -d`, just reached through a `gh` invocation rather than a literal `git` verb. `pre-tool-use-canonical-mutate-guard.py`'s `is_mutating_segment()` only classifies `git`/`git.exe`-prefixed invocations, so this command surface is currently invisible to the guard — a `gh pr merge -d` issued directly in a canonical (non-worktree) checkout can silently thrash HEAD out from under a concurrent session exactly as dev-env#453 describes, with no block and no override needed.

## Proposed fix

Add a sibling classifier (e.g. `is_mutating_gh_segment()`) alongside the existing `is_mutating_segment()` in `pre-tool-use-canonical-mutate-guard.py`: true iff a segment is a `gh pr merge` invocation carrying `-d`/`--delete-branch`. `classify()`'s per-segment loop checks both classifiers with the same severity (block, exit 2) — same file, same hook, same existing cd-scope/redirect-scope/override machinery, no new hook needed. A bare `gh pr merge` (no delete-branch flag) merges only remotely via the API and touches no local state, so it stays unblocked.

This is an amendment to ADR-071's existing decision (same harm model, previously-unrecognized command surface), not a new ADR — matches this repo's established amend-in-place convention (ADR-050, ADR-058, etc.). ADR-071's own docstring warns the mutating-verb list is independently re-spelled in four places (module docstring, `claude/CLAUDE.md`'s "Never mutate git state directly" bullet, ADR-071's own "Mutating verbs blocked" line, `docs/REFERENCE.md`'s pointer) — all four need the `gh pr merge -d` addition.

## Impact

Medium — same collision class as dev-env#453 (silent HEAD-thrash between two sessions sharing a canonical checkout), but narrower: only triggers when a session runs `gh pr merge` with `-d`/`--delete-branch` directly in a canonical checkout, which is common (it's in this repo's own documented `gh pr merge --squash --delete-branch` convention).

brownm09:
## Summary

`pre-tool-use-canonical-mutate-guard.py` (ADR-071) hard-blocks git-mutating Bash commands issued directly in a canonical (non-worktree) checkout — because two Claude Code sessions sharing one checkout can collide when one session's mutating command silently thrashes HEAD/working-tree state out from under the other (dev-env#453).

**The gap:** `gh pr merge --help` documents `-d, --delete-branch: Delete the local and remote branch after merge`. Run from the branch it's merging, this must locally check out the base branch and delete the local branch — the exact same silent local-HEAD-thrash harm model ADR-071 already blocks for `git checkout`/`git branch -d`, just reached through a `gh` invocation instead of a literal `git` verb. `is_mutating_segment()` only classifies `git`/`git.exe`-prefixed invocations, so this command surface was invisible to the guard. This repo's own documented convention (`gh pr merge --squash --delete-branch`, used throughout `claude/CLAUDE.md`'s Git Workflow section) is the common path that hits this gap.

A **bare** `gh pr merge` (no `-d`/`--delete-branch`) merges only remotely via the GitHub API and touches no local state — stays unblocked, matching dev-env#558's original conclusion that plain `git push`/`gh pr create`/bare `gh pr merge` don't need blocking (GitHub/git already reject double-merge/duplicate-PR loudly at the remote layer).

## Changes

- **`claude/scripts/pre-tool-use-canonical-mutate-guard.py`** — new, separate classifier `is_mutating_gh_segment(segment)` (NOT folded into `is_mutating_segment()`, which is specifically documented as a *git*-invocation classifier). Matches a `gh`/`gh.exe` + `pr` + `merge` invocation at the start of the (env-stripped) segment's first physical line, returning True when the remaining arguments contain `-d`/`--delete-branch` as a standalone token, **or** an explicit `--delete-branch=<value>` whose value isn't falsy (`false`/`0`/`no`) — see Review findings disposition below. Wired into `classify()`'s existing per-segment loop via `is_mutating_segment(seg) or is_mutating_gh_segment(seg)` — same severity (block, exit 2), same existing `cd`-scope-out / `-C`/`--git-dir`-redirect-scope / `ALLOW_CANONICAL_MUTATE=1`-override machinery, all of which needed zero changes. Deliberately does not special-case a `--repo owner/repo` flag, matching the hook's own "block when in doubt" judgment call for bare `git checkout <path>`. Tweaked `main()`'s block-message `reason` text from "a git-mutating command" to "a command that mutates local git state" since it now also matches `gh` invocations.
- **`claude/scripts/tests/test_canonical_mutate_guard.py`** — 10 new tests across the file's existing two-layer convention, plus 2 more cases added during review (see disposition below): pure-function coverage of `is_mutating_gh_segment()`/`classify()` (delete-branch flag matrix including the `=true`/`=false` explicit-value form, heredoc-mention non-trigger, override bypass) and end-to-end `main()`-via-subprocess coverage (`gh pr merge --delete-branch`/`-d` blocked from a canonical root; bare `gh pr merge`/`--squash` allowed; same command from a worktree-pattern cwd allowed out of scope; override token bypasses).
- **`docs/adr/071-canonical-checkout-mutate-guard-hook.md`** — amended in place (Amendment 1, matching this repo's amend-in-place convention for same-decision extensions — same harm model, same severity, same file/hook, only a previously-unrecognized command surface). Date line updated to `2026-07-01 (amended 2026-07-04)`.
- **Four sync locations** the hook's own docstring warns must stay in sync, all updated in this PR:
  1. Module docstring in `pre-tool-use-canonical-mutate-guard.py` ("Mutating verbs blocked:" section)
  2. `claude/CLAUDE.md`'s "Never mutate git state directly..." bullet
  3. ADR-071's own Amendment 1 section (the original "Mutating verbs blocked:" line is left unedited for history, per this repo's convention — see ADR-058, ADR-050)
  4. `docs/REFERENCE.md`'s Hooks-table ADR-071 pointer
- **`docs/adr/INDEX.md`** — ADR-071 row date updated to show the amendment; added `gh-cli` tag.
- **`CLAUDE.md`** (dev-env project root) — Testing-section item 33's description extended to document the new `is_mutating_gh_segment()` coverage (no new numbered item — this extends an existing test file, matching the repo's convention of updating in place for amendments).

## Testing

```
# ALL SCRIPTS SYNTAX OK

# Tests: 48 passed, 0 skipped, 0 failed (38 pre-existing + 10 new, incl. the 2 review-driven cases)

# Tests: 13 passed, 0 skipped, 0 failed (AST-based repo-wide gate)

# Tests: 44 passed, 0 skipped, 0 failed (regression check)
```

Pre-PR suppression grep and test-integrity grep both returned no matches — no suppressions, no test-integrity violations.

No `package.json` changed — lockfile-sync check not applicable.

Closes #558

- **[correctness]** `is_mutating_gh_segment()`'s exact-token match missed Cobra's `--delete-branch=<value>` explicit-value form (a false negative for a genuinely mutating invocation) — **fixed in `74938c9`**: now also recognizes `--delete-branch=<truthy-value>` while correctly treating an explicit `=false`/`=0`/`=no` as a real opt-out (no local mutation), with 2 new test cases added.


brownm09:
## Review: [fix] Recognize gh pr merge -d/--delete-branch as canonical-checkout mutation

### Summary
Ready to merge with one small fix applied in-PR. The change is precise and well-isolated: a new `is_mutating_gh_segment()` classifier sits alongside (not folded into) the existing git-verb classifier, reuses all existing cd-scope/redirect-scope/override machinery in `classify()` with zero changes there beyond the `or` condition, and all four ADR-071-documented sync locations were updated consistently. 48/48 tests pass, no suppressions, no test-integrity violations. One non-blocking correctness edge case below.

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:~120` (`is_mutating_gh_segment`)
The delete-branch check is an exact-token match: `any(t in _GH_DELETE_BRANCH_FLAGS for t in tokens[2:])` against `{"-d", "--delete-branch"}`. Cobra-style boolean flags also accept an explicit-value form (`--delete-branch=true`), which would fail this exact match and slip through unclassified — the one invocation shape that both (a) genuinely deletes the branch locally and (b) currently produces a false negative in this guard. Narrow in practice (the bare `--delete-branch`/`-d` form is what people and agents actually type — including this repo's own documented `gh pr merge --squash --delete-branch` convention), so not blocking, but it's a safety-hook gap in the same family this PR exists to close.
**Suggestion:** Also match a `--delete-branch=` prefix, e.g. `any(t in _GH_DELETE_BRANCH_FLAGS or t.startswith("--delete-branch=") for t in tokens[2:])`, and add one test case for `gh pr merge --delete-branch=true`.

---

### Findings Disposition
Fixing the one non-blocking finding directly in this PR (trivial, one-line change + one test) rather than filing a follow-up — see disposition note added to the PR body after the fix lands.

---

Issue #423, opened 2026-06-29
journal-compose: add pre-compose manifest field validator (Step-0 gate)


brownm09:
## Problem

Engineering-journal stubs written against an **older manifest schema** fail validation only *mid-compose*, and the missing fields are repaired by hand each time. The manifest schema enforces five required fields (`stub`, `topic`, `tokens`, `prs_opened`, `prs_closed`); when a stub predates a field's addition, the gap is silent until a `journal-compose` subagent trips over it.

Observed occurrences (last two weeks):
- **2026-06-13** — career-playbook ×5 entries missing `topic`/`tokens` (`sessions/meta/2026-06-14-automated-journal-compose-2026-06-13.md`)
- **2026-06-17** — lifting-logbook manifest line 3 missing `prs_closed` (`sessions/meta/2026-06-18-conflict-recovery-compose-2026-06-17.md`)

This is a **silent-skip class**: validation that should fire up front instead surfaces late, mid-run, and is hand-patched — risking a malformed journal if missed.

## Proposed remediation

Add a pure validator helper under `claude/scripts/` with an offline `claude/scripts/tests/test_*.py` test (matching the repo's pure-helper convention), wired into the `journal-compose` skill as a **Step-0 gate** that lists every manifest entry missing a required field **before** any subagent runs. Converts a late silent violation into an immediate visible one.

## Context

Surfaced by the 2026-06-29 silent-skip audit (hooks & instructions, last ~2 weeks). This was item #12 — the one open in-repo gap; the other 11 incidents were already remediated (ADR-049/050/051/052/053/055/058/060, etc.).

PR #561, opened 2026-07-04
feat(hooks): validate journal shard schema at write time, not just at compose

### LINKED ISSUE #556: Journal shards written after morning compose bypass the #423 field-schema gate for a full day
## Problem

`validate-manifest.py` (issue #423 lineage) gates engineering-journal manifest shards only at compose time — i.e. the **next day's** `/journal-compose` run (Step 0.7). Sessions keep writing bad shards all day, and someone repairs them by hand later.

## Evidence

From the 2026-07-02 draft (hand-repaired 2026-07-03, engineering-journal `draft/2026-07-02` commit 153966e):

- 15 career-playbook shards missing `topic`/`tokens` (PR-merge bookkeeping sessions re-creating manifests after the morning compose consumed the originals)
- 3 meta shards (2026-07-02_140000/210254/214323) missing `stub`/`topic`/`tokens`, carrying a `summary` field instead of `topic`
- 1 meta shard (2026-07-02_130000) + `sessions/meta/open-prs/147.json` written with a UTF-8 BOM — Node `JSON.parse` rejects BOM, so compose Step 8a dashboard aggregation and the Python shard readers (`_journal_shards.py`) silently skip them via try/catch

## Proposal

Add a PostToolUse hook (`journal-shard-write-advisory.py`) that validates any Write/Edit/Bash call touching engineering-journal `sessions/**/*.manifest.jsonl` and `sessions/**/open-prs/*.json` against the required-field schema, flagging missing fields and BOMs immediately in the writing session — an exit-2 advisory (memory-write-advisory.py's pattern), not a block. Reuse `validate-manifest.py`'s validation logic via a new shared module (`_journal_schema.py`) rather than duplicating it. Also add a one-line reminder to the global CLAUDE.md stub workflow that a manifest re-created after a compose must carry the full field set, not just the updated field.

Full design in the approved plan for this session.

brownm09:
## Summary

`validate-manifest.py` (#423) gates engineering-journal manifest shards only at compose
time — the *next day's* `/journal-compose` run. A shard written (or re-created) after the
morning compose bypasses that gate for a full day. This is exactly what happened on
2026-07-02 (hand-repaired 2026-07-03, engineering-journal `draft/2026-07-02` commit
`153966e`): 15 career-playbook shards missing `topic`/`tokens`, 3 meta shards missing
`stub`/`topic`/`tokens` (carrying `summary` instead of `topic`), and 1 meta shard +
`sessions/meta/open-prs/147.json` written with a UTF-8 BOM that every JSON reader
silently skips.

- Adds `claude/scripts/_journal_schema.py` — extracts `REQUIRED_FIELDS`,
  `missing_required_fields()`, `find_entries_missing_fields()`, `parse_manifest_text()`
  out of `validate-manifest.py` (re-imported there, so its existing tests are
  unaffected), adds `OPEN_PR_REQUIRED_FIELDS` / `missing_open_pr_fields()` for the
  open-PR shard schema, and a new `decode_shard_bytes()` that names a BOM instead of
  letting it surface as an opaque line-1 JSON parse failure.
- Adds `claude/scripts/journal-shard-write-advisory.py` — a PostToolUse hook wired on
  `Write`/`Edit`/`Bash` that validates any touched manifest or open-PR shard's **on-disk
  bytes** against the shared schema, immediately, in the writing session. Non-blocking
  exit-2 advisory (mirrors `memory-write-advisory.py`'s ADR-048 convention) — the write
  already happened, this surfaces the problem rather than blocking it.
- Wires the new hook into `claude/settings.json` (Bash, Write, and a new Edit matcher
  block).
- Adds a one-line note to the global CLAUDE.md stub workflow: a manifest re-created
  after a compose must carry the full five-field set, not just the updated field.
- Updates `README.md`, `docs/REFERENCE.md`, and `journal-compose/SKILL.md` per the
  doc-reconciliation checklist.
- **ADR-081** records the design (on-disk validation vs. payload content, why the Bash
  token harvest is deliberately not `scan_top_level`-anchored, the shared-module
  extraction, and full-strictness open-PR schema enforcement).

Closes #556

## Testing

Ran the full required set from dev-env's `## Testing` section for every touched/new
file:

- Item 1 (syntax): `py -3 -c "import ast,sys; ..." claude/scripts/*.py` → OK
- Item 2 (pyw stdio): `py -3 claude/scripts/tests/test_pyw_stdio.py` → 7 passed
- Item 25 (validate-manifest, amended): `py -3 claude/scripts/tests/test_validate_manifest.py` → 28 passed
- Item 39 (no-crude-substring AST gate): `py -3 claude/scripts/tests/test_no_crude_command_substring_checks.py` → 13 passed (0 unexpected — confirms the new hook's `re.findall` token harvest doesn't trip this gate, by design; see ADR-081)
- Item 40 (new — journal-shard-write-advisory): `py -3 claude/scripts/tests/test_journal_shard_write_advisory.py` → 61 passed
- Item 41 (new — `_journal_schema`): `py -3 claude/scripts/tests/test_journal_schema.py` → 18 passed

**Tests: 127 passed, 0 skipped, 0 failed** (28+13+61+18+7+syntax, across the six required commands, after the `/review` fixes below)

Also ran manual functional smoke tests reproducing the 2026-07-02 incident shapes
directly (a Write payload against a manifest shard missing `stub`/`tokens`, and a Bash
`git -C ... add open-prs/147.json` payload against a BOM'd open-PR shard) — both
produced the correct exit-2 advisory with the expected problem text. Also verified
end-to-end that a 100,000-character pathological Bash payload (the regex-backtracking
fix below) now completes in 0.24s through the real hook.

Pre-PR checks: suppression grep clean, test-integrity grep clean, no deleted test
files, no `package.json`/lockfile changes (N/A), `baseline_test_failure_tracking` not
enabled for dev-env (N/A — Jest-only feature).

`/review` (2 parallel Opus subagents — correctness/security, and reliability/performance/
maintainability) found 1 blocking + 3 non-blocking findings ([review comment](https://github.com/brownm09/dev-env/pull/561#issuecomment-4880960659)). All four fixed in commit `d135fc1`:

- **Fixed in `d135fc1`** — [blocking, performance] `_MANIFEST_TOKEN_RE`/`_OPEN_PR_TOKEN_RE` are
  O(n²) via `re.findall`'s per-start-position retries when their required literal suffix never
  appears in a long run of matching characters (verified directly: ~10s for one regex against a
  40,000-char blob; an atomic-group rewrite does **not** help, since the blowup is from retrying
  the whole match at every start position, not backtracking within one match). Fixed with a
  `MAX_COMMAND_CHARS=2000` guard before either regex runs. Verified end-to-end: a 100,000-char
  pathological payload through the real hook dropped from a would-be 60–100s+ to 0.24s.
- **Fixed in `d135fc1`** — [reliability] sibling-module imports (`_journal_schema`,
  `_journal_shards`) were outside the safe-exit guard — wrapped in a module-level
  `try/except: sys.exit(0)`.
- **Fixed in `d135fc1`** — [correctness] a non-dict `tool_input` payload would raise
  `AttributeError` before hitting the outer catch-all — added an explicit
  `isinstance(tool_input, dict)` guard in `main()`.
- **Fixed in `d135fc1`** — [reliability] the open-PR non-numeric-filename check was unreachable
  when the shard was also empty/malformed (an early return skipped it) — hoisted above those
  checks so both problems are reported together.

Two additional points the correctness/security subagent raised were investigated and dropped as
not real findings (not included in the posted review): a claimed `splitlines()`/U+2028 behavior
change from the bytes-vs-text-mode refactor turned out to produce byte-identical output old vs.
new (verified directly), and a claimed `MultiEdit` matcher gap named a tool that doesn't appear
anywhere in this repo's hook wiring.

4 new/updated tests added alongside the fixes (`test_extract_skips_oversized_command`,
`test_extract_pathological_input_completes_quickly`,
`test_validate_open_pr_non_numeric_stem_reported_even_when_empty`,
`test_validate_open_pr_non_numeric_stem_reported_even_when_not_json`), plus 2 pre-existing tests
updated to isolate their intended scenario from the now-earlier-firing non-numeric-filename check.


brownm09:
## Review: feat(hooks): validate journal shard schema at write time, not just at compose

### Summary

The design is sound and well-tested (141 passing tests across the six required Testing items), and the extraction into `_journal_schema.py` cleanly avoids duplicating the schema between the compose-time gate and the new write-time hook. One blocking finding: the Bash command-token harvest regexes are quadratic on pathological input and run on **every** Bash tool call session-wide, directly contradicting the ADR's "negligible cost" claim — verified empirically (a 40,000-character blob with no path suffix costs ~10s per regex). At 1,495 changed lines this PR is appropriately scoped as one cohesive unit (splitting the schema extraction from its only consumer would leave an awkward intermediate state), so no split is recommended — needs revision on the one blocking item before merge.

---

### Blocking Findings

**[performance]** `claude/scripts/journal-shard-write-advisory.py:63-64`
`_MANIFEST_TOKEN_RE` and `_OPEN_PR_TOKEN_RE` each lead with an unbounded greedy character class (`[\w./\\:~-]+` / `[\w./\\:~-]*`) followed by a required literal suffix, with no pre-filter on `command` length before `extract_candidate_tokens` runs both regexes over the full string on **every** `Bash` tool call in **every** session — not just journal-related ones. When the class-matching run never reaches the suffix, the engine backtracks one character at a time and retries the suffix match at every position, and `re.findall`/`search` additionally retries the whole pattern from every start position when no match is found — producing O(n²) work. I verified this directly: a 40,000-character run of word characters (a base64 blob, a long hash, an inlined `data:` URI, minified JSON pasted inline — all ordinary in real Bash commands) costs ~10s for one regex alone; both regexes combined roughly double that. This stalls the tool call for the user on an entirely ordinary command with no journal content at all.
**Fix:** Bound the input before scanning — add a length guard (e.g. skip the harvest above ~4–8 KB; legitimate journal paths are always short) mirroring the file's existing `MAX_CANDIDATES`/`MAX_SHARD_BYTES` defensive-cap convention. This is the minimal, structurally-consistent fix; re-verify with the same 40K-character fixture after the change.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/journal-shard-write-advisory.py:47-53`
The `from _journal_schema import ...` / `from _journal_shards import ...` statements execute at module import time, outside the `try/except Exception` safe-exit guard in `if __name__ == "__main__":`. If either sibling module ever fails to import (a future syntax error, a renamed export), the hook crashes before `main()` is reached and the guard never fires — violating the documented hook-authoring invariant that the safe-exit guard covers *every* code path (`docs/REFERENCE.md` → Authoring rules #2). Both sibling modules are currently stable, so practical risk is low, but the gap is real and cheap to close.
**Suggestion:** Wrap the sibling imports in a `try/except Exception: sys.exit(0)` at module level (or move them inside `main()`), so the invariant holds structurally rather than by luck.

**[correctness]** `claude/scripts/journal-shard-write-advisory.py:426-443` (`candidate_paths`)
`main()` builds `tool_input` via `data.get("tool_input", {}) or {}` — if the payload's `tool_input` were a non-dict truthy value (e.g. a JSON list), `or {}` would not replace it, and the subsequent `tool_input.get(...)` call raises `AttributeError`. PostToolUse payloads from Claude Code always send `tool_input` as a dict, so this is theoretical, and the outer safe-exit guard already keeps the hook's exit code at 0 either way — but the function's own type hint (`tool_input: dict`) implies a shape it doesn't actually enforce.
**Suggestion:** Guard with `isinstance(tool_input, dict)` in `main()` so the fail-open is explicit by design rather than incidental via the outer catch-all.

**[reliability]** `claude/scripts/journal-shard-write-advisory.py:189-207` (`validate_shard_bytes`)
For an open-PR shard, the empty-shard and non-JSON-object checks both `return problems` early, before the non-numeric-filename / stem-mismatch checks run. So a shard that is *both* empty/malformed *and* has a non-numeric filename (e.g. a stray `open-prs/index.json` that's also zero-length) gets flagged only for the parse/empty problem — the arguably more important "invisible to every reader" diagnosis for the filename itself never surfaces. The file is still flagged either way, so this is a minor completeness gap, not a missed detection.
**Suggestion:** Hoist the `pr_from_name is None` check above the empty/parse-failure early returns for the `open-pr` kind, since it depends only on the filename, not the parsed content.

**[maintainability]** `claude/scripts/_journal_schema.py:62-72` (`find_entries_missing_fields`)
This function hardcodes the manifest schema (`missing_required_fields(entry)` with no `fields` override) despite living in a module that now serves two schemas. Correct for its one caller (`validate-manifest.py`), but the name no longer signals "manifest-only" the way it did before this module served both shard kinds.
**Suggestion:** Add a one-line docstring note that this helper is manifest-schema-specific, so a future reader doesn't reach for it to validate a list of open-PR entries.

---

### Findings Disposition

All four findings above (one blocking, three non-blocking) will be fixed in this PR before merge; disposition recorded in the PR body per ADR-028/ADR-039 once committed.

---

PR #562, opened 2026-07-04
[fix] Gate merge-hook live-confirmation fallback on gh pr merge --help/-h

### LINKED ISSUE #557: gh pr merge --help misattributed as a completed merge via confirm-via-gh fallback
## What happened

While investigating dev-env#453-style canonical-checkout races (lifting-logbook session, 2026-07-03), I ran `gh pr merge --help` purely to check flag semantics. Three PostToolUse hooks — `post-pr-merge-project.py`, `post-merge-tile-checkpoint.py`, `usage-snapshot.py` — fired as if PR #664 (lifting-logbook) had just merged: the project-board hook actually moved issue #660 to Done.

No lasting harm: issue #660 was already `CLOSED`/`COMPLETED` (merged earlier by a concurrent session), so the move was a no-op, and `main`'s reflog showed no unexpected mutation. But the mechanism is a real, reproducible defect.

## Root cause

NOT the `-1` vs `0` exitCode default inconsistency I first suspected — `test_should_confirm_default_exit_code_attempts` in `test_hookio.py` proves the `-1` default is deliberate (favor a live confirmation attempt over silently missing a real merge, per the #275/#489 worktree-cleanup-failure case).

The actual gap: `gh pr merge --help` textually satisfies every `is_pr_merge_command`/`_check_merge_stmt` predicate in the hook family (it *is* a `gh pr merge` invocation, syntactically). Since `--help` prints no success marker, `should_confirm_via_gh` returns True (no marker + effectively-missing/non-zero exit), triggering `confirm_merge_via_gh(None, "", cwd)` — which, with no explicit PR number, resolves via `gh pr view` scoped to **cwd's current branch**. If that branch has ANY merged PR on record (regardless of whether it has anything to do with the command just run), the hook attributes that merge to the `--help` invocation.

This isn't unique to `--help` — any non-mutating `gh pr merge` invocation that prints no marker (e.g. a failed auth check, a dry inspection) would trigger the same misattribution. `--help`/`-h` is the clearest, most reproducible case: it can *categorically never* attempt a real merge, so it should never satisfy this check at all.

## Affected files (identical `scan_top_level → should_confirm_via_gh → confirm_merge_via_gh` shape)

- `post-pr-merge-project.py` (confirmed misfired — moved a project item)
- `post-merge-tile-checkpoint.py` (confirmed misfired)
- `usage-snapshot.py` (confirmed misfired)
- `post-pr-merge-pull.py` (same shape; would silently fetch/pull local main — always exits 0, so no visible reminder either way)
- `post-pr-merge-reclaim.py` (same shape; would spawn a detached disk-reclaim sweep across `C:/Users/brown/Git` — always exits 0)
- `pr-merge-reminder.py` (same shape, though its `0` exitCode default makes it less exposed)

Plus two PreToolUse siblings with a related (lower-severity) gap — they'd evaluate/potentially block a `--help` call based on an unrelated PR's findings or numbering state:
- `pre-merge-findings-gate.py`
- `pre-merge-numbering-check.py`

## Proposed fix

Add one shared, well-tested predicate to `_hookio.py`, e.g. `is_merge_help_only(command)` — true iff every top-level `gh pr merge` segment in the command carries `--help`/`-h` (segment-scoped via `split_top_level`, so a heredoc/quoted mention can't mask a real merge elsewhere, mirroring the #442/#511/dev-env#529 heredoc-scoping lessons already applied to merge detection in this hook family). Each of the 8 files above gets one new guard line (`if is_merge_help_only(command): sys.exit(0)`, placed alongside their existing `is_pr_merge_command`/`_check_merge_stmt` gate) plus an import — no changes to the existing, already-tested exit-code/marker logic.

## Impact

Medium — this incident was harmless only because the misattributed issue happened to already be Done. A different stale-PR scenario (a canonical checkout whose current branch carries an unrelated already-merged PR closing a *different* issue) would silently move the wrong issue to Done, and could trigger an unscheduled disk-reclaim sweep or a silent local-main fetch, all from an innocuous `--help`/documentation-check command.

brownm09:
## Summary

Fixes a live incident (#557) where `gh pr merge --help` — run purely to check flag semantics — was misattributed as a completed merge and moved an unrelated issue's project-board item to Done.

## Root cause

`gh pr merge --help` textually satisfies every `is_pr_merge_command` / `_check_merge_stmt` predicate in the merge-hook family (it *is* a `gh pr merge` invocation, syntactically). Since `--help` prints no success marker, `should_confirm_via_gh` returns `True` (no marker + non-zero/`-1` exit), triggering `confirm_merge_via_gh(None, "", cwd)` with no explicit PR number — which resolves via `gh pr view` scoped to **cwd's currently checked-out branch**. If that branch has any merged PR on record, the hook attributes that merge to the harmless `--help` invocation.

This is not unique to `--help` — any non-mutating `gh pr merge` invocation that prints no marker would trigger the same misattribution — but `--help`/`-h` is the clearest, most reproducible case: it can *categorically never* attempt a real merge.

**Not** the `should_confirm_via_gh` `-1` exit-code default (confirmed deliberate — see `test_should_confirm_default_exit_code_attempts` in `test_hookio.py`, unchanged by this PR) and not a `scan_top_level` false-positive (the command genuinely does invoke `gh pr merge` — this is a different class of gap than ADR-050's prior amendments).

## Fix

Added a shared, pure, well-tested predicate to `claude/scripts/_hookio.py`:

```python
def is_merge_help_only(command: str) -> bool:
    """True iff every top-level `gh pr merge` segment in *command* is a --help/-h invocation."""
```

- Finds every top-level segment (`split_top_level`) whose own first physical line matches `gh(?:\.exe)?\s+pr\s+merge\b` (case-insensitive, anchored at the lstripped segment start).
- Returns `False` if there are no such segments (callers already gate on their own merge-detection check first).
- Returns `True` only if **all** matched segments carry a standalone `--help`/`-h` flag token on that segment's own first physical line — so a chained `gh pr merge --help && gh pr merge 380 --squash` correctly returns `False` (a real merge attempt elsewhere is never suppressed).
- Segment-scoped via `split_top_level`, so a heredoc body or quoted argument merely *mentioning* `gh pr merge --help` as prose can't mask (or fake) help-only status.

Wired into all 8 affected files, each as an early exit treating a help-only command exactly like "not a merge command at all":

| File | Guard placement |
|---|---|
| `post-pr-merge-project.py` | Right after `if not merge_succeeded(output):`, before computing `exit_code` |
| `post-merge-tile-checkpoint.py` | Right after its existing `if not scan_top_level(command, _check_merge_stmt): sys.exit(0)` line |
| `usage-snapshot.py` | Same shape (predicate `merge_confirmed`) |
| `post-pr-merge-pull.py` | Same shape (predicate `is_successful_merge`) |
| `post-pr-merge-reclaim.py` | Same shape (predicate `is_successful_merge`) |
| `pr-merge-reminder.py` | Added `and not is_merge_help_only(command)` into its inline `live_confirmed` condition |
| `pre-merge-findings-gate.py` | Right after `if not is_pr_merge_command(command): sys.exit(0)` (PreToolUse) |
| `pre-merge-numbering-check.py` | Same shape (PreToolUse) |

No change to `should_confirm_via_gh`'s deliberate `-1` exit-code default, and no change to any file's own `_check_merge_stmt`/`is_pr_merge_command` regex.

## Testing

Run from the repo root — see [dev-env `CLAUDE.md` → Testing](https://github.com/brownm09/dev-env/blob/main/CLAUDE.md#testing) item 1 (syntax check) and item 13 (`_hookio` shared-read test), plus each touched hook's own numbered test item.

```
```
→ all `claude/scripts/*.py` parse clean.

**Tests: 234 passed, 0 skipped, 0 failed**

| Test file | Result | New cases |
|---|---|---|
| `test_hookio.py` | 57 passed | 13 (the `is_merge_help_only` predicate itself — bare `--help`/`-h`, real merge, bare merge no `--help`, no merge invocation, chained help-then-real-merge stays False, chained two-help stays True, heredoc/quoted-argument mentions ignored in both directions, flag-substring false-match guard, `cd`-prefix, case-insensitivity) |
| `test_post_pr_merge_project.py` | 18 passed | 2 (composition: `merge_succeeded` False + `is_merge_help_only` True for `--help`; unresolved real merge unaffected) |
| `test_post_merge_tile_checkpoint.py` | 9 passed | 2 (same composition shape, `is_successful_merge`) |
| `test_usage_snapshot.py` | 13 passed | 2 (same shape, `merge_confirmed`) |
| `test_post_pr_merge_pull.py` | 14 passed | 2 (same shape, `is_successful_merge`) |
| `test_post_pr_merge_reclaim.py` | 9 passed | 2 (same shape, `is_successful_merge`) |
| `test_pr_merge_reminder.py` | 48 passed | 2 (re-derives `main()`'s inline `live_confirmed` boolean expression directly, since it isn't its own function) |
| `test_pre_merge_findings_gate.py` | 8 passed | 2 (pure composition) + shell test `test-merge-findings-gate.sh`: 9 passed, 1 new case (`--help` fires the guard end-to-end with **no** `MERGE_GATE_TEST_JSON` seam set — proving it never reaches the live, unstubbed `gh pr view` call) |
| `test_pre_merge_numbering_check.py` | 36 passed | 2 (pure composition) |
| `test_no_crude_command_substring_checks.py` (repo-wide AST gate, required for any `claude/scripts/*.py` change) | 13 passed | Confirms `is_merge_help_only`'s implementation introduces no crude `if "<literal>" not in command` anti-pattern — 0 unexpected/stale/duplicated across 47 files |
| `test_pyw_stdio.py` | 7 passed | Unaffected — sanity re-run |

For the 5 PostToolUse files whose new guard sits inside `main()`, a `main()`-driving subprocess test was considered but rejected in favor of the lightweight pure-composition approach above — most of these hooks' `main()` need elaborate fixtures (e.g. `post-pr-merge-project.py` requires a real `.claude/hook-config.json`) to even reach the guard line, and the repo's established convention for this file family is pure-helper-only coverage (`main()` I/O, live `gh` calls, and subprocess spawns are explicitly documented as out of scope in each file's own docstring). `is_merge_help_only` itself is exhaustively covered in `test_hookio.py`; the composition tests prove the two predicates line up the way each guard depends on.

No suppressions, no skipped tests, no lowered thresholds — confirmed via the required pre-PR greps (both clean against `origin/main`).

## Documentation

Filed as [ADR-050 Amendment 13](https://github.com/brownm09/dev-env/blob/fix/issue-557-merge-help-only/docs/adr/050-shared-hookio-sibling-hook-fixes.md#amendment-13-2026-07-04--is_merge_help_only-excluding-non-mutating-gh-pr-merge-invocations-from-the-live-confirmation-fallback-dev-env557) (this ADR already tracks 12 prior amendments to this exact shared-`_hookio.py`-plus-sibling-hooks pattern) — `docs/adr/INDEX.md`'s row 050 date column and tags updated in the same PR.

Closes #557

- **[documentation]** `docs/REFERENCE.md` wasn't updated despite extensive `claude/scripts/**` changes — **fixed in `ca9e897`**: added an `is_merge_help_only`/dev-env#557 clause to each of the 8 affected hooks' trigger-condition rows.


brownm09:
## Review: [fix] Gate merge-hook live-confirmation fallback on gh pr merge --help/-h

### Summary
Ready to merge with one documentation fix applied in-PR. Verified the core logic (`is_merge_help_only` in `_hookio.py`) directly against the design intent — segment-scoped via `split_top_level`, anchored at each segment's first physical line, correctly returns `False` for a chained help-then-real-merge command so a genuine merge is never suppressed. Spot-checked the guard placement in `post-pr-merge-project.py` (early-exit shape) and `pr-merge-reminder.py` (the one structurally different file, folded into an inline boolean condition) — both match spec exactly. 234/234 tests pass, no suppressions, no test-integrity violations. One blocking documentation gap below, now fixed.

---

### Blocking Findings

**[documentation]** `docs/REFERENCE.md`
`claude/scripts/**` changed extensively (a new shared predicate plus guards in 8 hook files), but `docs/REFERENCE.md` — which documents each hook's trigger condition in the Hooks table — wasn't updated in the diff. Three of the eight rows (`post-merge-tile-checkpoint.py`, `post-pr-merge-pull.py`, `post-pr-merge-reclaim.py`) already mentioned `--help` in a way that, post-fix, undersells what's actually guarded (they described only "the exit code isn't consulted," not the live-confirmation-fallback misattribution this PR closes); the other five didn't mention `--help` at all. Per the Documentation Maintenance table in dev-env's `CLAUDE.md`, this is a required update, not optional.
**Fix:** Added a short `is_merge_help_only`/dev-env#557 clause to each of the 8 affected rows' trigger-condition column.

---

### Findings Disposition
Fixed in this PR — see disposition note added to the PR body.

---

PR #555, opened 2026-07-03
[docs] Extend ADR-058 + runbook: sibling-worktree squat blocks gh pr merge --delete-branch

brownm09:
## Summary

- lifting-logbook PR #664 hit `gh pr merge --squash --delete-branch`'s local-checkout failure —
  not from the canonical holding `main` (the case ADR-058's runbook already documents), but from
  an unrelated **sibling worktree** (`fix+issue-646-restrict-db-e2e-default-role`) squatting `main`.
  Confirmed via `git worktree list` that the squat is still live in lifting-logbook.
- Amends [ADR-058](docs/adr/058-worktree-squatting-main-detection-correction.md) with this incident
  and confirmation that its parking mechanism is repo-general (applies outside dev-env), plus a new
  `docs/REFERENCE.md` runbook subsection ("A sibling worktree squatting `main` blocks a different
  merge's `--delete-branch`") documenting:
  - The generalized trigger (any worktree holding `main`, not just canonical).
  - The clean two-step workaround: `gh pr merge <N> --squash` (no `--delete-branch`) then
    `gh api -X DELETE repos/{owner}/{repo}/git/refs/heads/<branch>` — both pure API calls, so
    neither depends on worktree topology.
  - The on-demand remedy to un-squat immediately instead of waiting for the next 8am
    `prune-stale-worktrees` run: `prune-merged-worktrees.py --repo-path <repo>` (verified by reading
    `prune_one()` that the squatter-park check runs unconditionally, before the `--include-named`
    branch-prefix gate).
- Updates `docs/adr/INDEX.md`'s ADR-058 row with the new amendment date.

## Testing

Pure documentation change — no hook, script, skill, or settings file touched, so none of the
`## Testing` items in dev-env's `CLAUDE.md` apply (each is scoped to a specific script/hook file).
Verified manually:
- `git worktree list` in lifting-logbook confirms the cited squat is real and current.
- Read `prune-merged-worktrees.py`'s `prune_one()` directly to confirm the squatter-park check
  (`if path == squatter_path`) runs before the `BRANCH_PREFIX`/`--include-named` gate, so the
  recommended remedy command is accurate as written.
- Markdown renders correctly; internal ADR/REFERENCE cross-links follow the existing relative-path
  convention used elsewhere in both files.

## ADR-warrant check

Warranted — extends [ADR-058](docs/adr/058-worktree-squatting-main-detection-correction.md) (2026-07-03
[...]

PR #551, opened 2026-07-03
fix(skills): journal-compose Step 10.5 misses old-style merge-tree conflict markers

brownm09:
## Summary

Step 10.5 of `claude/skills/journal-compose/SKILL.md` detected draft-branch merge conflicts with `git merge-tree "$MERGE_BASE" HEAD origin/main | grep -c "^<<<<<<<"`. On the installed git (**2.37.1.windows.1**) only the old-style 3-argument `merge-tree` exists, and its diff-style output prefixes conflict markers with `+` (`+<<<<<<< .our`) — the anchored grep matched nothing, so `CONFLICT_LINES` was always 0. Incident: engineering-journal [PR #150](https://github.com/brownm09/engineering-journal/pull/150) passed the check and failed at merge as `CONFLICTING`, requiring manual recovery.

The check is now version-probed:

- **Modern git (>= 2.38):** `git merge-tree --write-tree HEAD origin/main` — exit code 0 = clean, 1 = conflicts ([git-merge-tree docs](https://git-scm.com/docs/git-merge-tree)). Immune to content false positives; activates automatically when git is upgraded, with no skill edit.
- **Old git (< 2.38, the installed 2.37):** the probe exits ∉ {0,1} → fall back to 3-arg merge-tree with `grep -cE "^\+?<<<<<<<"`, which matches the `+`-prefixed markers while still ignoring mid-line *mentions* of markers in journal content.

`CONFLICT_LINES` keeps its `0` / `> 0` contract, so the step's downstream prose (PR_HEAD selection, compose-branch recovery, multi-project note — which references "this check" without repeating the recipe) is unchanged.

Closes #550

## ADR

`docs/adr/080-version-probed-merge-tree-conflict-detection.md` records the root cause, why the fallback pattern must tolerate the `+` prefix (guards against a future "cleanup" re-anchoring the grep), and rejected alternatives. `INDEX.md` row added.

## Testing

Tests: N/A (markdown-only skill change; no dev-env `## Testing` item maps to a SKILL.md-only diff — no hook script, shell script, or Python file touched).

Coverage gate (ADR-022): no testable code artifact — the change is LLM-executed skill instructions; the manual verification below is the equivalent.

Manual verification (throwaway repos under `C:/Users/brown/.claude/scratch/`, git 2.37.1.windows.1):

| Case | Result | Expected |
|---|---|---|
[...]

PR #563, opened 2026-07-04
fix(journal-compose): isolate compose into a dedicated worktree, reconcile open-PR shards, assert journal structure

### LINKED ISSUE #467: Harden journal-compose: validate composed-output headings + document shared-checkout concurrency pattern
## Problem

Two gaps surfaced while backfilling the 2026-06-10 and 2026-06-13 engineering-journal entries in a dedicated session (see engineering-journal PRs #143 and #145):

**1. No automated check that composed output matches the canonical Section 1-11 format.**

\`journal-compose\` Step 5 specifies an exact 11-section "Session Transcript" structure, but nothing in the skill (or in Step 6.5's self-check) actually validates a composed file's headings against that list. A Haiku subagent composing the dev-env 2026-06-10 journal invented its own ad hoc "Overview/Context/Decision/..." structure instead — and it wasn't the only offender: several *already-merged* dev-env journal entries (e.g. \`2026-06-28-workflow-hooks-and-tile-checkpoint.md\`, \`2026-06-27-tile-checkpoint-plan-approval-clarification.md\`) also deviate from the canonical format in their own distinct ways, while every meta-project journal consistently follows it (unsurprising, since meta entries are literally about running this skill). The drift was only caught this session via manual \`grep -E '^#+ '\` comparison against a known-good file — Step 6.5's self-check only verifies line-count fidelity, not structural conformance.

**Impact:** silent, compounding inconsistency across the journal corpus; each drifted entry makes the *next* compose more likely to copy the drift instead of the canonical skill text, since subagents have no automated backstop.

**2. `draft/YYYY-MM-DD` branches are not safe to assume exclusive access to, and the skill doesn't say so.**

While composing both backfill dates, the shared \`engineering-journal\` checkout was mid-flight modified by other concurrent sessions/agents — an in-progress stub file blocked a direct branch checkout (\`git checkout -b ... origin/main\` failed with "local changes would be overwritten"), and separately an unrelated \`sessions/lifting-logbook/open-prs/607.json\` deletion appeared and disappeared during the session. This is expected given how heavily this repo is used (the \`git worktree list\` for engineering-journal currently shows 40+ registered worktrees, many locked/in-use), but the skill's Step 10 assumes the primary checkout can be freely committed to and branch-switched.

**Impact:** a session that doesn't anticipate this can either corrupt/disrupt another session's uncommitted work, or get stuck unable to proceed.

## Suggested fix

1. Add a Step 6.5 (or new Step 5.5) automated heading check: grep the composed file's `^#+ ` lines and diff against the canonical list from Step 5, failing/flagging if any mandatory heading is missing or an unexpected one is present. Apply this both in the single-project flow and in each Multi-project-mode subagent's own self-check.
2. Document the concurrency-safe pattern that worked this session as the standard technique whenever the primary checkout can't be safely branch-switched: `git stash push -u -- <explicit pathspec>` scoped to only the files this compose run itself created/modified (never a bare stash or `git add -A`), then apply that stash inside a `git worktree add <path> -b compose/<slug> origin/main`, commit + push + PR + merge there, then clean up the worktree/branch. Add this as a named runbook in `docs/REFERENCE.md` (alongside the existing Git Workflow Runbooks) and reference it from the journal-compose SKILL.md Step 10 as the fallback when the primary checkout's branch differs from the compose date or shows unexpected concurrent modifications.

## Why this matters

Both gaps were discovered by manual investigation in a session that happened to be careful about verifying subagent output and shared-checkout state — a less careful session would have either shipped a mis-formatted journal silently, or disrupted another concurrent session's uncommitted work. Neither failure mode is currently caught by any test or self-check.

brownm09:
## Summary

Closes #467.

`/journal-compose` ran every operation — stub discovery, validators, subagent output, README
edits, deletions, commit/push/PR — directly against the shared canonical checkout
`C:/Users/brown/Git/engineering-journal`, which is heavily concurrent (40+ registered
worktrees). On 2026-07-03, a concurrent session ran `checkout main && pull` there mid-compose
(engineering-journal [PR #150](https://github.com/brownm09/engineering-journal/pull/150)):
two parallel composer subagents aborted on missing stubs, one composed a wrong journal from
stray stubs left on `main`, one wrote its output into the now-wrong-branch tree. Recovery was
manual (`git worktree add` + re-point composers + rerun Steps 7–11).

Three changes, one PR, per [ADR-082](docs/adr/082-journal-compose-worktree-isolation.md):

1. **Worktree isolation** — a new Step 0.6 creates a dedicated, disposable, **detached**
   worktree (`engineering-journal/.claude/worktrees/compose-YYYY-MM-DD`, built from
   `origin/draft/YYYY-MM-DD`) before any read/write; every subsequent step operates on that
   worktree, never the canonical. The canonical is touched only by read-only
   `fetch`/`show-ref`/`ls-remote`/`ls-tree` queries, `worktree add/remove` registration, and a
   post-merge local-branch cleanup (which git itself no-ops if the branch is still checked out
   elsewhere). Also fixes the merge step: `gh pr merge --delete-branch` fails outright if the
   branch is checked out as a named branch anywhere in the repo (a stub-writing session's
   worktree can hold `draft/YYYY-MM-DD` at the exact moment compose merges) — split into a
   squash-merge + a pure `gh api -X DELETE` ref delete instead, per the runbook landed in #555.
2. **Open-PR shard hygiene** — compose PRs #147 and #149 each self-wrote their own
   `open-prs/<N>.json` shard (the generic `pr-merge-reminder` hook's advice fires on any `gh pr
   create`, including the compose PR's own) and those went stale the instant they merged. Fix:
   a new Step 9.5 explicitly reconciles each project's shards (`gh pr view` state check → `git
   rm` if merged/closed, continuing PR #150's own verification precedent), no shard is written
   for the in-session compose PR at all, and a post-merge check surfaces (never silently fixes)
   any leak that lands on `origin/main` anyway.
3. **Structural assertion** — a grep-anchored check against the canonical 11-section heading
   list now runs in the single-project self-check and in a new subagent Step 6.6; the Phase 2
   coordinator treats a structure failure the same as a failed subagent (blocks all
   README/git work until fixed). Closes the other half of #467 (a composer subagent had
   previously emitted ad hoc section numbering with no `## Next Session Context`).

**Companion:** `daily-journal-compose` drops its Step 0 canonical-checkout sync (itself a
mutator of this incident's exact class) for remote-based stub discovery, and collapses its
per-project loop into a single `/journal-compose` call (the skill's own multi-project mode
already fans out internally).

**Also fixed while implementing:** the draft-branch recovery runbook (`docs/REFERENCE.md`)
told operators to point `/journal-compose` at a `draft/YYYY-MM-DD-recovery` branch by manually
positioning "the working tree" — meaningless once compose builds its own worktree. Step 0.6's
date resolution now accepts a full branch name (not just a bare date) so the recovery runbook
can invoke `/journal-compose draft/YYYY-MM-DD-recovery` explicitly.

**Known follow-ups, deferred (flagged on #467, tiles to be filed post-merge):** the
`daily-journal-compose` routine (and a separate `journal-compose-with-retry.sh` Task
Scheduler wrapper) compute `DATE` as *today* and never pass `--force`, so the automated 7am
run likely never successfully composes anything — a real, pre-existing defect orthogonal to
this PR's concurrency scope; fixing it requires a product decision (compose *yesterday*
instead? always force?) that isn't mine to make silently. Also: a sweep of ~24 stale remote
`draft/*` branches, and a revisit of `reconcile-open-prs.py`'s now-permanently-dirty
canonical-checkout unlinks (harmless — they converge to clean on the next `main` pull — but
worth cleaning up the hook itself).

## Testing

dev-env `## Testing` item 1 (hook-script `ast.parse` syntax check) run — no `.py`/`.sh` files
changed, clean no-op pass. Item 4 (docs-only `date -u` guard, since `claude/CLAUDE.md` was
touched) — zero matches, trivially clean. Items 2–39 N/A (no matching files touched). No
automated tests exercise skill/routine markdown; verification was a full occurrence-grep of
the edited skill confirming every remaining canonical-checkout path reference is on the
read-only/registration allowlist (all committed and pushed), plus a manual step-by-step
consistency walkthrough of the rewritten skill. The real end-to-end verification is the next
live compose, which should be watched under supervision.

Tests: no automated test suite for skill markdown; ast.parse clean, date -u guard clean (see above).

## Risk-dimension audit

1. **Testing** — see above; N/A beyond the two gates run.
2. **Observability** — N/A per dev-env's `## Observability` (no runtime; the skill's own
   step-by-step user-facing messages, including the new `STRUCTURE=` and `RECONCILED_SHARDS=`
   reports, are its diagnostic surface).
3. **Security** — N/A; no new credentials/secrets/auth surface (`gh pr view`/`create`/`merge`/
   `api` calls are the same class the skill already made).
4. **Resilience / failure modes** — improves failure isolation (a merge failure now leaves a
   self-contained, prune-safe worktree behind instead of an ambiguous canonical state);
   restart-safe (a stale compose worktree is detected and recreated by the next invocation).
5. **Performance** — one additional `git worktree add`/`remove` pair and a handful of bounded
   `gh pr view` calls per compose; negligible next to the subagent compose cost itself.
6. **Data integrity & migrations** — N/A; manifest/open-PR shard JSON schemas (ADR-056)
   unchanged.

## ADR

[ADR-082](docs/adr/082-journal-compose-worktree-isolation.md) — Journal-Compose Worktree
Isolation, Shard Reconciliation, and Structure Assertion. (Note: this PR's design work was
renumbered twice by unrelated concurrent PRs claiming the same number before either of us
merged — first ADR-080 (claimed by #551's Step 10.5 conflict-detection fix; this PR's own
Step 10.5 edit preserves that fix's version-probed `merge-tree` logic unchanged while
relocating it onto the compose worktree), then ADR-081 (claimed by #561's independent
write-time shard-validation work, discovered by reading that PR's own engineering-journal
stub before opening this one). Landed on 082 after confirming no other open PR claims it.)

`/review` posted at [#issuecomment-4880935594](https://github.com/brownm09/dev-env/pull/563#issuecomment-4880935594)
(two parallel subagents: correctness/security, reliability/performance/maintainability).
5 blocking + 5 non-blocking findings, all fixed in `e6442d3`:

- **[correctness] Shell-variable persistence across separate Bash calls** ($EJ/$WT/$SOURCE_BRANCH/$PREV) — fixed in `e6442d3`: added an explicit note that these are not persisted shell variables.
- **[correctness] Step 10.5 shard-resurrection leak** — fixed in `e6442d3`: empirically confirmed `git checkout <commit> -- <dir>` does not delete files absent from `<commit>`; converted the prose "re-apply reconciliation" comment into an executable removal loop (both single- and multi-project paths).
- **[reliability] Push-failure routing to Step 10.5 not explicit** — fixed in `e6442d3`: made the jump to the `CONFLICT_LINES > 0` recovery block explicit on a merged-draft-branch rejection.
- **[reliability] Step 0.6 worktree-creation TOCTOU race** — fixed in `e6442d3`: added a `.compose-creating` sentinel written immediately after `worktree add`.
- **[correctness] Step 11 post-merge shard-leak check greps the wrong placeholder** — fixed in `e6442d3`: rewritten to iterate the actually-reconciled shard numbers instead of the compose PR's own (by-design nonexistent) shard.
- **[reliability] Remote branch delete could race a live stub session** (non-blocking, pre-existing) — fixed in `e6442d3`: added a caution note; a full fix is a larger design change out of scope for this PR.
- **[maintainability] Silent shard on unparseable `url`** (non-blocking) — fixed in `e6442d3`: explicit warning echo.
- **[maintainability] Duplicated Step 6.5/6.6 structure-check block, no sync marker** (non-blocking) — fixed in `e6442d3`: added "keep in sync" comments to both copies.
- **[maintainability] Step 8a hardcoded root literal vs. `$WT`** (non-blocking) — fixed in `e6442d3`: now interpolates `$WT` directly (matching the existing `$TMPFILE` pattern).
- **[performance] Redundant `fetch origin` in Step 0.6** (non-blocking) — fixed in `e6442d3`: dropped the tier-3 fetch.


brownm09:
## Review: fix(journal-compose): isolate compose into a dedicated worktree, reconcile open-PR shards, assert journal structure

### Summary

The core design is sound — the worktree-before-branch-delete ordering, the detached-worktree
rationale, and the two-call merge (avoiding `--delete-branch`'s local-checkout failure) are all
correct, and quoting/injection surface is clean throughout. However, this needs revision before
merge: five blocking findings, the most serious of which (shard-resurrection leak, and a
post-merge leak-check that can structurally never fire) directly undermine Change 2's own safety
guarantee, and a new TOCTOU race in Step 0.6 introduces exactly the failure class — a concurrent
compose corrupting another's state — this PR exists to eliminate. At 961 changed lines across 7
files, the PR is appropriately scoped (one skill + its direct companions); it does not need to be
split.

---

### Blocking Findings

**[correctness]** `claude/skills/journal-compose/SKILL.md` Step 0.6 (`$EJ`/`$WT`/`$SOURCE_BRANCH`) and Step 10.5 (`$PREV`)
These shell variables are defined once in Step 0.6 (and once more, inline, in Step 10.5's
recovery block for `$PREV`) but referenced in Steps 9.5, 10, 10.5, and 11 — each of which the
executing agent runs as a *separate* Bash tool call, and the Bash tool's own documentation states
shell state does not persist between invocations.
If the executing agent treats `$WT`/`$EJ`/`$SOURCE_BRANCH` as literal persisted environment
variables rather than symbolic references to values it must carry in its own context (the way it
already must for `<project>`/`YYYY-MM-DD`/`<slug>` placeholders elsewhere in this same document),
an unset `$WT` silently retargets `git -C ""` at the wrong repository, or produces a malformed
push refspec — defeating the isolation guarantee this PR exists to provide, silently.
**Fix:** Add an explicit note immediately after the `$EJ`/`$WT` definitions in Step 0.6 stating
that these are not persisted shell variables — the executing agent must carry the resolved
literal values (worktree path, source branch name) across steps in its own context and substitute
them concretely into every subsequent command, exactly as it already does for the `<placeholder>`
tokens used throughout the rest of the skill.

**[correctness]** `claude/skills/journal-compose/SKILL.md` Step 10.5, conflict-recovery block (`git checkout "$PREV" -- sessions/<project>/open-prs`)
Empirically verified (throwaway git repo): `git checkout <commit> -- <directory>` does **not**
delete files present in the current working tree but absent from `<commit>` at that path — it
only adds/updates files that exist in `<commit>`. After `checkout -b compose/YYYY-MM-DD
origin/main`, the working tree has `origin/main`'s *stale* `open-prs/` (still containing shards
Step 9.5 already reconciled away in `$PREV`); the subsequent `checkout "$PREV" -- .../open-prs`
does **not** remove those stale shards, since they aren't in `$PREV`'s superset either way — they
simply remain. The skill's own "2b" note describes this exact resurrection risk but leaves the fix
as a prose comment ("re-apply... if the checkout brought it back, remove it again") with no
executable mechanism and no clear trigger for when to run it.
Every recovery-path compose silently re-leaks already-reconciled shards back into the PR it's
about to open — reintroducing the staleness (PRs #147/#149) this entire change exists to fix, on
precisely the code path (conflict recovery) most likely to run during a busy day.
**Fix:** Convert the "2b" comment into an actual executable step: after the `open-prs`/
`open-prs.jsonl` checkout, loop over this run's `RECONCILED_SHARDS` entries and `git -C "$WT" rm
--quiet -- sessions/<project>/open-prs/<N>.json` for each one still present. Apply the same fix to
the multi-project variant.

**[reliability]** `claude/skills/journal-compose/SKILL.md` Step 10 / Step 10.5 routing
The pre-push hook's merged-draft-branch block causes Step 10's `push` to fail outright (the commit
already succeeded; only the push is rejected). Both the Step 0.6 push-failure rule and Step 10's
note say this failure "routes to Step 10.5's `compose/YYYY-MM-DD` recovery," but nothing in Step
10.5 is triggered by a failed push — Step 10.5 is written to run only *after* a successful push,
as a merge-tree conflict probe. The routing is asserted, not instructed.
On a genuine prior-day recompose (the exact #147-morning/#150-evening shape this PR calls out by
name), the executing agent hits a rejected push with no in-band instruction for what "routes to
Step 10.5" concretely means — it may retry the same push, or stall.
**Fix:** In Step 10, state explicitly that a merged-draft-branch rejection means: proceed directly
to Step 10.5's `CONFLICT_LINES > 0` recovery block (skip the merge-tree probe — treat the
rejection as equivalent to a detected conflict).

**[reliability]** `claude/skills/journal-compose/SKILL.md` Step 0.6, pre-existing-worktree check
The stale-vs-active determination for a pre-existing `$WT` relies solely on `.draft-compose.lock`
freshness — but that lock is written by Step 1 (or the subagent's own Step 1), which runs *after*
Step 0.6 already created the worktree. A second compose invocation landing in the window between
"worktree created" and "lock written" sees no lock at all, concludes "stale," and runs `worktree
remove --force` — destroying the first invocation's worktree mid-run.
This is a new race condition, introduced by this PR's own Step 0.6, in exactly the failure class
(one compose corrupting another's in-flight state) this PR exists to eliminate.
**Fix:** Write a lightweight creation sentinel inside `$WT` immediately after `worktree add`
succeeds (before Step 0.6 returns), e.g. a timestamped `.compose-creating` file; have the
freshness check treat a sentinel younger than the same 600s window as "active" too, not just a
lock file.

**[correctness]** `claude/skills/journal-compose/SKILL.md` Step 11, post-merge shard-leak check
The check greps `origin/main`'s tree for `open-prs/<N>.json` where `<N>` is presented as the
*compose PR's own number* — but Decision §2's entire point is that the compose PR never gets a
shard, so a shard named after the compose PR's own number never exists by design. As written, the
check greps for something that structurally can never be found, so it always reports clean —
including on the exact resurrected-shard leak described above.
A safety-net check that can never fire is worse than no check: it creates false confidence that
shard leaks are being caught when none ever will be.
**Fix:** Rewrite the check to iterate this run's `RECONCILED_SHARDS` (the shards actually at risk
of resurrection via the Step 10.5 path above) and assert each one's `open-prs/<N>.json` is absent
from `origin/main`, not a check keyed on the compose PR's own number.

---

### Non-Blocking Findings

**[reliability]** `claude/skills/journal-compose/SKILL.md` Step 10.5, remote branch deletion (`push origin --delete "$SOURCE_BRANCH"`)
Deleting the remote `$SOURCE_BRANCH` while a stub-writing session's worktree still holds it (and
intends to push more stubs there that same day) silently destroys the branch that session will
push to next, forcing it to recreate a branch that lacks the composed journal. This delete
predates this PR (not a regression), but the ADR's own same-day-recovery framing now explicitly
invites this shape to run more often.
**Suggestion:** Add a one-line caution in Step 10.5 that this delete assumes no further stubs will
be written for the date (an end-of-day invariant); a narrower fix (e.g. only delete once no other
worktree holds the branch) is a larger design change out of this PR's scope.

**[maintainability]** `claude/skills/journal-compose/SKILL.md` Step 9.5, `$REPO` derivation
When a shard's `url` field is empty or malformed, `$REPO` resolves empty, `gh pr view --repo ""`
fails, and the shard is conservatively kept — a safe outcome, but silent. A permanently
unreconcilable shard accumulates forever with no signal.
**Suggestion:** Echo a warning naming the shard when `$REPO` is empty after the `sed`, so the gap
is visible instead of silently perpetual.

**[maintainability]** `claude/skills/journal-compose/SKILL.md` Step 6.5 and Step 6.6
The 11-pattern structural-assertion block is duplicated verbatim between the main-flow (Step 6.5)
and subagent-template (Step 6.6) copies, with no "keep in sync" marker (unlike the existing Step
0.5/subagent-template pairing, which already has one).
**Suggestion:** Add a "mirrors Step 6.5 — keep in sync" comment to the Step 6.6 copy.

**[maintainability]** `claude/skills/journal-compose/SKILL.md` Step 8a
The dashboard script hardcodes `const root = '.../.claude/worktrees/compose-YYYY-MM-DD'` with a
manual "substitute the compose date" comment, while every other post-Step-0.6 command uses the
already-resolved `$WT` shell reference — two representations of the same path, doubling the
chance of a stale-date mismatch if the worktree path scheme ever changes.
**Suggestion:** Pass `$WT` into the `node -e` invocation via an environment variable
(`WT="$WT" node -e '... process.env.WT ...'`) instead of a second hardcoded literal.

**[performance]** `claude/skills/journal-compose/SKILL.md` Step 0.6
`git -C "$EJ" fetch origin` runs once in the remote-scan tier (tier 3) and again, unconditionally,
right before worktree creation — back-to-back with no intervening mutation. `git ls-remote`
queries the remote directly and never needed the first fetch at all.
**Suggestion:** Drop the tier-3 fetch; the unconditional fetch immediately before worktree
creation already covers it.

---

### Findings Disposition

Five blocking + five non-blocking findings above; all will be fixed in this PR (see follow-up
commit(s) — disposition will be added to the PR body per ADR-028 before merge).

---

*Reviewed by `claude-sonnet-5` via Claude Code (two parallel Opus subagents: correctness/security,
reliability/performance/maintainability; one finding — the shard-resurrection leak — was
empirically verified against a throwaway git repository before inclusion)*

Issue #574, opened 2026-07-05
Design a mechanical pre-check gate for `gh pr merge --auto` (ADR-031 prerequisite)


brownm09:
## Problem

ADR-031's 2026-07-04 addendum ([dev-env#565](https://github.com/brownm09/dev-env/issues/565))
resolved the "Per-PR escape hatch" (`gh pr merge --auto`) as **permanently inert by design**,
because the repo-level `allow_auto_merge` toggle is currently the *only mechanical backstop*
preventing `--auto` from being invoked before the six in-session merge-time rules
(`usage-snapshot.py`, the post-merge journal stub, ADR-warrant checkpoint 3, doc-reconciliation
checkpoint 3, the `/review` all-findings gate, project-board automation) have completed.

The addendum named the exact condition that would justify revisiting this:

> a mechanical pre-check that verifies, at the moment `--auto` is invoked, that `/review`,
> ADR-warrant checkpoint 3, and doc-reconciliation checkpoint 3 have already completed in the
> current session.

Only these three, because the other three rules are tied to the physical (async, out-of-session)
merge moment itself — no pre-check can create an in-session moment for something `--auto`
inherently defers.

## Scope of this issue

Research and design (not yet implement) that pre-check: a `PreToolUse` hook on `Bash` that gates
any `gh pr merge ... --auto` invocation.

Open questions to resolve, evaluated against the closest prior art
([ADR-039](docs/adr/039-merge-gate-findings-enforcement.md), which already mechanically enforces
the `/review` all-findings gate for *plain* merges):

1. How does a hook mechanically verify "`/review` already ran" and "ADR-warrant / doc-reconciliation
   checkpoints were satisfied" — today these are behavioral disciplines a session follows, not
   externally observable state.
2. Candidate mechanisms: (a) a session-local marker file written by `/review` and the checkpoint
   steps; (b) parsing the PR's own comment/body history via `gh pr view --json comments` for a
   marker (ADR-039's existing pattern); (c) something else.
3. Whether the new gate should fail open or fail closed on a `gh`/network error, given it's
   protecting the one path that removes every other in-session safety net (unlike ADR-039's gate,
   which fails open because the routine merge path still has a live session as backup).
4. Whether shipping the hook is sufficient on its own, or whether it's only a prerequisite for a
   separate, later decision to flip the per-repo `allow_auto_merge` toggle.

## Deliverable

This is exploratory/research. A well-reasoned ADR draft proposing the mechanism, trigger,
[...]

PR #572, opened 2026-07-04
[fix] Resolve gh pr merge repo from the command's PR URL, not session cwd

brownm09:
## Summary

`post-pr-merge-project.py` always resolved which repo to act on via `load_config(cwd)` — the PostToolUse hook payload's `cwd` (the session's pinned working directory) — never from the `gh pr merge` command itself. `extract_pr_number_from_command()` already correctly parsed a full PR URL argument (e.g. `gh pr merge "https://github.com/owner/repo/pull/554" --squash`) for the PR **number**, but repo resolution ignored that same URL entirely.

**Confirmed incident:** a session pinned to `lifting-logbook`'s cwd ran `gh pr merge "https://github.com/brownm09/dev-env/pull/554" --squash --delete-branch` (a dev-env PR, no `cd` involved). The hook resolved `repo=lifting-logbook` from cwd, fetched **lifting-logbook's own PR #554** instead of dev-env's, parsed its unrelated `Closes #537`, and moved that (coincidentally already-Done) lifting-logbook issue to Done on lifting-logbook's own project board. Harmless only by coincidence of timing — a still-open issue would have been silently corrupted with no error.

## Fix

- `extract_repo_from_command()` parses the owner/repo out of a PR URL argument, mirroring the existing PR-number extraction (same `_MERGE_ARGS_RE`-scoped region, so a `--subject`/`--body` value or a chained sibling command can't hijack it). An explicit `--repo` flag is checked first (highest-confidence signal, mirroring `_effective_merge_repo` in `pr-merge-reminder.py`).
- `main()` now skips the whole operation when the parsed repo differs from `config.get("repo")` — **not just a "use the right repo for `get_pr_body`" fix**. `find_project_item`/`move_to_done` use cwd's own `project_number`/`project_node_id`/`status_field_id`/`done_option_id`, which are scoped to cwd's repo and don't apply to a different one regardless of which PR's body gets fetched. Fixing `repo` alone would still leave a same-numbered-issue collision able to reproduce the identical corruption via a different trigger — skipping outright is the safe behavior, consistent with `effective_merge_dir`'s existing "under-corrects rather than mis-fires" philosophy (ADR-067).
- A bare `gh pr merge --squash --delete-branch` (no URL) is unchanged — falls back to `config.get("repo")` exactly as before.

## Root cause note — amends ADR-067

[...]

PR #575, opened 2026-07-05
[docs] Propose mechanical pre-check gate design for gh pr merge --auto (ADR-083)

brownm09:
## Summary

Design-only PR (no hook code) proposing how to satisfy the condition ADR-031's
2026-07-04 addendum set for revisiting the `--auto` escape hatch: a mechanical pre-check that
verifies, at the moment `--auto` is invoked, that `/review`, ADR-warrant checkpoint 3, and
doc-reconciliation checkpoint 3 have already completed.

Adds [`docs/adr/083-auto-merge-checkpoint-gate.md`](docs/adr/083-auto-merge-checkpoint-gate.md)
(Status: **Proposed**) and its `INDEX.md` row.

**Key findings from the research:**

- Two of the three checkpoints (ADR-warrant, doc-reconciliation-at-checkpoint-3) have **no
  mechanical artifact anywhere today** — `/review`'s SKILL.md has no ADR-warrant step at all, and
  doc-reconciliation's only mechanical presence is the earlier Step 2b early-warning, not a
  checkpoint-3 record. Designing that missing artifact is the actual hard part, not hook plumbing.
- Recommended mechanism: extend `/review`'s existing marker-comment pattern (the same one
  [ADR-039](docs/adr/039-merge-gate-findings-enforcement.md) already ships for the findings gate)
  rather than a session-local marker file — a PR-durable artifact survives worktree/session
  churn, stays human-auditable on the PR, and correctly treats "was this PR (in its current form)
  evaluated" as a property of the PR, not of any one session.
- The new gate should **fail closed** (opposite of ADR-039's fail-open), with **no override
  token** (unlike the canonical-mutate-guard's `ALLOW_CANONICAL_MUTATE=1`) — both departures are
  explained in the ADR's Decision section, points 3–4.
- A **freshness check** (marker `createdAt` vs. the PR's head-commit `committedDate`, both fields
  verified live against a real PR during this design) closes a staleness gap ADR-039's own gate
  doesn't need to close, because `--auto` removes the in-session backstop that makes that gap
  tolerable there.
- Shipping the hook is **necessary but not sufficient** — `allow_auto_merge` stays `false`
  everywhere regardless; flipping it per-repo is explicitly scoped as separate, later work.

Full alternatives-considered (session-local file, human-label approval, a CI verification job,
extending `pre-merge-findings-gate.py` in place) are in the ADR's "Candidates evaluated" and
"Alternatives Considered" sections.

## What this PR does NOT do

No hook script, no `/review` SKILL.md changes, no `settings.json` wiring. The ADR's "Follow-up
work" section itemizes exactly what's left (7 items) and is intentionally not done here — per
[dev-env#574](https://github.com/brownm09/dev-env/issues/574), the ask was research + design,
with implementation as explicit follow-up. This PR lands the design only; issue #574 stays open
until the implementation items land.

## Test plan

Docs-only change (no `.py`/`.sh` files touched), so none of the numbered items in dev-env's
`## Testing` section apply. Manually verified:

- `docs/adr/INDEX.md` numbering: 083 was the next free number as of `origin/main` at branch-cut
  time (`34cb4fe`); `pre-merge-numbering-check.py` will re-verify against `origin/main` at actual
  merge time per its own design.
- The two live-verified `gh` field-name claims in the ADR (comment `createdAt`, commit
  `committedDate`) were checked against a real merged PR (#572) during this session, not assumed.

Tests: no automated tests apply (docs-only ADR addition).

[`/review` posted](https://github.com/brownm09/dev-env/pull/575#issuecomment-4887741019): 1 blocking, 3 non-blocking. All four fixed in this PR, no filing needed (all were prose-only clarifications to a not-yet-implemented design):

- **[correctness] Ambiguous marker co-location** — fixed in `151da67`: made explicit that the hook must find one comment carrying *both* markers together, not two independently-found "last comment per marker," and that freshness is checked against that one comment.
- **[documentation] Overstated "verified live" claim for `--auto=<value>`** — fixed in `151da67`: split the claim into what was actually confirmed live (bare `--auto` boolean) vs. assumed by Cobra/pflag convention and this repo's `--delete-branch=false` precedent (`=value` support); added a follow-up test item to close the assumption for real.
- **[documentation] PR comments aren't an authenticated channel** — fixed in `151da67`: added to Consequences' accepted-limitations list, cross-referenced from the "No override token" section.
- **[maintainability] `adr_warrant=filed` terminology collision** — fixed in `151da67`: renamed to `adr_warrant=written`.

Related: #574, ADR-031, ADR-039, ADR-011, ADR-019


brownm09:
## Review: [docs] Propose mechanical pre-check gate design for gh pr merge --auto (ADR-083)

### Summary
Solid, well-reasoned design proposal — merge with minor changes. The core recommendation (extend `/review`'s existing marker pattern rather than a session-local file) is well-argued and the fail-closed/no-override departures from ADR-039 are correctly justified by the actual difference in what each gate protects. One blocking finding: the spec leaves an implementation-relevant ambiguity in exactly which PR comment the two markers and the freshness check must be read from, which could let a future implementation build a version that's bypassable in a way the ADR's own design goals rule out. Three small non-blocking precision/terminology points round it out. Diff is appropriately scoped (389 lines, a single new ADR + its INDEX row).

---

### Blocking Findings

**[correctness]** `docs/adr/083-auto-merge-checkpoint-gate.md:215-224` and `:270-274`
The pass condition requires (a) a `review-findings` marker with a clean/disposed review, (b) "the **same** comment" also carrying a `premerge-checkpoints` marker, and (c) freshness checked against "the marker comment's `createdAt`." But the document never says how an implementation should *locate* that comment. `pre-merge-findings-gate.py`'s existing algorithm scans all comments and keeps whichever is *last* to carry the `review-findings` marker — a natural, easy-to-copy implementation for this new hook would independently do the same "last comment carrying `review-findings`" search and a separate "last comment carrying `premerge-checkpoints`" search, rather than requiring one single comment to carry both. In the normal flow both markers are emitted together by the same Step 8 template so this never diverges — but if it *did* (a re-review whose second marker line was somehow dropped, or a manual comment), the two-independent-searches implementation would accept a fresh clean review paired with a stale ADR-warrant/doc-reconciliation disposition evaluated against old code — exactly the bypass the freshness check (point 5) exists to close, just reached from a different angle. It's also unclear which comment's `createdAt` the freshness check should even use if the two markers could legitimately live on different comments.
**Fix:** Add a sentence making the single-comment requirement explicit and mechanical: the hook must find the single most recent comment that carries *both* markers together (not the most recent comment for each marker independently), and freshness is checked against that one comment's `createdAt`. A comment carrying only one of the two markers does not satisfy the gate.

---

### Non-Blocking Findings

**[documentation]** `docs/adr/083-auto-merge-checkpoint-gate.md:190-192`
"Cobra booleans accept the `=value` form even though `gh pr merge --help` only documents the bare flag; verified live via `gh pr merge --help`..." — the live check actually confirmed only that `--auto` is a plain boolean flag with no short form; it did not and could not confirm `=value` support (`--help` output doesn't show that). The `=value` claim is a reasonable inference from general Cobra/pflag convention and this repo's own precedent for `--delete-branch=false` in `is_mutating_gh_segment`, but the sentence's "verified live" currently reads as covering the whole claim. Given the rest of this document (and its References section) is unusually careful to distinguish "verified live" from "assumed by precedent," this one sentence undercuts that standard.
**Suggestion:** Split the claim: "`--auto` is confirmed live as a plain boolean with no short form; `=value` support is assumed by Cobra/pflag convention and this repo's existing `--delete-branch=false` precedent, not independently re-verified against `--auto` specifically." Worth an explicit note in "Follow-up work" item 3 to actually exercise `--auto=false` against a real `gh` invocation once the hook is implemented, closing the assumption for good.

**[documentation]** `docs/adr/083-auto-merge-checkpoint-gate.md:253-268` (Decision point 4) and `:312-334` (Consequences)
"No override token" argues the three checkpoints are "either satisfiably evaluated... or they are not," implying the only way to satisfy them is genuinely running `/review`. But per ADR-039's own already-shipped design (which this ADR extends unchanged), the hook only regexes comment *bodies* — it never checks comment authorship. Anyone or anything with comment access to the PR can post a hand-written comment containing the exact marker text and satisfy the gate without `/review` ever running, which is functionally equivalent to the bypass "no override token" is trying to foreclose. This isn't a new hole ADR-083 introduces (ADR-039 already has it, unremarked), but ADR-083's gate leans on it more heavily — it's the *sole* thing standing between `--auto` and an unattended merge, with no override and no fail-open safety valve, so the same gap matters more here than it did for ADR-039's routine, fail-open, human-supervised path.
**Suggestion:** Add a line to the Consequences' "Negative / accepted limitations" list naming this explicitly: PR comments are a procedural convention, not an authenticated channel; the gate assumes the operator (not an adversarial third party) controls what gets posted to their own PRs, consistent with ADR-039's identical, already-accepted assumption. Worth a sentence rather than a design change — real authentication would be disproportionate scope creep for a single-maintainer repo, but the document should say so rather than let "no override token" read as a stronger guarantee than it is.

**[maintainability]** `docs/adr/083-auto-merge-checkpoint-gate.md:170`
`adr_warrant=filed` risks being misread as "an issue was filed" — this document and the wider CLAUDE.md corpus use "file" almost exclusively for GitHub issues ("file a follow-up issue," "file it," etc.), while this value means "an ADR was written/included in this PR."
**Suggestion:** Rename to `adr_warrant=written` or `adr_warrant=included` to avoid the collision; cheap to change now since no code exists yet that depends on the exact string.

---

### Findings Disposition

Per ADR-028, all four findings above will be fixed directly in this PR (all are prose-only clarifications to a not-yet-implemented design — no code depends on the current wording) before merge. Disposition will be recorded in the PR body once applied.

---

PR #579, opened 2026-07-05
[fix] Nightly journal-compose targets yesterday's local date, not today's UTC

### LINKED ISSUE #577: fix: nightly journal-compose today-guard blocks every automated run (yesterday-vs-force decision)
## Problem

Both automated nightly-compose entry points resolve their target date to **today** and never
pass `--force`. Since `journal-compose/SKILL.md` Step 0.6's today-guard (ADR-017) refuses to
compose a same-day branch without `--force`, the automated ~7am run has almost certainly never
successfully composed a journal.

1. `claude/routines/daily-journal-compose/SKILL.md` computes `DATE=$(date -u +%Y-%m-%d)`
   (today, **UTC**) and invokes `/journal-compose ${DATE}` — no `--force`.
2. `claude/scripts/journal-compose-with-retry.sh` (the Windows Task Scheduler wrapper documented
   in `docs/REFERENCE.md` under "daily-journal-compose") invokes a bare
   `claude -p "Run /journal-compose. Merge the result. Create a stub for today."` — no date
   argument at all (falls through to Step 0.6's implicit branch-detection, which also resolves
   to today), and no `--force` either.

Both were discovered while implementing PR #563 (ADR-082, journal-compose worktree isolation)
but deliberately left unfixed there — see ADR-082 → "Judgment calls → Routine's DATE/--force
mismatch is explicitly out of scope" — since the fix requires a product decision, not a
mechanical one.

## The decision was already made once — it just drifted

[ADR-017](docs/adr/017-journal-compose-today-guard.md) already states the intended design:

> The `daily-journal-compose` nightly routine is unaffected: it runs at midnight UTC and always
> passes yesterday's local calendar date — never today's — so the guard never fires for it.

The routine as it exists today does neither: it computes **today's UTC date**, not **yesterday's
local date**. `docs/REFERENCE.md` confirms the routine's schedule is actually `0 7 * * *`
(7:09am **local**, with scheduler jitter) — not midnight UTC as ADR-017 assumed — which makes
"yesterday, local" the correct fix rather than a new design: at 7am local, the previous local
calendar day is genuinely complete, so composing *that* day requires no override of the
today-guard at all.

## Proposed fix

- Both entry points target **yesterday's local calendar date** (`date -d yesterday +%Y-%m-%d`,
  no `-u`), never "today" — restores ADR-017's original intent and requires no `--force`.
- This also fixes a secondary bug: the routine's `date -u` (UTC) computation can disagree with
  the **local**-time convention used for stub filenames and branch names
  (`claude/CLAUDE.md`'s stub-filename convention), which the switch to local-time `date`
  resolves as a side effect.
- `--force` is deliberately *not* adopted as the fix — permanently overriding the guard for every
  automated run would defeat its purpose (catching an in-progress day) for the one caller most
  likely to hit it silently and repeatedly.

## References

- ADR-017: `docs/adr/017-journal-compose-today-guard.md`
- ADR-082: `docs/adr/082-journal-compose-worktree-isolation.md` (→ "Judgment calls → Routine's
  DATE/--force mismatch is explicitly out of scope")
- dev-env#467, PR #563 (where this was discovered and deferred)

brownm09:
## Summary

Both automated nightly-compose entry points resolved to **today's** date and never passed
`--force`, so `/journal-compose`'s today-guard (ADR-017) silently refused every automated run:

- `claude/routines/daily-journal-compose/SKILL.md` computed `DATE=$(date -u +%Y-%m-%d)` — today,
  in UTC.
- `claude/scripts/journal-compose-with-retry.sh` (the Windows Task Scheduler wrapper) invoked a
  bare `/journal-compose` with no date at all, falling through to an implicit branch-detection
  scan that also resolves to today.

[ADR-017](docs/adr/017-journal-compose-today-guard.md) already documents the intended design —
the routine should "always pass yesterday's local calendar date... so the guard never fires for
it." The shipped code drifted from that intent. This PR restores it rather than introducing a new
design, and fixes a secondary UTC/local mismatch (stub filenames and branch names are local time
per `claude/CLAUDE.md`) as a side effect.

- Both entry points now compute `DATE=$(date -d yesterday +%Y-%m-%d)` — local time, yesterday —
  and pass it explicitly to `/journal-compose`. No `--force` is needed or added: yesterday is
  always a genuinely complete day at the routine's ~7am-local fire time, so the guard never fires
  by construction, rather than being permanently overridden for every automated run (which would
  defeat its purpose for the caller most likely to need it).
- `LOG_FILE`/`log()`'s UTC timestamps in the retry wrapper are unchanged — that's correctly UTC
  per `claude/CLAUDE.md`'s carve-out for internal operational artifacts.
- Added [ADR-084](docs/adr/084-nightly-compose-targets-yesterday.md) documenting the
  yesterday-vs-`--force` decision (this was flagged as an open product decision in ADR-082's
  "Judgment calls" section — this PR resolves it).

Closes #577. Discovered during #563/ADR-082 and deliberately deferred there.

## Testing

No `.py` files changed — the Python hook-script test suite (dev-env `## Testing` items 1–3, 8–41)
is not applicable. Ran the items that do apply:

- **Item 5 (script path-hygiene lint)** — required for the `claude/scripts/*.sh` change:
  `bash claude/scripts/tests/check-script-path-hygiene.sh` → `path-hygiene: clean (13 shell files
  scanned; no $HOME-rooted path passed to node).`
- **Item 7 (shellcheck)** — `bash claude/scripts/tests/run-shellcheck.sh` → `SKIP: shellcheck not
  found` (documented skip behavior; not installed locally).
- `bash -n claude/scripts/journal-compose-with-retry.sh` → syntax OK.
- Manually verified `date -d yesterday +%Y-%m-%d` resolves correctly against `date +%Y-%m-%d` /
  `date -u +%Y-%m-%d` in this Git Bash environment.
- Suppression grep (`git diff origin/main -- . | grep -E '(ts-ignore|...)'`) → clean.
- Test-integrity grep → clean.
- The `claude/routines/daily-journal-compose/SKILL.md` change is skill-markdown with no dedicated
  automated test, consistent with ADR-082's own precedent for this file ("skill-markdown ...
  change. Verification is a full occurrence-grep ... a manual step-consistency walkthrough, and
  the next real end-of-day compose as the actual integration test"). Grepped the edited file for
  any remaining stale "today" references — only intentional ones remain (explaining "yesterday
  rather than today", "today-guard").

Tests: N/A — no automated test suite covers routine SKILL.md or this specific `.sh` wrapper (no
`test-*.sh` companion exists for `journal-compose-with-retry.sh`, unlike `merge-stale-pr.sh`);
deferred as disproportionate to a 4-line change with no branching logic, per ADR-084's Testing
section.

## Risk-dimension audit

1. **Testing** — see above.
2. **Observability** — N/A, dev-env has no runtime; the retry wrapper's existing `log()` calls
   (unchanged) remain its diagnostic surface.
3. **Security** — N/A, no new credentials/secrets/auth surface.
4. **Resilience** — strictly improves resilience: the automated path now completes instead of
   guard-blocking every run; removes the retry wrapper's dependency on an ambiguous multi-branch
   scan a non-interactive session can't resolve interactively.
5. **Performance** — negligible; one `date` invocation changes.
6. **Data integrity** — N/A, no schema/migration surface.

- **[reliability] Overnight-session edge case newly activated by this fix** — fixed in `9fb5094` (documented in ADR-084's Consequences section rather than a code change; the failure mode is recoverable via the existing `-recovery` runbook, not data loss).
- **Question: does `journal-compose-with-retry.sh`'s 'Create a stub for today' mean the execution day, not the compose target?** — resolved in `9fb5094` (confirmed yes; added a one-line clarifying comment).


brownm09:
## Review: [fix] Nightly journal-compose targets yesterday's local date, not today's UTC

### Summary
Ready to merge with minor follow-up. The fix is correct and narrowly scoped: both automated entry points now compute yesterday's local calendar date instead of today's UTC date, which restores ADR-017's originally-documented intent and requires no change to `/journal-compose` itself. Verified the `date -d yesterday` computation live in this environment, confirmed the doc-reconciliation checkpoint (README.md/docs/REFERENCE.md) needs no changes since both already describe only the still-accurate higher-level behavior, and confirmed the new ADR-084 is properly indexed. One non-blocking reliability point worth documenting, and one clarifying question.

---

### Non-Blocking Findings

**[reliability]** `claude/routines/daily-journal-compose/SKILL.md:24`, `claude/scripts/journal-compose-with-retry.sh:17`
Before this fix, the automated compose never actually ran (guard-blocked every time), so a latent risk around overnight sessions was purely theoretical. After this fix, the ~7am-local run will actually execute `/journal-compose <yesterday>` for real — which activates a narrow but genuine edge case: a session that starts before local midnight and is still uncommitted past ~7am the next morning writes its stub under *yesterday's* date, but if the automated compose runs (and merges) before that session pushes, the session's later push to the now-merged `draft/YYYY-MM-DD` branch will hit the pre-push hook's merged-draft-branch block — and, critically, the hook's same-day exception (ADR-017 §2) won't save it, because by the time the push happens it's already the *next* local day relative to the branch's own date. This isn't silent data loss (the push fails with a clear error, and the existing `draft/YYYY-MM-DD-recovery` runbook already covers recovering a merged branch that needs more legitimate commits) — but it's a new-in-practice failure mode this PR should at least flag for the record, since nothing currently checks for an in-progress session before the automated compose proceeds.
**Suggestion:** No action required for this PR (the existing recovery runbook covers it, and this is a narrow edge case for this user's typical session timing). Worth a one-line mention in ADR-084's Consequences section, or a follow-up issue, if a future hardening pass wants to add a liveness check (mirroring `_worktree_liveness.py`'s `worktree_session_is_live()` pattern) before the automated compose proceeds.

---

### Questions for Author

**Question:** Is `journal-compose-with-retry.sh`'s unchanged `"Create a stub for today."` phrase intentionally referring to the *execution day* (when this maintenance session itself runs), not the compose *target* date (`${DATE}`, yesterday) — i.e., deliberately left as a different, correct concept rather than something the DATE fix should have touched?
**Context:** The diff adds `${DATE}` (yesterday) into the `/journal-compose` invocation on the same line, but leaves `"Create a stub for today"` untouched two sentences later in the same string (`claude/scripts/journal-compose-with-retry.sh:19`). Read cold, the juxtaposition of "yesterday" and "today" in one prompt string reads as a possible leftover inconsistency rather than a deliberate distinction.
**Tradeoffs:** (a) Leave as-is (current PR) — correct if "today" genuinely means "the day this retry session executes," which is a different, valid concept from the compose target; no functional impact either way since this phrase isn't consumed by `/journal-compose`'s own date-resolution. (b) Add a one-line comment above the `PROMPT=` assignment clarifying the distinction — costs nothing and removes the ambiguity for the next reader, without changing behavior.

---

### Findings Disposition
_1 non-blocking finding above needs a disposition (fixed in `<sha>` | filed `#N`) before merge, per ADR-028/ADR-039. The Question is advisory only and doesn't require a disposition entry._

---

PR #580, opened 2026-07-05
feat(hooks): detect Bash repo/branch drift at commit/PR-create/PR-merge checkpoints

### LINKED ISSUE #573: [bug] EnterWorktree cwd silently reverts after git checkout in a different repo via -C, causing gh pr create to target the wrong branch
## Problem

A session working in a git worktree (via `EnterWorktree`) had its Bash tool's tracked working
directory silently revert to the main canonical checkout — with no error, warning, or indication
— after running `git -C <other-repo> checkout` / `checkout -b` commands targeting a **completely
different repository** via `-C`.

## Concrete incident

Session was in worktree `C:\...\lifting-logbook\.claude\worktrees\issue-671-version-endpoints`
(branch `feat/issue-671-version-endpoints`). Sequence:

1. `git add` / `git commit` / `git push -u origin feat/issue-671-version-endpoints` — all
   succeeded correctly, confirming cwd was the worktree at this point (the push explicitly named
   and pushed the worktree-only local branch).
2. `gh pr create --title ... --body ...` (no explicit `--head`) — succeeded, returned a PR URL.
3. Between steps 1 and 2, **no other Bash tool call was made**.
4. Several turns later, `gh pr view` on that PR showed `headRefName` as an entirely different,
   stale branch (`fix/issue-660-worker-cap-node-gate` — the session's *original* branch from
   before `EnterWorktree` was ever called), with `additions`/`deletions`/`changedFiles` all 0.
   `gh pr view --json commits` confirmed the PR genuinely contained none of the intended work —
   just old, already-merged commits from the stale branch.
5. Re-checking `pwd`/`git branch --show-current` in the same session confirmed the Bash cwd had
   indeed reverted to the main canonical checkout, on the stale branch — with no error ever
   surfaced for this.

**Suspected trigger:** between steps 1 and 2 in the surrounding session (not directly between
those two calls, but earlier in the same stretch), the session had run a `journal-compose`-style
workflow against a **second, unrelated repo** (`engineering-journal`) using
`git -C C:/Users/brown/Git/engineering-journal checkout main`, `checkout -b draft/YYYY-MM-DD`,
etc. `-C` should scope those commands entirely to the other repo and have zero effect on the
Bash tool's cwd or the harness's worktree-tracking state for the *original* project. The
suspicion is that some worktree-tracking heuristic pattern-matches on the string `git checkout`
appearing anywhere in a Bash tool call (regardless of `-C` redirecting to a different repo
entirely) and resets/loses track of "which worktree is the session currently in" as a result.

## Impact

This is silent and has no error signal — the only tell was noticing wrong metadata on a
subsequently-created PR, well after the fact. In this incident it caused a PR to be opened
against the wrong branch, requiring it to be closed and recreated. In a worse case (e.g., if the
session had gone on to `git commit`/`git push` more work believing it was still in the worktree)
it could have committed real work to the wrong branch, or to the main canonical checkout directly
(which the `canonical-mutate-guard` hook would catch for mutating commands, but does not catch
retroactively for a `gh pr create` that merely *reads* the current branch to use as an implicit
default).

## Suggested investigation

1. Reproduce: `EnterWorktree` into a project worktree, do some work, then run a `git -C
   <other-repo> checkout -b <branch>` against a **second, unrelated repo**, then check whether
   the session's Bash cwd / harness-tracked "current worktree" is still correctly the first
   worktree.
2. If reproduced, the fix is presumably to scope whatever heuristic detects "did the session
   switch worktrees" to the **repo path being operated on**, not just the presence of a
   `checkout`/`checkout -b` substring in the command — a `-C <path>` (or equivalent `cwd`
   override) targeting a different repo should never affect tracking for the original one.
3. Consider whether `EnterWorktree`'s docstring already documents this class of issue ("After a
   further switch, previously-visited worktrees are no longer writable") — if so, the gap may be
   that switching into a **different repo entirely** (not another worktree of the *same* repo)
   isn't being excluded from that "switch" detection the way it should be.

## Suggested mitigation in the meantime (documented as a journal note, not yet in CLAUDE.md)

Until this is root-caused, sessions using the engineering-journal stub workflow mid-task in a
project worktree should re-issue `EnterWorktree path=<original worktree>` immediately after any
`git -C <other-repo> checkout ...` command, before trusting Bash cwd for further work in the
original project — and consider always passing `--head <branch>` explicitly to `gh pr create`
rather than relying on implicit current-branch detection, as a cheap defense-in-depth measure
regardless of root cause.

## Discovered during

lifting-logbook session, 2026-07-05, while opening a PR for issue #671 (deployment version
visibility). See `sessions/lifting-logbook/2026-07-05_132313.stub.md` in engineering-journal for
full narrative.

brownm09:
## Summary

Investigates dev-env#573: a session's Bash cwd (and separately, a checked-out branch) silently
reverted to a stale/default state with no error surfaced, once causing `gh pr create` to open a
PR against the wrong branch entirely.

**Root cause is not fixable from this repo.** Investigation findings (posted in full to #573):
- The refined disk-pressure theory is *plausible but not confirmed* — MSYS2's own FAQ documents
  the "unable to remap" stackdump class as an ASLR/DLL-conflict issue, not disk/memory
  exhaustion. There is an **open, unfixed upstream Claude Code issue for the exact same symptom**
  ([anthropics/claude-code#37920](https://github.com/anthropics/claude-code/issues/37920)),
  auto-closed by a stale-bot with no fix landed — one reporter saw 6 identical stackdumps with no
  resource pressure at all, suggesting a baseline rate on Windows regardless of cause.
- The harness's own cwd-tracking has a **documented reliability gap independent of the crash
  theory** ([anthropics/claude-code#11067](https://github.com/anthropics/claude-code/issues/11067)):
  its "Shell cwd was reset to `<dir>`" self-correction message can repeat without actually fixing
  subsequent command execution directory.
- A full audit of dev-env's hook architecture confirmed **no existing mechanism** tracks
  shell/process continuity across Bash calls, and `EnterWorktree`/`ExitWorktree` are not
  hook-integrated at all.

**What's in reach: make the symptom loudly visible where it caused real damage.** New shared
module `_bash_state.py` records the repo root + branch seen after every Bash call
(`post-tool-use-cwd-track.py`); `pre-commit-branch-check.py`, `pre-pr-create-check.py`, and the
new `pre-merge-branch-check.py` compare that record against the current state at `git commit` /
`gh pr create` / `gh pr merge` and surface a loud, **advisory-only** warning on a mismatch — never
blocking, confirmed with the user directly, since the mechanism can't structurally distinguish a
legitimate `EnterWorktree`/`cd` switch from a silent revert. Full design rationale in
[ADR-085](docs/adr/085-bash-repo-branch-drift-detection.md).

This does **not** fix the underlying crash and cannot catch a revert that never reaches one of
those three checkpoints — documented explicitly so it's never mistaken for a cure.

## Testing

Tests: 706 passed, 0 skipped, 0 failed (full existing suite + 4 new files, ~15s)

- `py -3 -c "import ast..."` syntax check over all `claude/scripts/*.py` — clean.
- Every test file in `claude/scripts/tests/test_*.py` (37 files, including the 4 new/modified
  ones for this change) — 100% pass, nothing broken by this change.
- Manual end-to-end dry run reproducing the incident's exact shape: recorded worktree state, then
  a simulated `git commit`/`gh pr create`/`gh pr merge` from the canonical root on `main` — all
  three correctly fire the drift warning naming both states; the no-drift case (same repo/branch)
  is confirmed byte-identical to the pre-existing output for `pre-commit-branch-check.py`.
- Suppression grep, test-integrity grep, deleted-test-file check — all clean.
- `docs-only guard` (`grep -n 'date -u' claude/CLAUDE.md`) — no matches.
- Not run: shellcheck / `get-project-item.sh` smoke test / pre-push self-test / script
  path-hygiene lint — N/A, no shell scripts touched by this PR.

Caught and fixed one real numbering collision before it could block at merge time: `origin/main`
picked up a concurrent PR claiming ADR-083 while this branch was in flight (`git fetch` +
`git diff <merge-base> origin/main` surfaced it) — renumbered this PR's ADR to 084 throughout.

All findings fixed in commit `2df8b4b`:
- **[correctness] Blocking** — spurious self-contradictory drift warning on a transient git-read failure. Fixed: early-return in `format_drift_warning` when both current values are `None`.
- **[reliability] Non-blocking** — no cleanup/expiration for the per-session state file. Fixed: added `cleanup_stale_state()` mirroring `_hookutil.py`'s `MAX_AGE_DAYS` pattern.
- **[performance] Non-blocking** — two subprocess spawns per Bash call in the every-call hook. Fixed: combined into one `git rev-parse --show-toplevel --abbrev-ref HEAD` call.
- **[maintainability] Non-blocking** — `current_branch()`/`current_repo_root()` duplicated across three files, with the risk already having caused one bug during development. Fixed: extracted into shared `_bash_state.current_repo_state()`.

ADR-085 updated to reflect the corrected design.

Closes #573

---
**Update:** caught a second, live numbering collision after opening this PR — PR #579 (unrelated, still open) already claims ADR-084 on its own branch. Renumbered this PR's ADR to 085 in a follow-up commit before either PR merged.


brownm09:
## Review: feat(hooks): detect Bash repo/branch drift at commit/PR-create/PR-merge checkpoints

### Summary

Solid, well-scoped mitigation with clean security posture and no missing test coverage — but one genuine correctness bug in the core comparison function needs a fix before merge: `format_drift_warning` can fire a self-contradictory warning (`<unknown> @ <unknown>` shown against an unchanged cwd) when a `git` call transiently fails or times out at the checkpoint, which undermines the exact "don't cry wolf" precision the design is built around. Two non-blocking findings are worth fixing in the same pass (unbounded per-session state-file accumulation with no cleanup, unlike every sibling per-session file in this codebase; and a maintainability risk in the `current_branch()`/`current_repo_root()` duplication across three files that has already caused one bug during this PR's own development). At 1311 changed lines across 16 files, the PR is appropriately scoped as a single change — it's one coherent mechanism (shared state module + one new hook + three checkpoint hooks + tests + docs + ADR), and splitting it would fragment a single design decision without reducing review risk.

---

### Blocking Findings

**[correctness]** `claude/scripts/_bash_state.py:231` (`format_drift_warning`)
When a `git` call at the checkpoint (inside `current_repo_root`/`current_branch` in any of the three consuming hooks) transiently fails or times out, both current values become `None`. Since the recorded state holds a real `(repo_root, branch)`, the tuple comparison treats `(real, real) != (None, None)` as drift and fires — producing a warning that reads `now: <unknown> @ <unknown> (cwd: <same path as "was">)`. The cwd is unchanged on both lines, which makes the warning self-contradictory and undermines trust in the mechanism exactly where the ADR's design intent was to avoid noisy false positives.
**Fix:** Early-return `None` from `format_drift_warning` when both `current_repo_root` and `current_branch` are `None` — "current git state unknown" can't be meaningfully compared, and genuine drift (worktree→canonical, branch-only reversion) always leaves at least one current value populated. Add a test case for this.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/_bash_state.py`, `claude/scripts/post-tool-use-cwd-track.py`
The per-session state file (`~/.claude/scratch/bash_state_<session_id>.json`) has no cleanup or expiration — every distinct session id leaves a file behind permanently. This is the only per-session-file producer in the codebase without one: `_hookutil.py`'s `MAX_AGE_DAYS = 30` + `cleanup_stale_sentinels()` already backs `posttooluse-inert-advisory.py` and `reconcile-open-prs.py`, and `turn-count-hook.py` implements the same cutoff independently. Notable given this specific fix's own context is a disk-pressure-adjacent incident.
**Suggestion:** Add a small cleanup routine to `_bash_state.py` mirroring `_hookutil`'s pattern (or reuse it directly), called opportunistically from `post-tool-use-cwd-track.py`.

**[performance]** `claude/scripts/post-tool-use-cwd-track.py:284-311`
Two sequential `subprocess.run` spawns (`git rev-parse --show-toplevel`, then `git branch --show-current`) run on every single Bash call across every session — this is the one place in the hook family paying that cost unconditionally rather than only at a checkpoint.
**Suggestion:** Combine into one call: `git rev-parse --show-toplevel --abbrev-ref HEAD` returns both values on two lines. Map a detached-HEAD `HEAD` second line back to `None` to stay consistent with the three readers' `git branch --show-current` convention.

**[maintainability]** `claude/scripts/pre-commit-branch-check.py:47-72`, `pre-pr-create-check.py:138-166`, `pre-merge-branch-check.py:60-88`
`current_branch()`/`current_repo_root()` are duplicated near-identically across three files. The ADR defends this by analogy to an existing single-function duplication elsewhere in the codebase, but this PR's own history already demonstrates the coupling risk concretely: `pre-commit-branch-check.py`'s `current_branch()` originally returned a display placeholder on failure instead of `None`, which would have manufactured a spurious drift warning on every detached-HEAD commit — caught and fixed during development, not hypothetically.
**Suggestion:** Given the risk already manifested once, extract the two functions into `_bash_state.py` (already the shared import for all three files) so the `None` convention is enforced by one implementation instead of three synchronized copies.

---

### Findings Disposition

Per ADR-028, each finding above needs a disposition (fixed in `<sha>` | filed `#N`) in the PR body before merge. Proceeding to fix all three in this same session.

---

Issue #346, opened 2026-06-09
biweekly-retro: do report git work in a dedicated worktree, not the shared canonical EJ tree


brownm09:
## Context

Follow-up from the `/review` of [#345](https://github.com/brownm09/dev-env/pull/345) (adds the `biweekly-retro` routine).

## Finding (non-blocking, fails safe today)

`biweekly-retro` Step 5 does `cd $EJ; git checkout -b retro/${RUN_DATE} origin/main` directly in the **shared canonical** engineering-journal working tree (`C:/Users/brown/Git/engineering-journal`). That tree is frequently parked on a `draft/YYYY-MM-DD` branch with uncommitted stub edits from interactive journal sessions.

Today this **fails safe**: Step 0's `sync-routine-worktree` rebases a non-main branch onto `origin/main` and ABORTs (clean exit + push notification) if the tree is dirty, so the routine simply skips that cycle rather than corrupting anything. The downside is that the retro may silently no-op on any Sunday the canonical tree is left mid-draft.

## Suggested fix

Do the report commit in a dedicated throwaway worktree cut from `origin/main`, never touching the shared tree's branch state:

```bash
RETRO_WT="$EJ/.claude/worktrees/retro-${RUN_DATE}"
git -C "$EJ" worktree add "$RETRO_WT" -b "retro/${RUN_DATE}" origin/main
# write report + README under $RETRO_WT/sessions/meta/retro/
git -C "$RETRO_WT" add sessions/meta/retro/ && git -C "$RETRO_WT" commit -m "..." && git -C "$RETRO_WT" push -u origin "retro/${RUN_DATE}"
git -C "$EJ" worktree remove "$RETRO_WT"
```

When applied, update BOTH the dev-env source (`claude/routines/biweekly-retro/SKILL.md`) AND re-register the live scheduled task via the `scheduled-tasks` MCP (separate copies — see #344).

Found while implementing #343.

brownm09:
Scope clarification, cross-referenced from #747: this issue's fix is specifically about
`biweekly-retro` Step 5's report-writing operation (`git checkout -b retro/${RUN_DATE}` in the shared
canonical). It does not cover, and closing it will not fix, the daily Stub file workflow's own use of
the canonical directly for `draft/YYYY-MM-DD` — that exemption is intentional, permanent
infrastructure, not a stopgap pending this issue (`claude/CLAUDE.md` → Engineering Journal → Stub file
workflow).

[...]

PR #584, opened 2026-07-05
[fix] canonical-mutate-guard: resolve -C/--git-dir/--work-tree redirect target, not just cwd

### LINKED ISSUE #576: canonical-mutate-guard: -C/--git-dir/--work-tree redirect target and worktree cwd let a mutation of a different canonical checkout through
## Problem

`pre-tool-use-canonical-mutate-guard.py` (ADR-071) is **cwd-centric**: it blocks a mutating git verb only when *cwd itself* resolves to a canonical (non-worktree) checkout. A command that mutates a **different** canonical checkout via a `-C <path>` / `--git-dir=<path>` / `--work-tree=<path>` redirect is not caught — the guard never resolves the redirect target and applies the same canonical-root check to it. `git -C <path>` is exactly equivalent to running the command with that cwd, so it should be treated the same.

## Incident (2026-07-05, win11-init-tools session)

During engineering-journal bookkeeping from a win11-init-tools **worktree** (cwd was the worktree, not the journal repo), `git -C C:/Users/brown/Git/engineering-journal pull` ran **un-blocked** while another concurrent session was actively mutating that same shared canonical journal checkout: its checked-out branch flipped `draft/2026-07-03` → `draft/2026-07-02` between two read-only checks seconds apart, with a large uncommitted stub/manifest set across career-playbook, dev-env, and lifting-logbook. No damage — git's own "local changes would be overwritten" safety aborted the pull, and I isolated into a fresh worktree to finish — but the guard that exists to prevent exactly this collision never fired.

## Confirmed mechanisms (read from the deciding code, not the docstring)

Two independent reasons the command was allowed:

1. **Worktree short-circuit** — `main()` runs `if _WORKTREE_RE.search(cwd): sys.exit(0)` as its first substantive check, so a worktree cwd exits 0 **before the command is ever parsed**. This is what fired in the incident. Pinned by `test_main_allows_any_command_from_worktree_cwd`.
2. **`_REDIRECT_RE` segment-skip** — even from a non-worktree cwd, `classify()` matches `git -C <path>` / `git --git-dir=<path>` and `continue`s, deliberately skipping the segment. Pinned by `test_dashC_redirect_only_skips_its_own_segment`, which literally asserts `git -C C:/Users/brown/Git/dev-env checkout -b foo` is skipped and returns `None`.

Both the module docstring and ADR-071 document this as a **deliberate v1 deferral** — *"a command that `cd`s or `-C`s into the canonical root from a worktree's Bash session is not caught… extendable later if it recurs in practice."* This incident is that recurrence.

**Adjacent minor gap:** `--work-tree=` isn't even in `_REDIRECT_RE`, and `git --work-tree=<path> commit` misclassifies as non-mutating (the verb detector sees the flag, not `commit`).

## Proposed fix — make the guard target-aware

Resolve `-C` / `--git-dir=` / `--work-tree=` targets and apply the canonical-root check to the **resolved target**, not just cwd:

1. Capture the redirect target from a mutating segment's first line (preserve the `_first_line()` anchoring so a heredoc/`$()` body mention can't spoof it).
2. Teach the git-global-flag walker to skip **and** capture `-C <v>` / `--git-dir[=| ]<v>` / `--work-tree[=| ]<v>` so the real verb is classified (also fixes the `--work-tree=` misclassification).
3. Resolve each target via `git -C <dir> rev-parse --show-toplevel`; if it's a canonical (non-worktree) root, block (exit 2, `ALLOW_CANONICAL_MUTATE=1` override honored).
4. Narrow the worktree short-circuit surgically: a worktree cwd with **no** redirect stays a zero-cost exit 0; a worktree cwd **with** a redirect falls through to the target check.

## Design tension — the documented journal `git -C` workflow

The engineering-journal stub workflow (global CLAUDE.md + ADR-066) **automatically** runs `git -C <journal-canonical> checkout/commit/pull` on every PR open/merge. A naive block would force `ALLOW_CANONICAL_MUTATE=1` on that automated path — untenable (trains reflexive override use; needs a large CLAUDE.md doc rewrite). Chosen resolution: a narrow, documented **carve-out** exempting the engineering-journal canonical checkout, temporary pending the journal→worktree migration (#346). Consequence to accept: the incident's *exact* command stays allowed by the carve-out; the general gap closes for every other repo (the real #453 collision surface).

## References

- ADR-071 (this hook) — the v1-deferral note this issue lifts
- #453 — the guard's motivating collisions
- #558 — the closed `gh pr merge -d` Amendment 1 (same "extend coverage" shape)
- #573 — distinct harness cwd-tracking bug, same `git -C <other-repo> checkout`-from-worktree shape; this fix incidentally blocks the mutating command that triggers its cwd-revert
- #346 — move journal git work off the shared canonical EJ tree (the journal path's real fix; carve-out removal tracked here)

Behavior change → warrants ADR-071 Amendment 2 + the 4 sync-doc locations (module docstring, global CLAUDE.md bullet, ADR-071, docs/REFERENCE.md).

brownm09:
## Summary

Makes `pre-tool-use-canonical-mutate-guard.py` **target-aware**. It was cwd-centric: it blocked a mutating git verb only when *cwd itself* resolved to a canonical (non-worktree) checkout. A command mutating a *different* canonical checkout via a `-C`/`--git-dir`/`--work-tree` redirect slipped through — the exact gap the 2026-07-05 incident hit (`git -C <engineering-journal> pull` from a project worktree ran un-blocked while another session was actively mutating that shared canonical checkout).

Closes #576.

## What changed

- **Hook (`pre-tool-use-canonical-mutate-guard.py`)** — `_parse_git_prefix` now consumes `-C`/`--git-dir`/`--work-tree` (both `=` and space forms) as git-level options, exposing the real verb *and* capturing the target dir. This simultaneously fixes the pre-existing `git --work-tree=<path> commit` misclassification (the flag was mistaken for the verb). `find_mutating_segments` stays pure/offline (string work only); `main()` resolves each redirect target and blocks when it lands on a canonical (non-worktree) root — regardless of cwd. The worktree short-circuit is narrowed: an *ambient* command in a worktree still exits 0 cheaply; a redirect-into-canonical from a worktree is now evaluated.
- **Journal carve-out** — an exact-path-matched allowlist (see Review findings disposition below for why it's exact-path, not basename) exempts the journal canonical checkout, overridable via `CANONICAL_MUTATE_GUARD_JOURNAL_PATH` for tests. The documented stub workflow runs `git -C <journal-canonical> checkout/commit/pull` automatically on every PR open/merge; blocking it would force `ALLOW_CANONICAL_MUTATE=1` onto an automated path. Temporary, pending the #346 journal→worktree migration.
- **Docs synced** (doc-reconciliation): module docstring, global `claude/CLAUDE.md` bullet, ADR-071 **Amendment 2**, `docs/REFERENCE.md` (hooks-table cell + runbook), `README.md`, `docs/adr/INDEX.md`, and dev-env `CLAUDE.md` item 33.

## Design decision — reconciled with ADR-082

ADR-082 recorded "Extend ADR-071's guard to parse into `git -C` targets" as a **rejected alternative**, on two grounds: (a) `-C` is "deliberate, visible authorship", and (b) it would block compose's legitimate journal cross-repo ops. This PR reverses that, and addresses both explicitly in ADR-071 Amendment 2:
- **(a)** The incident is new evidence the "deliberate authorship" distinction is incomplete — the author deliberately typed `git -C <journal> pull`, but the *collision* with a concurrent session was still silent and unintended. This is the "extend if it recurs in practice" trigger ADR-071's v1 deferral named.
- **(b)** The carve-out surgically preserves exactly those legitimate journal ops.

ADR-082's rejected-alternative entry is updated in this PR to forward-reference Amendment 2, so the two ADRs don't contradict.

**Consequence stated plainly:** the incident's *exact* command (`git -C <journal> pull`) stays allowed by the carve-out; the general gap closes for every *other* repo (the real #453 collision surface). The journal's real fix is worktree isolation (#346), which removes the carve-out.

## Testing

Ran from the worktree root:

- `py -3 claude/scripts/tests/test_canonical_mutate_guard.py` → **Tests: 64 passed, 0 skipped, 0 failed** (48 at branch start → 58 after the initial implementation → 64 after addressing all 5 `/review` findings below).
- `py -3 claude/scripts/tests/test_worktree_path_check.py` → 5 passed (sibling-hook regression).
- `py -3 claude/scripts/tests/test_no_crude_command_substring_checks.py` → 13 passed (the new parsing is tokenized via `_parse_git_prefix`/`_tokenize`, not a crude `"<literal>" in command` substring check).
- `py -3 -c "import ast … " claude/scripts/*.py` → all scripts parse.

New coverage: `_parse_git_prefix` capture/verb-exposure; `--work-tree`/`--git-dir`/`-C` mutating classification; `_segment_redirect_dirs` first-line-only anchoring; `_is_allowlisted_root` exact-path carve-out (plus a same-basename-wrong-path negative case); `_tokenize`'s quoted-path capture and unbalanced-quote fallback; `_resolve_git_toplevel`'s null-byte fail-open; and end-to-end — `-C <canonical>` from a worktree blocked (target named), `-C <journal>` allowed (carve-out), a same-basename-different-path repo correctly NOT exempt, `-C <worktree>` allowed, `-C <other-canonical>` from a canonical cwd blocked, `--work-tree=<canonical> commit` blocked, override bypasses a redirect block, a quoted space-bearing `-C` target blocked, and a relative `-C` target resolved against the command's own cwd and blocked.

## Gate checks

- **Coverage gate:** new behavior fully covered (above).
- **Test integrity:** one existing test intentionally changed — `test_dashC_redirect_only_skips_its_own_segment` (which asserted `classify()` returned `None` for `git -C <canonical> checkout`, i.e. it asserted *the gap*) is repurposed to `test_dashC_redirect_captured_and_classified` with an inverted assertion matching the new behavior. This is a behavior-driven update, not a degradation — total test count rose 48→64, none skipped/deleted. Integrity grep + deleted-test-file check clean.
- **Suppression grep:** clean (Python change; no TS suppressions).
- **Lockfile:** N/A — no `package.json` change.
- **ADR-warrant:** ADR-071 Amendment 2 written (plus a "Review hardening" subsection after `/review`); INDEX date/tags updated.

`/review` ([comment](https://github.com/brownm09/dev-env/pull/584#issuecomment-4887916855)) found 3 blocking + 2 non-blocking findings via two parallel Opus subagents (correctness/security; reliability/performance/maintainability), each independently re-verified end-to-end before being accepted. All 5 fixed in-PR in commit `0427886` (no findings filed out-of-scope):

1. **[correctness, blocking]** Quoted `-C`/`--git-dir`/`--work-tree` targets containing a space (common on Windows) shattered under a naive `.split()` tokenizer, silently defeating the block — **fixed** via a new `_tokenize()` helper (`shlex.split(posix=True)` with a fallback to plain `.split()` on `ValueError`, required for the existing heredoc-truncation cases to keep passing).
2. **[correctness, blocking]** A relative redirect target resolved against the hook script's own unrelated process cwd instead of the command's actual cwd, silently missing the exact collision this PR exists to prevent — **fixed** by joining a non-absolute target onto the payload `cwd` before resolution.
3. **[reliability, blocking]** A null byte in a redirect value crashed the hook (uncaught `ValueError` from `subprocess.run`) instead of failing open — newly exposed because redirect dirs are now command-string-derived, not just the harness-provided cwd — **fixed** by adding `ValueError` to `_resolve_git_toplevel`'s except tuple.
4. **[correctness, non-blocking]** The journal carve-out matched on basename alone, exempting any canonical checkout anywhere named "engineering-journal" — **fixed**: hardened to an exact path match, overridable via `CANONICAL_MUTATE_GUARD_JOURNAL_PATH` so tests never touch the real journal checkout.
5. **[performance, non-blocking]** No memoization across redirect resolutions let a crafted command spawn up to 15 subprocesses — **fixed** with a dict-based memo shared across `main()`'s per-segment loop (everyday commands unaffected, still 0–1 spawns).

## Follow-up

- **#346** (journal git work → worktree) is the path to removing the carve-out. A post-merge tile will capture this.
- **#573** (distinct harness cwd-tracking bug, same `git -C <other-repo> checkout`-from-worktree shape) is incidentally mitigated: the guard now blocks the mutating cross-repo command that triggers its cwd revert. #573 still needs its own harness-level fix.


brownm09:
## Review: [fix] canonical-mutate-guard: resolve -C/--git-dir/--work-tree redirect target, not just cwd

### Summary
**Needs revision before merge.** Three confirmed, independently-verified correctness/reliability bugs undermine the PR's own stated purpose: quoted redirect paths with embedded spaces silently bypass the new block, relative redirect paths resolve against the wrong working directory (also bypassing the block), and a null-byte-bearing redirect path crashes the hook outright rather than failing open. The highest-severity finding is the quoted-path bypass, since Windows paths routinely contain spaces and this hook explicitly targets Windows. DIFF_SIZE (977 lines) is appropriately scoped despite its size — it's one cohesive change (hook logic + its own test suite + the doc-sync this repo's own Documentation Maintenance policy requires in the same PR); splitting it would violate that policy, not reduce risk.

Documentation reconciliation (Step 2b), documentation coverage (Step 2c), and the test coverage gate (Step 2d) all pass clean — the hooks table, `docs/REFERENCE.md`, and ADR-071 Amendment 2 are already updated in this PR, no ancestor directory has a README that needs touching, and new behavior is fully covered by tests. For Step 2e's language-scope preamble: this diff is Python-only, so the JS/TS skip-marker scanners don't apply — I manually read the full diff instead and confirmed no pytest-style skip markers, deletions, or threshold changes exist.

---

### Blocking Findings

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:315` (`_git_rest_tokens`)
Tokenizes a segment's first line with plain `rest.split()` — a whitespace split with no quote-awareness. A redirect value containing a space (e.g. `git -C "C:/Program Files/some-canonical" checkout -b x`, or any Windows path under a directory with a space in its name) shatters into multiple bogus tokens, so `_parse_git_prefix` captures a mangled path and the verb never lands where the classifier expects — the invocation is not recognized as mutating at all.
I verified this end-to-end: from a worktree cwd, `git -C "<canonical path with a space>" checkout -b x` exits **0** (allowed) where the PR's own stated goal requires exit 2. This is not a new regression (the pre-PR code also let `-C` redirects through, just via a different mechanism) — but it means the fix does not actually close the gap for any space-bearing path, which is common on the exact platform (Windows) this hook targets.
**Fix:** Tokenize with a quote-aware splitter (e.g. `shlex.split(rest, posix=True)`) with a fallback to the current plain `.split()` on a `ValueError` (an unbalanced quote can legitimately occur when `_first_line()` truncates a multi-line heredoc segment — the existing `test_commit_with_heredoc_command_sub_message_still_classified` case relies on exactly that fallback path staying intact). Add an end-to-end test asserting a quoted, space-bearing `-C` target is blocked.

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:606-609` (`_blockable_redirect_root` → `_resolve_git_toplevel`)
A relative redirect value (e.g. `git -C ../other-repo checkout`, or `--git-dir=.git` after normalization) is passed straight to `git -C <value> rev-parse --show-toplevel`. Since `subprocess.run` doesn't override `cwd=`, that `git` subprocess inherits the *hook script's own* process working directory — which has no relationship to the Bash command's actual `cwd` (the JSON payload field). The *ambient* path correctly resolves against payload `cwd`; the *redirect* path silently doesn't.
I verified this end-to-end: from a worktree at `.../worktrees/wt2`, `git -C ../../canonical-rel-target checkout -b x` (where `../../canonical-rel-target` is a real canonical repo relative to `wt2`) exits **0** — the exact collision this PR exists to prevent, missed because the relative path was never resolved against the command's own cwd.
**Fix:** Before resolving, join a non-absolute redirect value onto the payload `cwd` (`os.path.isabs(d)` guard + `os.path.join(cwd, d)` otherwise) and thread `cwd` into `_blockable_redirect_root`. `git rev-parse --show-toplevel` will correctly canonicalize any `..` segments in the joined path. Add an end-to-end test with a relative `-C` target resolving to a canonical sibling of the worktree.

**[reliability]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:565-581` (`_resolve_git_toplevel`)
The `except (FileNotFoundError, subprocess.TimeoutExpired, OSError)` tuple doesn't include `ValueError`. A redirect value containing a null byte makes `subprocess.run`'s `Popen` raise `ValueError: embedded null character`, which propagates uncaught out of `main()` — I reproduced this directly (`git -C /tmp/a\x00b checkout -b x` from a non-worktree cwd crashes with exit code 1 and a full traceback, rather than exiting 0). This function's whole contract, exercised everywhere else in the module, is "never propagate an exception — return `None` and let the caller fail open." That contract was safe when the only input was harness-provided `cwd`; this PR newly routes a *command-string-derived* value into the same function, and command strings are far less constrained than a harness-provided cwd.
**Fix:** Add `ValueError` to the `except` tuple (one line). Add a pure-function test mirroring the existing `test_resolve_git_toplevel_failsopen_on_timeout_and_oserror` pattern, asserting `_resolve_git_toplevel("<null-byte-path>")` returns `None` rather than raising.

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:513-517` (`_is_allowlisted_root`)
The temporary journal carve-out matches on resolved-toplevel *basename* only, so a canonical checkout at **any** path whose last segment happens to be named `engineering-journal` is exempted — not just the one intended shared checkout. Low real-world likelihood on a single-user machine, but the carve-out is a guard-scope boundary and this widens it further than the one case it's meant to cover.
**Suggestion:** This repo's own global `CLAUDE.md` already hardcodes the journal's absolute path (`C:/Users/brown/Git/engineering-journal`) elsewhere; matching the carve-out against that exact (separator- and case-normalized) path instead of a bare basename closes the extra surface at no added cost — no new subprocess, no portability loss beyond what the codebase already accepts for this single-machine tool.

**[performance]** `claude/scripts/pre-tool-use-canonical-mutate-guard.py:597-664` (`_blockable_redirect_root` + the `main()` loop)
Each matching segment's redirect dirs are resolved independently with no memoization, so a command with several `&&`-chained mutating segments, each carrying several redirect flags, can spawn one `git rev-parse` subprocess per redirect-dir occurrence — verified up to 15 spawns for a crafted 5-segment × 3-redirect-dir command, versus exactly 1 on `origin/main`. This is bounded and adversarial-only (every ordinary Bash call — `ls`, `git status`, an ambient commit, `npm test` — still triggers 0 or 1 spawn, unchanged), so it doesn't affect everyday latency.
**Suggestion:** A small dict-based memo keyed by resolved directory string, shared across the `main()` loop, would collapse repeated identical targets to one subprocess each and bound the worst case to the count of *distinct* dirs in the command.

---

### Findings Disposition

All 5 findings are being fixed in this PR — each is small, self-contained, and clearly introduced by this branch, so "file and link" doesn't apply per the repo's fix-everything-always policy. Disposition recorded in the PR body before merge.

---

PR #587, opened 2026-07-05
Add pre-compose liveness guard against a still-active journal-writing session

### LINKED ISSUE #586: Nightly journal-compose can merge a draft branch while a session is still writing to it
## Problem

PR #579 fixed the nightly journal-compose automation so it actually runs (previously it silently guard-blocked every time). That fix activated a previously-theoretical race documented in ADR-084's Consequences section:

A session that starts before local midnight and is still uncommitted past ~7am the next morning writes its engineering-journal stub under **yesterday's** date. The nightly automation (`claude/routines/daily-journal-compose/SKILL.md`, invoked via `claude/scripts/journal-compose-with-retry.sh`) runs `/journal-compose <yesterday>` around 7am and merges that date's `draft/YYYY-MM-DD` branch. If the automated compose merges the branch before the overnight session pushes its stub, the session's later push hits the pre-push hook's merged-draft-branch block.

This is **not data loss** — the existing `draft/YYYY-MM-DD` recovery runbook (docs/REFERENCE.md → Engineering Journal Internals) already covers recovering a merged branch that needs more commits — but nothing currently checks for an in-progress session before the automated compose proceeds, so the failure is silent until the session tries to push.

## Investigation findings

Read the existing liveness-guard pattern (ADR-051 / `claude/scripts/_worktree_liveness.py`) and the journal-compose flow (`claude/skills/journal-compose/SKILL.md` Step 0.6, `claude/routines/daily-journal-compose/SKILL.md`, `claude/scripts/journal-compose-with-retry.sh`):

- **Insertion point**: Step 0.6 of the compose skill already runs a divergence guard (local ref ahead of origin → unpushed stubs exist somewhere) and a compose-worktree concurrency lock check, both before creating the isolated compose worktree (per ADR-082). A new liveness check fits right after the divergence guard and before worktree creation.
- **The divergence guard already catches "committed but not yet pushed."** The actual gap is narrower than "is any session active" — it's specifically "has a session written but not yet *committed* its stub for this date." Because engineering-journal is a single shared checkout (not a per-session worktree — sessions across every project write to it via `git -C C:/Users/brown/Git/engineering-journal ...`), an uncommitted stub would show up as a dirty working tree there.
- **ADR-051's `worktree_session_is_live()` doesn't directly transfer.** It signals liveness for one specific worktree path via transcript mtime. But a session that might write to yesterday's draft branch could be running from *any* project's worktree (lifting-logbook, career-playbook, dev-env, etc.) — there's no single worktree path to check. A transcript-mtime-based check would have to scan system-wide session activity rather than one worktree, a meaningfully broader and noisier signal than ADR-051's original use case.

## Design options (need a decision before implementing)

**A — Git-status check + soft-fail into the existing retry wrapper (recommended).** In Step 0.6, after the divergence guard, run `git -C C:/Users/brown/Git/engineering-journal status --porcelain` scoped to `sessions/*/<date>_*` paths. If dirty, exit non-zero with a clear message. `journal-compose-with-retry.sh` already retries 3x every 5 minutes on non-zero exit — no new retry infrastructure needed. If still dirty after all retries, proceed anyway (accepting the residual risk; the recovery runbook covers it). Cheap, no cross-worktree scanning, reuses existing infra.

**B — Hard block until clean.** Same detection as A, but compose refuses to proceed at all (no retry budget) until the tree is clean for that date. Safest against the race, but an abandoned/never-pushed session would wedge that date's automated compose indefinitely, requiring manual `/journal-compose` later.

**C — System-wide session-liveness check.** Beyond git status, also treat any recent Claude Code session transcript activity (any project, mirroring `worktree_session_is_live()` generalized across all worktrees) as a signal. Most conservative, but highest false-positive rate (most active sessions have nothing to do with the journal) and the most complex to build/maintain.

**D — Defer; rely on the recovery runbook.** Don't build a check now. The race is narrow (ADR-084 calls it "narrow under typical session timing") and already has a working recovery path. Keep this issue open as a backlog item.

## Recommendation

Option A. Smallest change, reuses the existing retry wrapper rather than inventing new blocking behavior, and directly targets the actual gap (uncommitted, not-yet-pushed stub content in the shared checkout) rather than a noisier system-wide liveness signal.

---

Related: ADR-051 (docs/adr/051-worktree-liveness-guard.md), ADR-084 (docs/adr/084-nightly-compose-targets-yesterday.md), ADR-082 (docs/adr/082-journal-compose-worktree-isolation.md).

brownm09:
Closes #586

## Summary

PR #579 fixed nightly `journal-compose` to actually run (previously it silently guard-blocked on every run). That fix activated a previously-theoretical race documented in [ADR-084](docs/adr/084-nightly-compose-targets-yesterday.md)'s Consequences section: a session that starts before local midnight and is still uncommitted past ~7am the next morning writes its stub under *yesterday's* date. If the automated compose merges that date's `draft/YYYY-MM-DD` branch before the session pushes, the push hits the pre-push hook's merged-draft-branch block. Not data loss (the existing recovery runbook covers it), but nothing checked for an in-progress session before compose proceeded.

Adds a two-layer git-status-based liveness check — see [ADR-085](docs/adr/085-journal-compose-liveness-guard.md) for the full design writeup, including why a git-status check on the shared `engineering-journal` checkout was chosen over generalizing ADR-051's `worktree_session_is_live()` (no single worktree path to check — sessions across every project write to the journal via `git -C`), and why the check runs directly in the retry wrapper's bash loop rather than depending on `claude -p`'s exit code reflecting an in-session abort (Claude Code's public docs don't specify this):

- **Primary** — `claude/scripts/journal-compose-with-retry.sh` checks the shared checkout before each retry attempt except the last; a dirty working tree for the target date skips that attempt without spending a `claude -p` call. The final attempt proceeds regardless (bounded risk, existing recovery runbook covers it).
- **Defense-in-depth** — `claude/skills/journal-compose/SKILL.md` Step 0.6 runs the same check for manual/interactive invocations, following the existing divergence-guard's `exit 1`/ABORT convention.
- The shared check script (`check-journal-compose-liveness.py`) stays pure I/O (porcelain text via stdin), matching this repo's established testability convention.

## Testing

Tests: 12 passed, 0 skipped, 0 failed (`test_check_journal_compose_liveness.py`) — grew from 9 after addressing review findings (see disposition below).

Also ran, per dev-env's `## Testing` section:
- Item 1 (hook-script syntax / `ast.parse` over `claude/scripts/*.py`) — OK
- Item 5 (script path-hygiene lint) — clean, 13 files scanned
- Item 7 (shellcheck) — SKIP (not installed locally; documented fallback)
- Suppression grep (global policy Rule 3) — no matches
- Test-integrity grep (global policy Rule 3) — no matches

No `package.json` touched — lockfile policy N/A.

`/review 587` ran two parallel opus subagents (correctness/security; reliability/performance/maintainability). All 5 findings fixed in commit `9c8b5c4`, none filed:

1. **[correctness, blocking]** Missing `set -o pipefail` masked a `git status` failure as "clean" in both check call sites — fixed in `9c8b5c4`.
2. **[reliability]** Malformed/empty `$DATE` would silently degrade the guard — fixed in `9c8b5c4` (date-format validation in `main()`).
3. **[maintainability]** Unsubstituted `"YYYY-MM-DD"` placeholder in the SKILL.md would pass vacuously — fixed in `9c8b5c4` (same validation).
4. **[reliability]** Retry budget shared between liveness-skips and transient-failure retries — documented as an accepted tradeoff in ADR-085's Consequences, fixed in `9c8b5c4`.
5. **[maintainability]** Loose substring date-marker matching, broader than its documented contract — tightened to require the shard suffix, fixed in `9c8b5c4`.

Full review: https://github.com/brownm09/dev-env/pull/587#issuecomment-4889032273


brownm09:
## Review: Add pre-compose liveness guard against a still-active journal-writing session

### Summary
Small, well-scoped, thoroughly documented PR (ADR-085 covers the design rationale in depth). Two independent review passes (correctness/security; reliability/performance/maintainability) converged on the same root bug — a missing `set -o pipefail` that silently masked a `git status` failure as "clean," defeating the guard exactly when concurrent-checkout contention makes it most needed — plus two related silent-degradation paths sharing one fix. All findings below have already been fixed in commit `9c8b5c4`, pushed after this review ran. With those fixes in place, this is ready to merge.

---

### Blocking Findings

**[correctness]** `claude/scripts/journal-compose-with-retry.sh:44` (and the identical pattern at `claude/skills/journal-compose/SKILL.md`'s Step 0.6 liveness check)
A bare `VAR=$(git -C "$EJ" status --porcelain | py -3 check-journal-compose-liveness.py "$DATE" 2>&1)` followed by checking `$?` only reflects the *last* command in the pipe (the Python script), not `git status`. Confirmed empirically: a left-side pipe failure with a clean-exiting right side yields `$?=0`.
If `git status --porcelain` itself fails — missing `$EJ`, corrupt checkout, an index lock held by a concurrent process (itself a signal that contention is highest) — the Python script receives empty stdin, finds no dirty paths, and exits 0. The failure is silently read as "clean" and the wrapper proceeds to compose exactly when a session is most likely to be active, defeating the one guard this PR exists to add. Additionally, without capturing the whole pipeline's stderr, `git`'s own error text was lost rather than logged.
**Fix:** Fixed in `9c8b5c4` — wrapped the pipeline in a subshell with `set -o pipefail` scoped locally (doesn't affect the rest of the script) and `2>&1` applied to the whole subshell (captures both commands' stderr, not just the last one's). Applied identically to both call sites.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/journal-compose-with-retry.sh:22`, `claude/scripts/check-journal-compose-liveness.py`
The guard's correctness depends on `DATE=$(date -d yesterday +%Y-%m-%d)` producing a well-formed string, and the check script did no validation of its date argument. A malformed or empty value would produce a marker (`f"/{date}_"`) that matches no real path, silently degrading the guard to always-clean.
**Fix:** Fixed in `9c8b5c4` — `main()` now rejects any argument that doesn't match `^\d{4}-\d{2}-\d{2}$` with exit 2, converting a silent no-op into a loud, testable error. Covered by a new test (`test_main_rejects_malformed_date`).

**[maintainability]** `claude/skills/journal-compose/SKILL.md` (Step 0.6 liveness block)
The defense-in-depth check passed the literal placeholder string `"YYYY-MM-DD"` as a runtime argument to a script that string-matches on it — unlike other Step 0.6 placeholders that resolve to real shell values, this one is inert (matches nothing, passes vacuously) if the model ever fails to substitute the resolved date before running the command.
**Fix:** Fixed in `9c8b5c4` by the same date-format validation above (the literal `"YYYY-MM-DD"` fails the regex just like any other malformed date) — converts a silent no-op into exit 2. Covered by a new test (`test_main_rejects_unsubstituted_placeholder`). Also documented explicitly in the SKILL.md comment.

**[reliability]** `claude/scripts/journal-compose-with-retry.sh` retry loop / ADR-085
A liveness-triggered skip consumes a retry attempt exactly like a genuine `claude -p` failure does. In the worst case, only the final attempt ever invokes `claude -p` — if that attempt then hits a transient API failure, no retries remain to absorb it, where before this guard existed all 3 tries were available for transient issues alone.
**Fix:** This is an accepted, narrow-intersection trade rather than a bug (already covered by the existing recovery runbook on exhausted retries either way) — documented in `9c8b5c4` as a new Consequences bullet in ADR-085 rather than changed in code.

**[maintainability]** `claude/scripts/check-journal-compose-liveness.py` (`has_uncommitted_target_date_changes`)
The original substring match (`marker in path`) was broader than its documented contract (stub/manifest shards only) — a hypothetical future path merely containing `/YYYY-MM-DD_` anywhere (e.g. a differently-named directory) would register as dirty. Not a live bug given the current `sessions/` layout.
**Fix:** Fixed in `9c8b5c4` — now also requires the path end in `.stub.md` or `.manifest.jsonl`, matching the predicate to the docstring's own grammar. Covered by a new test (`test_wrong_extension_at_matching_date_is_clean`).

---

### Findings Disposition

All 5 findings above were fixed in commit `9c8b5c4` (pushed as a second commit on this branch, after this review ran). None filed as separate issues — all were small, directly-scoped fixes to the exact lines under review. Test suite grew from 9 to 12 passing tests to cover the new validation and tightened-matching behavior.

---

*Reviewed by `claude-opus-4-8` (two parallel subagent passes) via Claude Code*

PR #588, opened 2026-07-05
feat(hooks): implement ADR-083 mechanical pre-check gate for gh pr merge --auto

brownm09:
## Summary

Implements [ADR-083](docs/adr/083-auto-merge-checkpoint-gate.md) — a mechanical pre-check gate that lets `gh pr merge --auto` be used safely. Ships Follow-up items 1–6 (item 7, the separate per-repo `allow_auto_merge` toggle decision, is explicitly out of scope — no repo's auto-merge setting is touched).

- **New hook:** `claude/scripts/pre-auto-merge-checkpoint-gate.py` — only acts on `gh pr merge --auto` (not a plain merge, not an explicit `--auto=false/0/no`). Requires the PR's single most recent comment to carry both a clean-or-disposed `review-findings` marker and a complete, fresh `premerge-checkpoints` marker. **Fails closed** on any gap or `gh`/network error (a deliberate inversion of the sibling gate's fail-open behavior) — no override token.
- **`/review` SKILL.md:** new Step 2f (ADR-Warrant Check, mirrors Step 2b), Step 8's template now emits the `premerge-checkpoints` marker alongside the existing `review-findings` marker in the same comment/code-fence, and Step 6 lists ADR-warrant gaps as a Blocking `[documentation]` finding.
- **Wired** in `settings.json` immediately after `pre-merge-findings-gate.py`.
- **`claude/CLAUDE.md`** and **ADR-031** updated to describe the new gated path (repo-level `allow_auto_merge` still defaults to `false` everywhere — this hook doesn't change that; it only prepares the day a repo opts in).
- **`README.md`** and **`docs/REFERENCE.md`** updated with the new hook.

### Three questions ADR-083 left open, resolved here (see the ADR's new "Implementation" section for full rationale)

1. **Single PR**, not split — the hook and SKILL.md changes define one shared marker contract; splitting risked a half-shipped, inconsistent contract on `main`.
2. **ADR-warrant gaps are Blocking**, category `[documentation]` (not a new category) — matches ADR-011's own "enforcement-style, not advisory" framing and Step 2b/2c's existing precedent.
3. **A third literal marker value, `missing`**, exists for both fields — needed because the marker's regex requires a non-empty value, so a blank field would be indistinguishable from the whole marker being absent (i.e. a pre-ADR-083 review).

### Also closed out

- **Confirmed live** (not just assumed): `gh pr merge 999999 --auto=false --repo brownm09/dev-env` returned a PR-resolution error, not a flag-parsing error — `gh` accepts the `--flag=value` form on `--auto`, closing the assumption ADR-083 flagged as unverified.
- **Small discrepancy noted:** ADR-083's Follow-up item 5 described replacing a "do not attempt `--auto`" line in `claude/CLAUDE.md` — that exact phrasing only ever existed in ADR-031's 2026-07-04 addendum, never in CLAUDE.md itself. I wrote accurate new CLAUDE.md text rather than chasing a find-replace target that wasn't there, and added a short cross-referencing addendum to ADR-031 so the two docs stay consistent.

## Testing

Ran the dev-env root `CLAUDE.md` → Testing section's relevant checks plus new/sibling test suites, all from this worktree post-rebase onto latest `origin/main`, both before and after applying `/review`'s findings (commit `1087349`):

- Hook-script syntax check (`ast.parse` over `claude/scripts/*.py`): pass
- `py -3 claude/scripts/tests/test_pyw_stdio.py`: 7 passed, 0 skipped, 0 failed (confirms `settings.json` wiring resolves and `_winsubp` import discipline holds repo-wide)
- `py -3 claude/scripts/tests/test_pre_auto_merge_checkpoint_gate.py` (new): **30 passed**, 0 skipped, 0 failed (added 4 tests covering the review-fixed bugs: quoted `--auto` detection, `--auto`-in-unrelated-flag-value non-detection, unparseable-tail defensive default, last-match-within-a-comment marker parsing)
- `bash claude/scripts/tests/test-auto-merge-checkpoint-gate.sh` (new): **15 passed**, 0 failed (added 3 behavioral scenarios: commits-page-truncation defensive block, quoted `--auto` end-to-end, `--auto`-as-prose end-to-end)
- `py -3 claude/scripts/tests/test_pre_merge_findings_gate.py` (sibling, regression check): 8 passed, 0 skipped, 0 failed
- `bash claude/scripts/tests/test-merge-findings-gate.sh` (sibling, regression check): 9 passed, 0 failed
- `bash claude/scripts/tests/check-script-path-hygiene.sh`: clean (14 shell files scanned)
- `bash claude/scripts/tests/run-shellcheck.sh`: SKIP — shellcheck not installed on this machine (pre-existing environment limitation, documented as non-blocking by the script itself)

**Tests: 69 passed, 0 skipped, 0 failed** across the five numbered suites above (aggregate, post-fix).

## Findings Disposition

Per [/review](https://github.com/brownm09/dev-env/pull/588#issuecomment-4887974853) (parallel subagent analysis — DIFF_SIZE 731 lines):

- **[security]** Shell-quoted `--auto` (`gh pr merge ... "--auto"`) bypassed the gate entirely (the whitespace-bounded regex didn't recognize a flag whose boundary was a quote character, not whitespace) — **fixed in `1087349`** (tokenize with `shlex.split()` instead of regex; `shlex` strips the quote as part of tokenizing).
- **[correctness]** `--auto` mentioned as prose inside an unrelated `--body` value's quoted text was incorrectly detected as the real flag (same regex, same fix) — **fixed in `1087349`**.
- **[correctness]** Marker parsing bound to the first `review-findings`/`premerge-checkpoints` match within a comment body instead of the last, so a comment quoting an old stale marker for context ahead of its own current one would read the stale value — **fixed in `1087349`** (new `_last_match()` helper).
- **[reliability]** `commits[-1]` assumed `gh pr view --json commits` never truncates; it's an undocumented, paginated GraphQL connection — **fixed in `1087349`** (defensive threshold, fails closed above a conservative page-size guess rather than trust a possibly-truncated array).
- **[performance]** The new hook and `pre-merge-findings-gate.py` each call `gh pr view` independently on the same `--auto` command — **filed [dev-env#590](https://github.com/brownm09/dev-env/issues/590)**, deliberately not fixed here: consolidating would trade away the independent-revertability ADR-083's Decision section explicitly chose a separate script to preserve, for a latency saving on a path (`--auto`) that isn't usable on any repo yet.

Two subagent claims were checked and did not hold up, so are not listed as findings: a claimed missing print-statement in the test runner's failure path (verified the actual file already prints test name + exception — an artifact of an abbreviated transcript, not a real gap), and a comment-edit-vs-list-order staleness concern (the freshness check compares against `createdAt`, which GitHub never changes on edit, so the scenario described is already mitigated by the existing design; the residual risk is the same "PR comments aren't an authenticated channel" limitation ADR-083's Consequences section already documents and accepts).


brownm09:
## Review: feat(hooks): implement ADR-083 mechanical pre-check gate for gh pr merge --auto

### Summary
DIFF_SIZE (731 lines) routed this to parallel subagent analysis (correctness/security; reliability/performance/maintainability). Both passes surfaced real, confirmed issues in the new hook's flag-detection and marker-parsing logic — most notably a genuine bypass (a shell-quoted `"--auto"` skipped the gate entirely) and a false-positive (plain-prose `--auto` inside a `--body` value incorrectly triggered it) in the same regex, both traced to the same root cause: matching on the raw command string instead of a proper shell-quote-aware tokenization. All four blocking findings below were fixed and verified (new/updated tests passing) in commit `1087349` before this comment was posted; one performance finding was deliberately deferred to a filed follow-up (dev-env#590) rather than fixed here, since fixing it would mean coupling the two independently-revertable merge gates ADR-083 explicitly chose to keep separate. Two subagent claims did not survive verification and are omitted below (see closing note) rather than recorded as soft-dismissed findings. **Ready to merge** once this disposition is in the PR body (below).

Also confirms the PR's own new machinery on its first real use: Step 2f resolves `adr_warrant=written` (ADR-083 and ADR-031 are both amended in this very diff), and Step 2b resolves `doc_reconciliation=updated` (README.md and docs/REFERENCE.md are both in the diff) — dogfooding both new checkpoints cleanly.

---

### Blocking Findings

**[security]** `claude/scripts/pre-auto-merge-checkpoint-gate.py:97` (pre-fix line; now fixed)
The original `_AUTO_FLAG_RE` regex required `--auto` to be bounded by whitespace/start/end in the *raw* command string. A shell-quoted invocation, `gh pr merge 123 --squash "--auto"`, is identical to an unquoted `--auto` once the shell strips the quotes and hands `gh` its argv — but the quote character sat at the regex's boundary position instead of whitespace, so the match failed and `wants_auto_merge()` returned `False`. Since this hook is the *sole* mechanism restoring `--auto`'s "impossible to skip" property, with no override token, this was a real bypass: a quoted `--auto` (plausible from tab-completion, copy-paste, or a generated script) silently skipped the entire ADR-warrant/premerge-checkpoints/staleness gate.
**Fix:** Verified live, then fixed by tokenizing the merge tail with `shlex.split()` instead of a raw regex — `shlex` correctly strips shell quoting as part of tokenizing, so `"--auto"` now arrives as the standalone token `--auto`. New test: `test_auto_shell_quoted_is_still_detected`.

---

**[correctness]** `claude/scripts/pre-auto-merge-checkpoint-gate.py:97` (same regex; now fixed)
The same whitespace-bounded regex also over-matched: `--auto` mentioned as prose *inside* an unrelated flag's quoted value (e.g. `--body "please --auto merge this"`) was incorrectly detected as the flag itself, since the regex has no concept of quoting and just looks for a whitespace-bounded substring anywhere in the string. A plain merge whose PR body/commit message happens to mention "--auto" in prose would be forced through the stricter gate it was never subject to.
**Fix:** Same `shlex.split()` fix as above — a quoted multi-word value tokenizes as one token, so `--auto` embedded inside it is no longer mistaken for a standalone flag. New test: `test_auto_inside_unrelated_flag_value_not_detected`.

---

**[correctness]** `claude/scripts/pre-auto-merge-checkpoint-gate.py:170-174` (pre-fix; now fixed)
`_qualifying_comment` used `_MARKER_RE.search(body)` / `_CHECKPOINTS_RE.search(body)`, which bind to the **first** match within a comment's body text, not the last. A comment that quotes an earlier, stale marker for context ("the old review said `<!-- review-findings: blocking=5 ... -->`, now fixed: `<!-- review-findings: blocking=0 ... -->`") before its own real, current marker would silently read the stale value. This pattern is inherited unchanged from the sibling `pre-merge-findings-gate.py`'s identical `.search()`-per-comment approach (so it isn't newly introduced), but this hook's fail-closed, no-override design raises the stakes of the same latent gap considerably.
**Fix:** Added a `_last_match()` helper (iterates `finditer` and keeps the final hit) and use it for both regexes. New test: `test_qualifying_comment_last_match_wins_within_a_single_comment_body`.

---

**[reliability]** `claude/scripts/pre-auto-merge-checkpoint-gate.py:260` (pre-fix; now fixed)
`commits[-1].get("committedDate")` assumed `gh pr view --json commits` always returns the PR's true HEAD commit. Verified live that the underlying GraphQL `commits` field is a paginated connection (it exposes `totalCount` like any other connection) with no documented page size and no `--paginate` equivalent exposed by `gh pr view` — for a PR with more commits than that page, `commits[-1]` would silently be the last commit of a *truncated* page, not the true head, and the freshness check would compare against the wrong date with no error signal.
**Fix:** Added a conservative defensive threshold (`_COMMITS_PAGE_SIZE_SUSPECT = 100`) — at or above it, fail closed with a specific message rather than trust a possibly-truncated array. Not a confirmed exact limit (gh doesn't document one), but costs nothing in the common case and closes the silent-wrong-answer tail case. New behavioral test: scenario [13].

---

### Non-Blocking Findings

**[performance]** `claude/scripts/pre-auto-merge-checkpoint-gate.py` + `claude/scripts/pre-merge-findings-gate.py`
Both hooks independently call `gh pr view` on the same PR when `--auto` fires, fetching overlapping fields (`comments`, `body`, `number`) — one extra network round-trip per `--auto` invocation versus a consolidated fetch.
**Suggestion:** Filed as [dev-env#590](https://github.com/brownm09/dev-env/issues/590) rather than fixed here — ADR-083's own Decision section chose a separate sibling script specifically so a bug in the new code can't regress the proven one and the change stays independently revertable; consolidating the `gh pr view` calls would trade that isolation away for a latency saving that only applies to the currently-unusable (`allow_auto_merge` still `false` everywhere) `--auto` path.

---

### Findings Disposition

- **[security]** Shell-quoted `--auto` bypass — fixed in `1087349`.
- **[correctness]** `--auto` false-positive inside unrelated flag value — fixed in `1087349`.
- **[correctness]** First-match vs. last-match marker parsing — fixed in `1087349`.
- **[reliability]** `commits[-1]` truncation risk — fixed in `1087349`.
- **[performance]** Duplicate `gh pr view` calls — filed [dev-env#590](https://github.com/brownm09/dev-env/issues/590).

---

PR #593, opened 2026-07-06
feat(hooks): register disk-space-check.py under PreToolUse(Bash) too

### LINKED ISSUE #592: Extend disk-space-check.py to PreToolUse(Bash) - catch mid-turn disk exhaustion
## Problem

`claude/scripts/disk-space-check.py` watches free space on `C:` and warns/auto-reclaims at 20GB/10GB thresholds, but it's wired only to `UserPromptSubmit` — it fires once per user prompt, not once per tool call. ADR-085's Context section names this as a known structural gap: disk exhaustion occurring *within* a single long agentic turn (many Bash calls with no new user prompt in between) can outrun this safety net, since it only re-checks between prompts.

ADR-085 explicitly calls this out as "a candidate fast-follow, not built here" (its own mechanism — Bash repo/branch drift detection — is an independent concern).

## Goal

Close this gap by also triggering the same free-space check from `PreToolUse(Bash)`, so a long tool-call-only stretch gets re-checked between individual Bash calls, not just between prompts. `shutil.disk_usage()` is a cheap syscall (not a subprocess spawn), so running it on every Bash call is affordable — consistent with the six existing `PreToolUse(Bash)` hooks already on this hot path.

## Non-goals

- Not changing the WARN_GB/ACT_GB thresholds or the reclaim mechanism itself.
- Not building a brand-new script — reuse `disk-space-check.py`'s existing hook-event-agnostic `main()` (mirrors the `awake-blocker.py` precedent of one script registered under multiple hook events).

Refs: ADR-085 (docs/adr/085-bash-repo-branch-drift-detection.md), dev-env#573

brownm09:
## Summary

- Registers `disk-space-check.py` as a second hook entry under `PreToolUse(Bash)` (in addition to its existing `UserPromptSubmit` registration), closing the mid-turn free-space gap ADR-085's Context section named as a candidate fast-follow: the once-per-prompt check can be outrun by disk exhaustion occurring within a single long agentic turn (many Bash calls, no new prompt in between).
- Reuses the same script unmodified — `main()` already only reads `session_id`/`cwd` from stdin and ignores `hook_event_name`/`tool_name`/`tool_input`, so it works correctly under either hook event with zero parsing changes.
- Extracts a pure `classify_free_space(free_gb, warn_gb, act_gb)` helper from `main()`'s inline threshold if/elif (behavior-preserving — identical boundaries), making the three-band classification unit-testable offline for the first time.

See [ADR-086](docs/adr/086-pretooluse-disk-space-check.md) for the full design rationale, including why this reuses the existing script rather than creating a new one (the `awake-blocker.py` precedent of one script under multiple hook events, not ADR-071's "new script, different decision tree" precedent), and its "No reset-on-recovery" subsection (added after `/review`) addressing repeat within-session exhaustion.

## Testing

Tests: 791 passed, 0 skipped, 0 failed (full existing suite + 1 new file, 41 files total, ~30s)

- New `claude/scripts/tests/test_disk_space_check.py` (CLAUDE.md Testing item 46) — pins `classify_free_space()`'s three bands (ok/warn/act) and both threshold boundaries (inclusive on the healthier side).
- `py -3 -c "import ast..."` syntax check over all `claude/scripts/*.py` — clean.
- `py -3 claude/scripts/tests/test_no_crude_command_substring_checks.py` (item 39) — clean, no crude substring checks introduced.
- Suppression grep, test-integrity grep, deleted-test-file check — all clean, no matches.
- Test-integrity check manually extended to this diff's Python files (`/review` Step 2e's automated scanners are JS/TS-only and can't inspect them): no skip markers, deleted tests, lowered thresholds, or test-input-shaped hardcoding found.
- Not run: shellcheck / pre-push self-test / `get-project-item.sh` smoke test / `merge-stale-pr` self-test / auto-merge-checkpoint-gate self-test — N/A, no shell scripts or those specific files touched by this PR.

Reviewed via `/review` ([comment](https://github.com/brownm09/dev-env/pull/593#issuecomment-4889120549)). No blocking findings.

- **[maintainability] Non-blocking** — Step 2e's JS/TS-only integrity scanners can't inspect this diff's Python files. **Fixed in `cfdbb2c`**: manually verified clean (see Testing section note above) — no code change needed, the verification itself was the fix.
- **Question** — Should the marker gate reset on recovery to catch a repeat within-session exhaustion episode? **Fixed in `cfdbb2c`**: added ADR-086's "No reset-on-recovery for a repeat within-session exhaustion (accepted gap)" subsection, documenting this as a deliberate, reasoned tradeoff rather than an oversight.

Closes #592


brownm09:
## Review: feat(hooks): register disk-space-check.py under PreToolUse(Bash) too

### Summary
Small (331-line), well-scoped change that reuses an existing script rather than adding a new one, backed by a clear ADR justifying that call and a full green test suite (791 passed). No blocking findings. One non-blocking finding (a mechanical language-scope reminder from the test-integrity gate — trivially resolved) and one genuine design question about a repeat-exhaustion edge case the ADR's judgment calls don't explicitly address. Ready to merge once both are addressed.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/disk-space-check.py`, `claude/scripts/tests/test_disk_space_check.py`
Step 2e's automated skip-marker/deletion/threshold scanners target JS/TS test frameworks (Jest, Vitest, Mocha); this diff's test-relevant files are Python. A clean result from those JS/TS-only patterns is not itself evidence the integrity policy held for this Python code.
I manually read both files in full during this review: no skip markers (`pytest.mark.skip`, `.skip(`), no deleted tests, no lowered thresholds, and no test-input-shaped hardcoding in `classify_free_space()` — genuinely clean.
**Suggestion:** Add a one-line note to the PR body's Testing section recording that the integrity check was manually extended to the Python files in this diff, so the verification is on record rather than only in this review.

---

### Questions for Author

**Question:** Is "fires at most once per session per band" intended to mean *ever*, or only for a *sustained* low-space condition — i.e., if space recovers (reclaim succeeds, or a worktree gets pruned) and then drops into the same band a second time later in the same session, should the second drop re-warn/re-reclaim?
**Context:** `_already_fired`/`_mark_fired` in [`claude/scripts/disk-space-check.py`](claude/scripts/disk-space-check.py) key the marker file on `(session_id, band)` only, with no recovery-reset path, so a band that already fired once stays silent for the rest of the session even after a full recovery and a later, independent re-drop into the same band. ADR-086's "Marker-file gate intentionally shared across both event types" judgment call explains why sharing the gate *between* the two hook registrations is correct, but doesn't address this *within-session recurrence* case — and this PR raises its likelihood, since `PreToolUse(Bash)` now samples far more often than the old once-per-prompt check, making a recover-then-redegrade cycle more observable within one long session.
**Tradeoffs:** (a) Keep as-is — simplest, and a second "act"-band episode in one session still benefits from the *first* firing's reclaim having already run, so the practical gap is a missing second warning/spawn, not silent total inaction. (b) Reset the marker once free space is observed back above `WARN_GB` — closes the gap exactly, at the cost of one extra comparison and a slightly more complex marker lifecycle. (c) Time-box the marker (re-arm after N minutes) — bounds staleness without tracking recovery explicitly, but introduces a second tunable with no obvious right default.

---

### Findings Disposition
Per ADR-028, the non-blocking finding above must be fixed in this PR or filed as a tracked issue and linked before merge. Before `gh pr merge`, record its disposition (fixed in `<sha>` | filed `#N`) in a "Review findings disposition" section of the PR body; `pre-merge-findings-gate` blocks the merge until that disposition exists (ADR-039).

---

PR #598, opened 2026-07-06
[docs] Force post-merge tile enumeration + re-key tile checkpoint to merged state (#595)

brownm09:
## Summary

Fixes the recurring skip of the post-merge **tile checkpoint** (global `claude/CLAUDE.md` → Git Workflow; [ADR-046](https://github.com/brownm09/dev-env/blob/main/docs/adr/046-post-merge-followup-tiles.md)). Per [#595](https://github.com/brownm09/dev-env/issues/595), the checkpoint was missed repeatedly because of three mechanisms: a discretionary `if`-trigger with no forcing function, no required artifact/hook so a skip was invisible, and a trigger keyed to the literal `gh pr merge` command (blind to auto-merge and the two-step REST merge).

This is a **docs/governance change** — the wording floor. It does two things:

1. **Forces an explicit enumeration** at every post-merge checkpoint — record each considered follow-up as `→ tiled` or `→ not tiled, because …`. **"No follow-ups" is valid only as the visible result of that scan, never a bare assertion.**
2. **Re-keys the trigger** from "after `gh pr merge`" to **"when a PR reaches merged state, however it merged"** (a `gh pr merge` you ran, the two-step REST merge, or auto-merge landing it while away).

Also adds the **tile-don't-ask** facet (observed 2026-07-06 in the #595 thread): identifying deferred work at the checkpoint means tiling it yourself, not deflecting the triage into a user-facing question.

### Changes
- `claude/CLAUDE.md` — rewrite the "Capture post-merge follow-ups as tiles" bullet's opening sentence (mechanisms 1 + 2 + tile-don't-ask); the rest of the bullet is byte-for-byte unchanged.
- `docs/adr/046-post-merge-followup-tiles.md` — new **2026-07-05 addendum** documenting the three mechanisms, the refinement, the relationship to the existing **command-keyed** ADR-060 hook (blind to auto-merge), and the motivating incident (lifting-logbook PR #700); References extended.
- `docs/adr/INDEX.md` — add `enumeration, merged-state` tags to the ADR-046 row.
- `docs/REFERENCE.md` — re-key the "Post-merge follow-up tiles (chips)" runbook to merged state to match (otherwise the runbook would contradict the re-keyed rule).

### Deferred (separate follow-up, per #595)
[...]

Issue #595, opened 2026-07-06
[meta] Post-merge tile checkpoint is skippable — force follow-up enumeration + re-key trigger to merged-state


brownm09:
## Problem

The post-merge follow-up **tile checkpoint** (global `claude/CLAUDE.md` → Git Workflow → "Capture post-merge follow-ups as tiles"; [ADR-046](docs/adr/046-post-merge-followup-tiles.md)) is repeatedly missed until the user explicitly asks for the tile. This is a recurring follow-through failure, not a one-off.

**Most recent incident:** lifting-logbook PR #700 (settings hub, closes #679, merged 2026-07-05). In the same message the session flagged that an idle worktree was now removable *and* asserted "the finalization work surfaced no new follow-ups" — creating no tile. The user had to prompt twice ("Where's the tile, then?" and then "why do I keep having to ask?").

## Root cause — three mechanisms

1. **Discretionary trigger with no forcing function.** The rule fires "*if* the work surfaced any follow-ups." That `if` is a self-assessment made at the moment the agent is most motivated to be done, and it gets resolved to "none" as a bare assertion. Nothing forces the agent to *enumerate* candidates before concluding zero. Completion-bias reliably pushes the judgment to "none."

2. **No required artifact and no hook, so a skip is invisible.** Contrast the review-findings gate: the agent MUST write a "Review findings disposition" section and the `pre-merge-findings-gate` hook blocks the merge without it. A skipped tile checkpoint leaves no trace — the only detector is the user noticing the absence.

3. **Trigger keyed to "after `gh pr merge`."** When a PR lands via **auto-merge** or the **two-step REST merge** (as #700 did — auto-merge landed it while the agent was away, and post-merge was pure bookkeeping), the literal `gh pr merge` command never runs, so the checkpoint's salience is already low exactly when it should fire.

## Proposed fix (amends ADR-046)

1. **Force an explicit enumeration** at every post-merge checkpoint: record "Follow-ups considered: … → tiled (`task_id`/#N) or → not tiled, because <reason>", covering out-of-scope fixes, deferred work, tech debt, and ideas-noticed-in-passing. **"No follow-ups" is valid only as the visible result of that enumeration — never as a bare assertion.**

2. **Re-key the trigger** from "after `gh pr merge`" to **"when a PR reaches merged state, however it merged"** (a `gh pr merge` you ran, the two-step REST merge, or auto-merge landing it while you were away).

[...]

PR #604, opened 2026-07-06
feat(hooks): state-keyed post-merge tile-enumeration Stop gate (#599)

### LINKED ISSUE #599: [meta] Enforce the post-merge tile checkpoint with a state-keyed hook (catch auto-merge)
## Problem

The post-merge tile checkpoint (ADR-046, re-keyed to merged-state in #595 / #598) is now a documented rule but still has **no state-keyed enforcement**. The existing hook `post-merge-tile-checkpoint.py` (ADR-060) is **command-keyed** (`"gh pr merge" in command`): it fires on a `gh pr merge` you run — including the two-step workaround's `gh pr merge <N> --squash` first step — but is **blind to auto-merge**, which lands the PR server-side with no local `gh pr merge`. That auto-merge case is exactly the lifting-logbook PR #700 incident that motivated #595.

The `claude/CLAUDE.md` wording (landed in #598) is the floor; the durable enforcement is a hook.

## Proposed

Add a **state-keyed** Stop/PostToolUse hook — analogous to `pre-merge-findings-gate` — that observes a merged-state transition within a session and blocks/reminds when there is **no recorded tile-enumeration** ("Follow-ups considered: … → tiled (`task_id`/#N) or → not tiled, because <reason>"). Detect merged state via the PR reaching MERGED (e.g. a `gh pr view`-observed transition, or a Stop-time scan), **not** the `gh pr merge` command string.

Open design question: extend `post-merge-tile-checkpoint.py` (ADR-060) to add a state-keyed path, or add a sibling hook. The state-keyed gate augments/supersedes the command-keyed reminder and additionally covers auto-merge and pure-`gh api` merges.

## Context
- Parent: #595 — mechanism 3 / proposed-fix 3 explicitly scoped this hook as **separate** follow-up work.
- Wording landed: #598 (ADR-046 2026-07-05 addendum, `claude/CLAUDE.md`, `docs/adr/INDEX.md`, `docs/REFERENCE.md`).
- Enforcement precedent: `pre-merge-findings-gate` ([ADR-039](docs/adr/039-merge-gate-findings-enforcement.md)); command-keyed hook: [ADR-060](docs/adr/060-post-merge-tile-checkpoint-hook.md).

brownm09:
## Summary

Adds the **state-keyed** post-merge tile-checkpoint enforcement hook deferred from [#598](https://github.com/brownm09/dev-env/pull/598) and scoped as separate follow-up in [#595](https://github.com/brownm09/dev-env/issues/595). Closes #599.

The existing `post-merge-tile-checkpoint.py` ([ADR-060](https://github.com/brownm09/dev-env/blob/main/docs/adr/060-post-merge-tile-checkpoint-hook.md)) is **command-keyed** (`"gh pr merge" in command`): it fires on a `gh pr merge` you run but is **blind to auto-merge** (GitHub merges server-side, no in-session command) and pure `gh api` merges — the lifting-logbook PR #700 blind spot. This adds `stop-tile-enumeration-gate.py`, a **Stop** hook that scans the just-ended transcript and **blocks the stop (exit 2)** when a PR reached MERGED state this session *by any path* but **no** tile-enumeration artifact was recorded — the Stop-hook analog of `pre-merge-findings-gate` ([ADR-039](https://github.com/brownm09/dev-env/blob/main/docs/adr/039-merge-gate-findings-enforcement.md)).

## Design decision: new Stop hook, not an extension of ADR-060

- **Stop, not PostToolUse/PreToolUse** — the tile-enumeration is a *post-merge artifact* written after the merge; only at turn-end can the gate verify "merged **and** enumerated". A PostToolUse hook fires too early and can't see auto-merge; a PreToolUse gate has no command to gate on auto-merge. Same proven shape as `posttooluse-inert-advisory.py`.
- **State-keyed detection** — a `gh pr merge` success marker, a `gh api .../pulls/N/merge` with `"merged":true`, **or** a `gh pr view` MERGED state correlated with a PR the session created/enqueued (the auto-merge case). Correlation prevents a false-fire from merely inspecting an unrelated old merged PR.
- **Complement, not replace** — command-keyed = immediate in-the-moment nudge; state-keyed = Stop-time verification covering all paths. `post-merge-tile-checkpoint.py` is unchanged. Bonus: the Stop event still fires in background/SDK sessions where every PostToolUse hook is inert ([ADR-053](https://github.com/brownm09/dev-env/blob/main/docs/adr/053-posttooluse-hooks-inert-in-background-sessions.md)) — so this is the only tile enforcement that survives there.
- **Enumeration detection** — a `spawn_task` tile, or the prescribed text (`Follow-ups considered … → tiled / → not tiled, because …`). A **bare "no follow-ups" does NOT satisfy** the gate — that is the exact #700 skip the ADR-046 refinement invalidated.
- **Exit 2, once per session** — advisory reminders are ignored (mechanism 2 of #595), so it blocks to force acknowledgment; loop-safe via the `stop_hook_active` flag + a once-per-session scratch sentinel. Honors an explicit "skip tiles" instruction. Fail-open on any error.

Full rationale, limitations, and alternatives: [ADR-088](https://github.com/brownm09/dev-env/blob/main/docs/adr/088-state-keyed-tile-enumeration-gate.md).

## Changes
- **`claude/scripts/stop-tile-enumeration-gate.py`** — new Stop hook (pure transcript scan; no gh/network/subprocess).
- **`claude/scripts/tests/test_stop_tile_enumeration_gate.py`** — offline pure-helper suite + subprocess end-to-end layer.
- **`claude/settings.json`** — registered in the `Stop` list after `posttooluse-inert-advisory.py`.
- **`docs/adr/088-…md`** + **`docs/adr/INDEX.md`** — ADR documenting the state-keyed vs command-keyed decision.
- **`README.md`** + **`docs/REFERENCE.md`** — Stop-hook table rows + tiles-runbook cross-ref (doc-reconciliation).
- **`CLAUDE.md`** — Testing item 48.

## Testing
Ran from the worktree root:
- **New suite** (`py -3 claude/scripts/tests/test_stop_tile_enumeration_gate.py`) — merged detection across all three paths incl. auto-merge correlation; the non-matches (old-PR inspection, `--auto` enqueue alone, `--help`, heredoc/subshell anchoring); enumeration incl. the bare-"no follow-ups" rejection; skip-override incl. tool_result exclusion; `evaluate()` composition; `iter_bash_calls` id-pairing; cp1252 encodability; and the subprocess end-to-end layer (exit-2 stderr-only, sentinel re-fire suppression, `stop_hook_active` guard). **`Tests: 32 passed, 0 skipped, 0 failed`.**
- **Hook-script syntax check** (Testing item 1) — all `claude/scripts/*.py` parse.
- **Crude command-substring AST gate** (item 39) — `Tests: 13 passed, 0 skipped, 0 failed` (uses `split_top_level`-anchored regex, not `X in command`).
- **Sibling no-regression** — `test_post_merge_tile_checkpoint` (9), `test_hookio` (57), `test_hookutil` (8) all pass.
- **`settings.json`** — valid JSON.

## Coverage gate
New behavior is fully covered by the new suite (pure helpers + end-to-end). No deferral.

## Suppression / test-integrity (ADR-026 / ADR-029)
Both greps against `origin/main` are **clean** — no suppressions, no skip markers, no `0 skipped` violations. Skipped count is 0.

## Doc reconciliation
New event-driven hook → README Hooks table + REFERENCE Hooks section updated (+ the tiles runbook cross-ref). ADR + INDEX added. Numbering verified against `origin/main` (highest ADR 087 → 088; highest Testing item 47 → 48; no collision).

## ADR-warrant
Warranted (new hook + settings change) → **ADR-088** written and indexed in this PR.

`/review` (2 Opus subagents) surfaced 2 blocking + 5 non-blocking findings; no hard blocker. All addressed in this PR:
- **[correctness] observed-MERGED PR number (A1)** — fixed in `78da066` (prefer authoritative JSON `"number"`).
- **[correctness] compact-summary/`isMeta` skip-waiver (A2)** — fixed in `78da066` (synthetic user records no longer waive).
- **[correctness] cross-repo same-number false-fire (A4)** — fixed in `78da066` (correlate on `(repo, number)`).
- **[correctness] malformed-record silent-disable (A3)** — fixed in `78da066` (drop non-dict records + `isinstance` guards + `evaluate` in try/except).
- **[performance] redundant `split_top_level` (~5×)** — fixed in `78da066` (single pass per command).
- **[performance] unbounded per-turn scan** — fixed in `78da066` (`"merged"` substring pre-filter; documented in ADR-088).
- **[maintainability] extract shared transcript readers to `_hookutil`** — filed as #605 (judgment call; both reviewers non-blocking).

+7 regression tests (39 total, `Tests: 39 passed, 0 skipped, 0 failed`). Review comment: https://github.com/brownm09/dev-env/pull/604#issuecomment-4889737053


brownm09:
## Review: feat(hooks): state-keyed post-merge tile-enumeration Stop gate (#599)

### Summary
Merge-ready after the fixes below (all applied in `78da066`). The design is sound — exit-2/stderr discipline, sentinel write-before-emit, `stop_hook_active` guard, id-pairing, enumeration detection (incl. the bare-"no follow-ups" rejection and Unicode arrows), and the direct `gh pr merge`/`gh api` detection are all correct, and the transcript scan has no injection/unsafe-path surface. Two Opus subagents (correctness/security; reliability/perf/maintainability) surfaced **no hard blocker**; the highest-severity items were two that *silently weaken the gate along its primary paths* (auto-merge PR-number resolution and a compact-summary skip-waiver), both now fixed.

---

### Blocking Findings

**[correctness]** `claude/scripts/stop-tile-enumeration-gate.py` — observed-MERGED PR number
The auto-merge path took the PR number from the first positional integer after the `gh pr view` verb, only falling back to the authoritative JSON `"number"` when that returned `None`. A numeric flag value could win over the correct `"number"`, silently missing the auto-merge — the hook's headline capability.
**Fix:** Prefer the JSON `"number"` for observed state; positional is the fallback.

**[correctness]** `claude/scripts/stop-tile-enumeration-gate.py` (`skip_override`) — synthetic-record waiver
`skip_override` accepted any `type=="user"` record's text, including synthetic `isMeta` / `isCompactSummary` records. A `/compact` summary that merely *restates* an earlier "skip tiles" mention would waive the gate for the whole session — and the workflow prompts `/compact` right after PR-create, making this likely.
**Fix:** Skip `isMeta`/`isCompactSummary` records; only genuine user text waives. Regression test added.

---

### Non-Blocking Findings

**[correctness]** `stop-tile-enumeration-gate.py` (`session_merged_prs` correlation) — cross-repo same-number
The observed∩acted correlation was by PR number only. A session that creates its own PR #N and later inspects an unrelated already-merged PR #N in a *different* repo would false-fire (safe direction — over-fires a reminder — but undermines the "can't mistake an unrelated PR" guarantee).
**Fix:** Correlate on `(repo, number)`; a `None` (cwd-inferred) repo falls back to a number match, preserving same-repo true positives. Fixed in `78da066`.

**[correctness]** `stop-tile-enumeration-gate.py` (`load_records` / helpers) — malformed record silently disables gate
A non-dict transcript line (`null`/number/string/array) or a non-dict `message` raised `AttributeError` in a helper, caught only by the outer `__main__` guard → silent exit 0, disabling an otherwise-fireable session. Low impact today (no such records in real transcripts) and safe direction, but contradicts the robustness claims.
**Fix:** Drop non-dict records at parse time, add `isinstance` guards, and wrap `evaluate()` in `main`'s try/except so a scan failure is a deliberate skip. Fixed in `78da066`.

**[performance]** `stop-tile-enumeration-gate.py` (`session_merged_prs`) — redundant `split_top_level`
The three merge-path helpers each re-ran `split_top_level(command)` (~5× per command per Stop) — the dominant cost (~127 ms of ~176 ms on a 2.4 MB transcript).
**Fix:** Single pass computing segments once per command. Fixed in `78da066`.

**[performance]** `stop-tile-enumeration-gate.py` (`main`) — unbounded per-turn scan in no-merge sessions
Stop fires every turn; a no-merge session re-read and re-parsed the whole transcript each time with no early-out (the state design can't resolve early the way `posttooluse-inert-advisory.py` does).
**Fix:** Cheap `"merged"` substring pre-filter (every merge signal contains it) short-circuits the common no-merge case before any JSON parse. Fixed in `78da066`; documented in ADR-088 Limitations.

**[maintainability]** `stop-tile-enumeration-gate.py` (transcript readers) — duplication vs `posttooluse-inert-advisory.py`
`load_records`/`iter_bash_calls`/`_result_text` (~80 lines) are near-identical to the sibling's; two files now parse the same transcript format and can drift. ADR-088 documents the replication as deliberate (the `_first_line` precedent), but the repo made the opposite call for this class (ADR-050/ADR-064).
**Suggestion:** Extract to `_hookutil` returning `(command, output, cwd)`. Filed as #605 (deferred — judgment call, both reviewers non-blocking).

---

### Findings Disposition
- **[correctness] observed PR-number (A1)** — fixed in `78da066`
- **[correctness] compact-summary skip-waiver (A2)** — fixed in `78da066`
- **[correctness] cross-repo same-number (A4)** — fixed in `78da066`
- **[correctness] malformed-record silent-disable (A3)** — fixed in `78da066`
- **[performance] redundant split_top_level** — fixed in `78da066`
- **[performance] unbounded per-turn scan** — fixed in `78da066` (+ ADR-088 note)
- **[maintainability] extract shared readers** — filed #605

---

brownm09:
Following up on the review question raised here about whether this Stop gate's **exit 2** could delay `awake-blocker.py`'s sleep-lock release — since `awake-blocker` (stop) is registered *after* `stop-tile-enumeration-gate.py` in the `claude/settings.json` Stop list.

**Answer: no.** Claude Code runs all matching hooks for an event **in parallel** — a blocking exit 2 does **not** short-circuit sibling hooks — so `awake-blocker`'s Stop-time sleep-release fires regardless of this gate's exit 2. The settings.json Stop order is not an execution order (non-deterministic parallel dispatch); reordering `awake-blocker` would be a no-op.

Primary source (https://code.claude.com/docs/en/hooks-guide):
> "all matching hooks run in parallel … every hook's command runs to completion before Claude Code merges the results. One hook returning `deny` doesn't stop sibling hooks from executing."

Documented in **ADR-088** (new "Stop-hook parallelism" subsection) and the `docs/REFERENCE.md` Stop section via **#612 / PR #613** (merged `daf4aa3`). No code/settings change — the concern rests on a sequential-short-circuit model Claude Code's hook dispatch doesn't have.

Bonus correction found in passing: `journal-stop-check.py` never actually exits 2 (it's exit-0/stdout), so this gate is the **first** blocking Stop hook, not a second — REFERENCE.md's "via exit 2" line for it was corrected too. Whether that exit-0 reminder actually reaches Claude is now tracked as a follow-up.

PR #608, opened 2026-07-07
[docs] Correct stale "allow_auto_merge not yet made for any repo" claim

### LINKED ISSUE #607: [docs] Correct stale "allow_auto_merge not yet made for any repo" claim (ADR-083 item 7)
## Problem

The global `claude/CLAUDE.md` (Git Workflow → "Auto-merge is off by design" bullet) and
ADR-083 both currently claim `allow_auto_merge` "stays `false` across every `brownm09/*` repo" /
"not yet made for any repo" (ADR-083 Follow-up item 7). This is stale.

## Live check (2026-07-06/07)

`gh api repos/brownm09/<repo> --jq .allow_auto_merge` against the repos referenced in dev-env/CLAUDE.md:

| Repo | `allow_auto_merge` |
|---|---|
| lifting-logbook | **`true`** |
| dev-env | `false` |
| career-playbook | `false` |
| engineering-journal | `false` |
| win11-init-tools | `false` |

lifting-logbook is the sole exception. Corroborated independently: a `gh pr merge --squash --auto`
was run successfully in a lifting-logbook session on 2026-07-06, which requires
`allow_auto_merge: true` at the GitHub API level regardless of the local
`pre-auto-merge-checkpoint-gate.py` hook's own checkpoint gate.

## Root cause traced, not just drift

No `.github/settings.yml` exists in lifting-logbook (repo settings aren't declarative there), and
personal GitHub accounts have no settings-change audit-log API, so the exact toggle moment can't
be reconstructed directly. But circumstantial evidence ties it to the same work stream as
[lifting-logbook#718](https://github.com/brownm09/lifting-logbook/issues/718) ("Add mandatory
Review Gate required status check", opened 2026-07-06T03:59 — exactly the item-7 decision
ADR-083's 2026-07-06 addendum was written to anticipate). Sub-issue
[#720](https://github.com/brownm09/lifting-logbook/issues/720) and the workflow it ships
([PR #722](https://github.com/brownm09/lifting-logbook/pull/722)) were opened the same day. This
reads as **intentional-in-direction, not unnoticed drift**.

## But it's premature relative to ADR-083's own recommended sequencing

The 2026-07-06 addendum's **Recommendation** was explicit: pair the `allow_auto_merge` flip with a
required "Review Gate" GitHub Actions check *before or alongside* flipping the toggle, specifically
because the toggle alone reopens the web-UI/GitHub-triggered-merge gap. As of this writing:

- `review-gate.yml` exists only on **open, unmerged** [PR #722](https://github.com/brownm09/lifting-logbook/pull/722) — not on `main`.
- Even once merged, it still needs sub-issue 2 of #718 (the branch-protection mutation) before it's *required*.
- Until both land, a human merging via the lifting-logbook web UI, or GitHub's own auto-merge
  firing once CI goes green, bypasses **all** review-findings enforcement — live, right now.

## Fix

1. Correct the global CLAUDE.md bullet to state the per-repo, live-checked nature of
   `allow_auto_merge` instead of a blanket "false everywhere" claim — point at live verification
   (`gh api repos/brownm09/<repo> --jq .allow_auto_merge`) rather than a snapshot that will go
   stale again.
2. Add a new dated addendum to ADR-083 recording the corrected per-repo status, the intentional-
   but-premature characterization, and the currently-live sequencing gap on lifting-logbook.
3. Do **not** flip any toggle as part of this fix — that's a judgment call for whoever owns
   lifting-logbook's rollout timeline, flagged here and to the user directly, not resolved by a
   docs PR.

## Scope

Docs-only: `claude/CLAUDE.md` + `docs/adr/083-auto-merge-checkpoint-gate.md`.

### LINKED ISSUE #610: [docs] Document that EnterWorktree targets the session's primary repo, not an arbitrary target repo
## Problem

`EnterWorktree` creates its new worktree inside **the session's primary repository** — not an
arbitrary repo named in the task. When a session's primary working directory is repo A (e.g.
lifting-logbook) but the actual task needs isolated work in repo B (e.g. dev-env), calling
`EnterWorktree` still creates the worktree under repo A, silently. Nothing about the tool call
itself signals the mismatch.

## Concrete incident

A session launched with primary directory `lifting-logbook` needed to edit `dev-env`'s
`claude/CLAUDE.md` and an ADR. Calling `EnterWorktree({name: "docs/issue-607-..."})` created
`lifting-logbook/.claude/worktrees/docs+issue-607-...` — an empty, useless worktree in the wrong
repo. Caught immediately because the very next step (finding the CLAUDE.md text to edit) surfaced
the mismatch, but a less careful session could easily proceed to edit files there, discover nothing
matches, and waste real effort — or worse, silently commit unrelated changes into the wrong repo's
history.

## Why this matters

Per `EnterWorktree`'s own tool description, the recommended recovery once already inside a
mis-targeted worktree — `ExitWorktree` — is explicitly scoped to "only call when the user asks,"
so a session can't self-correct by exiting; the practical fix is to leave the stray worktree in
place and instead reach the correct repo via a directly-issued `git worktree add -C <target-repo>`
(or equivalent), never `cd`-ing there, per the existing "Worktree Bash commands must not `cd` to
the main repo" rule this is adjacent to.

## Fix

Add a bullet to the global CLAUDE.md (Git Workflow section, near the existing worktree-safety
bullets) stating: `EnterWorktree` targets the session's primary repo only; for a different repo,
use `git worktree add` directly against that repo's path, and interact with it via explicit paths
or `git -C`, never `cd`.

## Scope

Docs-only: `claude/CLAUDE.md`. Small, single-bullet addition — bundling into the already-open
dev-env PR #608 (same worktree, same file) rather than a separate branch/PR cycle.

brownm09:
## Summary

Corrects a stale claim in the global CLAUDE.md and ADR-083: both currently say `allow_auto_merge`
"stays `false` across every `brownm09/*` repo" / "not yet made for any repo" (Follow-up item 7). A
live check found lifting-logbook already flipped to `true`.

## What changed

- **`claude/CLAUDE.md`** (Git Workflow → Auto-merge bullet): replaced the blanket "false
  everywhere" claim with guidance to verify live per repo (`gh api repos/brownm09/<repo> --jq
  .allow_auto_merge`), since a web-UI toggle leaves no git history and can drift silently. Points
  to ADR-083's dated addenda for current per-repo status instead of hardcoding a snapshot that will
  go stale again.
- **`docs/adr/083-auto-merge-checkpoint-gate.md`**: added a dated **Addendum (2026-07-07)**
  recording the live per-repo table, the evidence tying lifting-logbook's flip to its in-flight
  [lifting-logbook#718](https://github.com/brownm09/lifting-logbook/issues/718) Review Gate work
  (intentional-in-direction, not unnoticed drift), and the currently-live gap: the companion
  required check that ADR-083's own 2026-07-06 addendum recommended pairing with the toggle
  ([lifting-logbook#722](https://github.com/brownm09/lifting-logbook/pull/722)) is still an
  unmerged, non-required PR. Also added a one-line pointer near the top Status line so a reader
  doesn't have to reach the end of the file to learn the Context/Implementation sections describe
  dated snapshots, not current state.

## What this PR deliberately does NOT do

Flip any `allow_auto_merge` toggle, merge lifting-logbook#722, or make Review Gate required. Those
are judgment calls for whoever owns lifting-logbook's rollout timeline — this PR only corrects the
documentation and surfaces the gap; disposition is flagged to the user directly outside this PR.

## Testing

Docs-only change (2 markdown files, no code, no hook scripts). Per `## Testing` item 4 (docs-only
guard): `grep -n 'date -u' claude/CLAUDE.md` — no matches. Suppression grep
(`git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|...)'`) and
test-integrity grep — both clean (no matches; N/A for a markdown-only diff, confirmed rather than
assumed). No hook scripts, ADR index entries, or code touched, so the remaining numbered Testing
items don't apply.

**Tests:** N/A — no automated test suite covers `claude/CLAUDE.md` / `docs/adr/*.md` content; this
is the project's documented exception (docs-only change, existing-tests-must-pass tier).

`/review` ([comment](https://github.com/brownm09/dev-env/pull/608#issuecomment-4919191404)) found
one blocking finding and one non-blocking question, both fixed in `ad6c5e3`:

- **[documentation] Blocking** — the 2026-07-07 ADR-083 addendum described lifting-logbook#722 as
  "open, unmerged... not on `main`," which went stale before this PR itself merged: #722 merged
  2026-07-08 (`0d9dce4`) while this PR was still open. **Fixed:** appended a third dated
  `## Addendum (2026-07-08)` recording the merge, that `review-gate.yml` now ships and reports on
  lifting-logbook `main`, and that it is not yet a *required* check (lifting-logbook#718 stays open
  for that piece) — same append-only-addenda convention as the two sections above it.
- **[question, non-blocking]** — whether `claude/CLAUDE.md`'s auto-merge bullet should embed a
  dated per-repo snapshot at all, given a second copy of the same fact can drift from the ADR
  addenda the same way the original blanket claim drifted from live GitHub state. **Fixed:**
  dropped the specific per-repo value from the bullet, keeping only "at least one repo differs from
  `false`, see the addenda" — cuts the duplicate-state surface without losing the pointer.

Re-ran the docs-only guard, suppression grep, and test-integrity grep against `ad6c5e3` — all three
still clean (no matches).

Closes #607

---

## Follow-up commit: EnterWorktree targeting gotcha (dev-env#610)

While isolating this fix into a dev-env worktree, `EnterWorktree` silently created a worktree in
the **wrong repo** (lifting-logbook, this session's primary directory) because it targets the
session's primary repo only, not an arbitrary repo named in the task. Caught immediately, no harm
done, but genuinely reusable across sessions — filed
[dev-env#610](https://github.com/brownm09/dev-env/issues/610) and added a one-bullet documentation
fix to `claude/CLAUDE.md` in a second commit on this same branch, next to the existing worktree
`cd`-safety rule it's adjacent to. Same testing coverage as above (docs-only guard + suppression
grep, both clean).

Closes #610


brownm09:
## Review: [docs] Correct stale "allow_auto_merge not yet made for any repo" claim

### Summary
Solid, well-evidenced documentation correction — the CLAUDE.md bullet rewrite and the EnterWorktree
follow-up commit are both accurate and appropriately hedged against future drift. One blocking issue:
the new ADR-083 addendum describes lifting-logbook PR #722 as "open, unmerged... not on main," but
#722 merged today (2026-07-08, commit `0d9dce4`) — after this PR's content was written, before this
PR itself merges. Shipping the addendum as-is would land a "here's the current state" doc that is
already wrong at merge time, which cuts against the exact live-verification point the PR is making.
Small, fast fix (append a third dated addendum, matching the file's own convention) — otherwise
ready to merge.

---

### Blocking Findings

**[documentation]** `docs/adr/083-auto-merge-checkpoint-gate.md:557` (Addendum (2026-07-07) section)
The addendum states lifting-logbook's paired required-check work "exists only on an open, unmerged
PR (lifting-logbook#722) — not on `main`" and frames the gap as unresolved "while #722 is in flight."
As of 2026-07-08, [lifting-logbook#722](https://github.com/brownm09/lifting-logbook/pull/722) is
merged (`0d9dce4`, 2026-07-08T20:51:47Z) — `review-gate.yml` is live on `main` and reporting a real
pass/fail on every PR. It is *not* yet a required branch-protection check (lifting-logbook#718 stays
open for that), so the underlying gap this addendum is warning about is only partially closed, not
fully — but the specific "not on main" / "in flight" framing is now factually wrong. This diff hasn't
touched `main` yet, so there's no historical-record cost to getting it right before it does; landing
it stale would undercut the credibility of the exact "verify live, don't trust a snapshot" argument
this PR is making one section above.
**Fix:** Append a `## Addendum (2026-07-08)` section — consistent with this file's own established
pattern of dated, append-only addenda (the 2026-07-06 and 2026-07-07 sections directly above it) —
noting: #722 merged, `review-gate.yml` ships and reports on `main`, but is not yet a required check;
lifting-logbook#718 stays open for that remaining piece. Do this in the same PR rather than editing
the 2026-07-07 prose in place, to keep the addendum-chain's own audit trail intact.

---

### Questions for Author

**Question:** Should the auto-merge bullet in `claude/CLAUDE.md` restate the current known per-repo
values at all (even dated), or should it point to ADR-083's addenda without repeating a snapshot?
**Context:** The rewritten bullet in [`claude/CLAUDE.md:96`](claude/CLAUDE.md) adds "(as of
2026-07-07, `lifting-logbook` is flipped `true` ... while every other repo checked is still
`false`)" alongside the "verify live" instruction. This is the same class of claim (a point-in-time
snapshot embedded in a rule file) that this PR exists to correct — just now dated and paired with a
pointer to the ADR's addenda as the authoritative source.
**Tradeoffs:** (a) Keep the dated snapshot — convenient quick reference, but it's a second copy of
the same fact the ADR addendum tracks, and the two can now drift from each other (e.g. if
lifting-logbook's toggle is later reverted, per the ADR's own still-open "Disposition" judgment
call) the same way the original bug drifted from live GitHub state, just one hop removed. (b) Drop
the parenthetical entirely and only point to "ADR-083's dated addenda" — zero duplicate state to
keep in sync, but a reader loses the one-line current-best-guess without an extra doc hop. (c) Keep
it but drop the specific values, e.g. "as of 2026-07-07, exactly one repo differs from the default —
see ADR-083" — flags that an exception exists without committing to which repo or which direction,
cutting the drift surface roughly in half.

---

### Findings Disposition

Per ADR-028/ADR-039, the one blocking finding above must be fixed in this PR (a small addition, not
a revert of existing content) or filed and linked before merge — record the disposition in a "Review
findings disposition" section of the PR body before running `gh pr merge`. The question above is
non-blocking judgment; answer inline or via a follow-up commit, whichever the author prefers.

---

PR #611, opened 2026-07-08
fix: link claude/templates/ into ~/.claude/ (setup.sh + CLAUDE.md + ADR-003)

brownm09:
## Summary

- Add `templates` to `setup.sh`'s `~/.claude/` link loop (both `setup_windows()` and `setup_unix()`) — `claude/templates/` has existed since 2026-04-13 (#10) but was never wired into the automated bootstrap, silently breaking `/propose`'s (and `/journal-compose`'s, and `/journal-onboard`'s) required reads of `proposal.md`/`pr-body.md`/`project-claude.md` on any machine.
- Document the fix in `CLAUDE.md` § Dev-Env Architecture's junction table, in `README.md`'s Contents table, and as a new Amendment to [ADR-003](docs/adr/003-config-in-version-control.md) — the second occurrence of that ADR's junction-table-itself-is-wrong gap class (the first: scheduled-tasks topology, 2026-07-01).

## Changes

- `setup.sh` — add `templates` to the `scripts skills hooks` link loop in both platform functions
- `CLAUDE.md` — add `templates/` row to the Dev-Env Architecture junction table
- `README.md` — correct the `claude/templates/` Contents-table row to document the junction, matching its scripts/skills/routines siblings
- `docs/adr/003-config-in-version-control.md` — add `templates/` row to the Decision table, new Amendment (2026-07-08) section (includes a check confirming `usage-config.json` and `setup-prompt.md` are not the same gap), updated References
- `docs/adr/INDEX.md` — update ADR-003's amended-date list and tags

## Test plan

- [x] `bash -n setup.sh` — syntax valid
- [x] `bash claude/scripts/tests/run-shellcheck.sh` — SKIP (shellcheck not installed locally; advisory-only per Testing item 7)
- [x] Created the missing `~/.claude/templates` junction directly on the affected machine (`mklink /D`, same mechanism as `setup.sh`'s `win_link ... dir`) and confirmed `proposal.md` and `pr-body.md` are now readable through it

## Review findings disposition

Per [review](https://github.com/brownm09/dev-env/pull/611#issuecomment-4911915222):

- **[documentation]** (blocking) `README.md:14` stale `claude/templates/` row → fixed in `0a0986e`
- **[documentation]** (non-blocking) ADR-003 amendment overstated "third occurrence" → fixed in `0a0986e`
- **[maintainability]** (non-blocking) no automated test for `setup.sh`'s linking logic → filed [dev-env#614](https://github.com/brownm09/dev-env/issues/614)

Closes #606

[...]

PR #623, opened 2026-07-08
fix(hooks): recognize gh's -R shorthand for --repo across sibling merge hooks

### LINKED ISSUE #616: post-pr-merge-project.py (and likely sibling hooks) don't recognize gh's -R shorthand for --repo
## Incident

Ran `gh pr merge 611 -R brownm09/dev-env --squash` from a session whose cwd was `lifting-logbook` (merging [dev-env#611](https://github.com/brownm09/dev-env/pull/611), closing dev-env#606). The `post-pr-merge-project.py` PostToolUse hook fired and reported: `[project-hook] Issue #608 moved to Done (item PVTI_lAHOAjEKvM4BTuEFzgxGmvM)`.

That item ID belongs to **lifting-logbook's** project board (project #2), not dev-env's (#3) — the hook moved [lifting-logbook#608](https://github.com/brownm09/lifting-logbook/issues/608) ("TS2540: Cannot assign to read-only NODE_ENV..."), an issue with zero relation to my merge.

## Root cause

`extract_repo_from_command()` in `claude/scripts/post-pr-merge-project.py` resolves the merge's target repo via two regexes: `_REPO_FLAG_RE` (`--repo\s+(...)`) and `_PR_URL_REPO_RE` (a full `https://github.com/owner/repo/pull/N` URL). **Neither matches `gh`'s `-R` shorthand** for `--repo`. My command used `-R brownm09/dev-env`, so `extract_repo_from_command` returned `None`, and `main()` fell through to `repo = config.get("repo", "")` — cwd's own `.claude/hook-config.json`, which was lifting-logbook's (session cwd), not dev-env's.

From there: `pr_number = extract_pr_number_from_command(command)` correctly found `611` (a bare number, repo-agnostic), then `get_pr_body(611, repo="brownm09/lifting-logbook")` fetched **lifting-logbook PR #611** — a real, unrelated, already-merged-in-the-past PR (`[fix] TS2540: Replace direct NODE_ENV assignment...`) that coincidentally shares the number 611 with my dev-env PR. That PR's body says `Closes #608` (a real lifting-logbook issue), so the hook moved *that* issue's board item to Done.

## Why this was harmless this time

lifting-logbook#608's board item was already `status: "Done"` (the NODE_ENV fix had already shipped previously) — the hook's write was a redundant no-op. A generally different pairing of coincidental PR numbers and issue board states could have silently reverted real in-progress work on an unrelated repo's board.

## Scope

The docstring in `extract_repo_from_command` says it "mirrors `_effective_merge_repo`'s resolution order in `pr-merge-reminder.py`" (dev-env#470) — that sibling function, and any other hook using the same `_REPO_FLAG_RE`-only pattern (grep `claude/scripts/*.py` for `_REPO_FLAG_RE` or an equivalent `--repo`-only regex), likely has the identical `-R`-blindness gap. Worth auditing all of them in the same fix, not just `post-pr-merge-project.py`.

## Suggested fix

Extend the repo-flag regex to recognize both forms, e.g. `(?:--repo|-R)\s+([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+)`, and add a test case to `claude/scripts/tests/test_post_pr_merge_project.py` (and the equivalent test for any sibling hook fixed in the same PR) pinning that `-R owner/repo` resolves identically to `--repo owner/repo`.

brownm09:
## Summary

`extract_repo_from_command()` in `post-pr-merge-project.py` (and three sibling repo-flag checks) matched only the long `--repo owner/repo` flag via `_REPO_FLAG_RE = re.compile(r"--repo\s+(...)")`, never `gh`'s `-R` shorthand. A `gh pr merge <N> -R owner/repo` command silently fell through to `None`, and callers fell back to cwd's own `.claude/hook-config.json` (or cwd/cd-chain resolution) instead of the command's actual target repo.

**Confirmed incident:** running `gh pr merge 611 -R brownm09/dev-env --squash` from a lifting-logbook cwd (merging dev-env PR #611, closing dev-env#606) made `post-pr-merge-project.py` move **lifting-logbook#608** to Done instead of any dev-env issue — harmless only because that item was already Done.

**Scope — audited beyond the two hooks named in the issue.** Grepped every `--repo`-flag check in `claude/scripts/*.py`, not just files with a `_REPO_FLAG_RE`-named constant (one gap site, `post-pr-merge-pull.py`, is an unnamed inline `re.search()` that a name-scoped grep would have missed). Four hooks needed the fix; two already handled `-R` correctly and needed no change:

| File | Function | Status |
|---|---|---|
| `post-pr-merge-project.py` | `extract_repo_from_command()` | **Fixed** — the filed incident |
| `pr-merge-reminder.py` | `_effective_merge_repo()` | **Fixed** — named in the issue as a likely sibling |
| `posttooluse-inert-advisory.py` | `_devenv_merge_pr()` | **Fixed** — found via audit |
| `post-pr-merge-pull.py` | `extract_repo()` | **Fixed** — found via audit (inline regex, no named constant) |
| `post-tool-use.py` | `extract_repo_flag()` | Already correct (dev-env#542) |
| `stop-tile-enumeration-gate.py` | (own `_REPO_FLAG_RE`) | Already correct |

Each fix extends the existing regex to `(?:--repo|-R)` in the same shape the file already used — deliberately not porting `post-tool-use.py`'s richer `=`-form/URL-form normalization into the other four, since that's unrelated to the reported `-R`-recognition gap.

Also adds **ADR-050 Amendment 14** documenting the fix (this repo's established home for this exact "sibling hook fix" pattern — 13 prior amendments), and brings the ADR's own drifted header (`Amended:` date list, tag list) back in sync with `docs/adr/INDEX.md`, which had already accumulated two more amendments' worth of dates/tags than the ADR file's own header reflected.

**Review round:** `/review` found one non-blocking finding (unanchored `-R` could mis-match mid-word) — fixed in-PR with a standalone-token lookbehind; the residual quote-aware gap is filed as dev-env#626 rather than expanded here. See disposition below.

Closes #616

## Testing

Ran the required test suites for all four fixed files, the repo-wide hook-script syntax check, and the repo-wide crude-command-substring-check regression gate (final counts, after the review-response commit):

```
Tests: 28 passed, 0 skipped, 0 failed

Tests: 50 passed, 0 skipped, 0 failed

Tests: 31 passed, 0 skipped, 0 failed

Tests: 16 passed, 0 skipped, 0 failed

SYNTAX OK (no warnings, verified with -W error)

Tests: 13 passed, 0 skipped, 0 failed (0 unexpected, 0 stale, 0 duplicated)
```

Each fixed file's suite gained one `-R`-shorthand regression case plus one mid-word non-match regression case (pinned against the review finding), all pinned against concrete repro shapes. Suppression / test-integrity pre-PR greps (`git diff origin/main`) both returned no matches, re-run clean after the review-response commit too.

- **[correctness] Unanchored `-R` could mis-match mid-word** — fixed in `c9e6db8` (standalone-token lookbehind across all four files, `_AUTO_FLAG_RE`-pattern precedent, + mid-word regression test per file). Residual quote-aware gap (not newly introduced by this PR) filed as [dev-env#626](https://github.com/brownm09/dev-env/issues/626), Impact=Low.


brownm09:
## Review: fix(hooks): recognize gh's -R shorthand for --repo across sibling merge hooks

### Summary

Merge with minor changes. The fix is correct, well-tested (122 passing tests across the four touched files, all new cases pinned against the exact reported repro), and the audit genuinely goes beyond the filed issue — it found and fixed a fourth gap site (`post-pr-merge-pull.py`) that a name-scoped grep for `_REPO_FLAG_RE` would have missed, and correctly identified two siblings that already handled `-R` and needed no change. The one finding below is a narrow, low-likelihood, self-limiting edge case worth a follow-up, not a blocker.

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/post-pr-merge-project.py:63`, `claude/scripts/pr-merge-reminder.py:157`, `claude/scripts/posttooluse-inert-advisory.py:93`, `claude/scripts/post-pr-merge-pull.py:75`
All four fixed regexes match `-R` as an unanchored substring within the merge command's argument text (`(?:--repo|-R)\s+(...)` / `(?:--repo|-R)[=\s]+(...)`), the same way the pre-existing `--repo` branch already did. Extending the alternation to `-R` measurably widens the false-positive surface, though: `--repo` is a distinctive 6-character literal unlikely to appear by coincidence inside an unrelated `--subject`/`--body` value, but `-R` is a 2-character token that's plausible to hit accidentally (e.g. `gh pr merge 42 --subject "see -R other/thing for context"` would now misparse the subject text as an explicit repo flag). The blast radius differs by file: in `post-pr-merge-project.py` and `posttooluse-inert-advisory.py` a false match degrades to a safe "skip / no advisory" no-op (both already fail closed on a repo mismatch), but in `post-pr-merge-pull.py` — whose `extract_repo()` searches the *entire* raw command, not just the merge invocation's own argument span like its three siblings — a coincidental match could redirect the post-merge local fast-forward to fetch the wrong repo's clone.
**Suggestion:** Anchor the flag match to a standalone token, mirroring the `_AUTO_FLAG_RE = re.compile(r"(?<!\S)--auto(?:=\S+)?(?=\s|$)")` lookaround pattern already used in the same file (`posttooluse-inert-advisory.py:91`) for the identical class of problem — e.g. `(?<!\S)(?:--repo|-R)(?=\s)`. Reasonable to defer as a follow-up issue given the low likelihood and self-limiting failure mode in 3 of the 4 files; if deferred, note it in the PR body per ADR-028's fix-or-file rule.

---

### Findings Disposition

Per ADR-028, record this finding's disposition (fixed in `<sha>` | filed `#N`) in a "Review findings disposition" section of the PR body before merge.

---

Issue #606, opened 2026-07-06
~/.claude/templates/ never linked — setup.sh and CLAUDE.md junction map omit claude/templates/


brownm09:
## Problem

`~/.claude/templates/` does not exist on this machine. Confirmed via `ls ~/.claude/templates/` → "No such file or directory". This silently breaks the `/propose` skill's Step 3 (read `~/.claude/templates/proposal.md`) and Step 11 (read `~/.claude/templates/pr-body.md`) — both are required reads. Discovered during a lifting-logbook session on 2026-07-06 running `/propose`; worked around by using an existing proposal doc from the target repo as a structural reference instead of the documented template.

## Root cause

`claude/templates/` (containing `proposal.md`, `pr-body.md`, `contributing.md`, `project-claude.md`, `propose-config.json`) exists and is fully committed in this repo — first added 2026-04-13 (#10). But it was **never wired into the `~/.claude/` symlink/junction map**:

- `setup.sh`'s `setup_windows()` links `CLAUDE.md` and `settings.json`, loops `scripts skills hooks` as directory links, and separately links `routines` — `templates` is absent from every one of these lists. Identical gap in `setup_unix()`. Confirmed via `git log -p --follow -- setup.sh | grep -i templates` → no hits, ever.
- `CLAUDE.md` § Dev-Env Architecture's documented junction table (the "owned by dev-env, symlinked" table) omits `templates/` too.
- [ADR-003](docs/adr/003-config-in-version-control.md)'s decision table has the identical omission, unchanged across both of its prior amendments (2026-07-01, 2026-07-06).

`setup.sh`'s cross-platform bootstrap was added 2026-04-24 — 11 days after `claude/templates/` already existed — and simply never enumerated it. Not a regression; a day-one gap that every subsequent edit to `setup.sh`, `CLAUDE.md`, and ADR-003 has carried forward unnoticed. This is the third occurrence of ADR-003's "documented map doesn't match reality" gap class (see its 2026-07-01 and 2026-07-06 amendments for the first two, in scheduled-tasks topology and the routine self-reference convention).

## Impact

Every `/propose` invocation on this machine has been silently unable to read its own documented template files, forcing ad-hoc workarounds instead of the skill's documented behavior. Any other skill that comes to reference `~/.claude/templates/*` would hit the same failure. `lifting-logbook`'s own `docs/proposals/README.md` links to `dev-env/claude/templates/proposal.md` on GitHub, confirming the file is expected to be locally resolvable too.

[...]

PR #624, opened 2026-07-08
test: add smoke test for setup.sh's ~/.claude link-loop enumeration

### LINKED ISSUE #614: No automated test for setup.sh's ~/.claude/ linking logic
## Context

Raised as a non-blocking finding on [PR #611](https://github.com/brownm09/dev-env/pull/611)'s review. Every hook/script in this repo has a dedicated pure-function test under `claude/scripts/tests/` (16+ items in `CLAUDE.md`'s Testing section), but `setup.sh`'s `~/.claude/` linking logic (`win_link` / the per-platform link loops in `setup_windows()` / `setup_unix()`) has never had one. PR #611 extended that untested logic (adding `templates` to the link loop) rather than introducing new risk, but the gap means the next addition to this list will also rely solely on manual verification.

## Suggestion

A lightweight smoke test: a temp `$HOME`, run the link loop (or a refactored-out pure list of expected subdirs), assert each expected `~/.claude/<subdir>` resolves to the right junction/symlink target. Doesn't need to cover the UAC self-elevation path or `win_link`'s actual `mklink` invocation — just the enumeration logic (are `scripts skills hooks templates` all actually linked) the way the review flagged.

brownm09:
## Summary

- `setup.sh` had no automated test for its `~/.claude/` linking logic — the enumeration that missed `templates` in #606 (fixed by #611) had no gate that would catch a similar future omission.
- Extracts `link_claude_windows()` / `link_claude_unix()` out of `setup_windows()` / `setup_unix()`, and shares one `CLAUDE_FILE_LINKS` / `CLAUDE_DIR_LINKS` array pair between them (previously each OS branch listed the same names separately — exactly the kind of divergence that let `templates` go missing from one spot without the other catching it).
- Adds `claude/scripts/tests/test-setup-link-loop.sh`: sources `setup.sh` (a new sourcing guard around the dispatch block makes this safe) with `win_link`/`ln` stubbed, and pins both the shared arrays and the extracted functions' exact call sequence against a throwaway `$HOME`. No UAC elevation, no real `~/.claude` or global git config touched.
- **Incidental fix, same gap class:** `claude/setup-prompt.md`'s Step 3 verification checklist independently enumerates the same linked set and was missing *both* `templates/` and `routines/` — a fourth occurrence of the gap ADR-003 tracks, missed by that ADR's own "checked the remaining entries" amendment (which looked for a *programmatic* dependency and missed this *manual* checklist). Added an addendum to ADR-003 correcting that claim and noting the new test.
- Also updated `README.md`'s "Adding new configs" guidance to point at the new array-based mechanism instead of the old "add a `ln -sf`/`mklink` line" instruction.

## Why this scope

Discovered the `setup-prompt.md` gap while researching ADR-003's history to decide where this test's rationale belongs (not a new ADR — ADR-003 already documents and amends this exact linking-enumeration gap class three times). Same root cause, same investigation, small fix — included here per the "fix errors on encounter" rule rather than filing separately.

## Testing

```
bash claude/scripts/tests/test-setup-link-loop.sh
bash claude/scripts/tests/check-script-path-hygiene.sh
bash claude/scripts/tests/run-shellcheck.sh
bash -n setup.sh
```

Tests: 10 passed, 0 skipped, 0 failed (test-setup-link-loop.sh). Path-hygiene: clean (15 shell files scanned). Shellcheck: SKIP (not installed locally, per the documented fallback in item 7 of the Testing section). `bash -n setup.sh`: syntax OK.

Also manually confirmed the new test fails (3 of 10 checks) when `templates` is removed from `CLAUDE_DIR_LINKS` on a throwaway copy — regression coverage verified before relying on it.

No other dev-env script/doc references `setup_windows`/`setup_unix`/`win_link`'s old internal structure (grepped repo-wide).

`/review` posted [here](https://github.com/brownm09/dev-env/pull/624#issuecomment-4916275544) — 0 blocking, 2 non-blocking. Both fixed in `4e3a165`:

- **[documentation]** `test-setup-link-loop.sh` missing from `docs/REFERENCE.md`'s Script verification suite table → added a row mirroring `test-merge-stale-pr.sh`'s entry.
- **[maintainability]** Shell test file outside the JS/TS-oriented Test Integrity scanners' language scope → no code change needed; manually verified no skip/degradation mechanism exists in the new file.

Closes #614


brownm09:
## Review: test: add smoke test for setup.sh's ~/.claude link-loop enumeration

### Summary
Ready to merge. This is a well-scoped, behavior-preserving refactor (verified: `bash -n` clean, the new test's own regression check confirms it fails when `templates` is dropped from the array) plus a genuinely new test, with two small incidental doc fixes from the same investigation. No blocking findings. At 285 changed lines across 7 files, the PR is appropriately scoped for a single coherent change — no need to split.

---

### Non-Blocking Findings

**[documentation]** `docs/REFERENCE.md` — Utilities → Script verification suite table
The new `claude/scripts/tests/test-setup-link-loop.sh` isn't listed in this table, even though its closest sibling in spirit — `test-merge-stale-pr.sh` (also "drives the real script against a throwaway fixture with the side-effecting primitive stubbed") — has an entry there. Precedent is genuinely mixed (only 5 of 49 test files in `claude/scripts/tests/` appear in this table, so this isn't a hard gate), but since the new test is structurally closest to the one entry that *is* documented there, adding a row keeps the table meaningful rather than arbitrary.
**Suggestion:** Add a row to the Script verification suite table mirroring the `test-merge-stale-pr.sh` entry's style.

**[maintainability]** `claude/scripts/tests/test-setup-link-loop.sh` — language-scope coverage gap
This diff's only test-relevant file is a Bash script. The repo's automated Test Integrity patterns (skip markers, deleted tests, lowered thresholds) target JS/TS frameworks and can't inspect shell test files. A clean result from those scanners isn't evidence this file is free of integrity issues.
**Suggestion:** No action needed here — manually verified: no skip/degradation mechanism exists in this file (all three scenarios always run unconditionally), no tests were deleted, no thresholds changed. Noting this for the record per the language-scope check rather than as an actionable gap.

---

### Style Notes
Clean. Comments consistently use `--` instead of em-dash inside shell comments (avoids encoding ambiguity in `.sh` files, unlike prose in `.md` files where em-dash is this repo's convention) — good attention to the split. Function naming (`link_claude_windows` / `link_claude_unix`) matches the existing `setup_windows` / `setup_unix` / `set_hooks_path` convention.

---

### Findings Disposition
Both non-blocking findings will be fixed in this PR: adding the `docs/REFERENCE.md` table row (cheap, directly analogous to existing precedent), and the language-scope note requires no code change, just the manual-verification record above.

---

PR #635, opened 2026-07-08
[fix] Mask quoted spans before repo-flag regex checks

### LINKED ISSUE #626: Repo-flag checks (_REPO_FLAG_RE family) are unanchored substring searches, not quote-aware
## Context

While fixing #616 (gh's `-R` shorthand for `--repo` not recognized), `/review` on PR #623 flagged that all four fixed repo-flag checks (`post-pr-merge-project.py`, `pr-merge-reminder.py`, `posttooluse-inert-advisory.py`, `post-pr-merge-pull.py`) run `_REPO_FLAG_RE.search()` (or an inline equivalent) over the raw merge-invocation argument text with no quote-awareness.

PR #623 applied a cheap, proportionate mitigation: a `(?<!\S)` standalone-token lookbehind, preventing a mid-word match like `add-R other/repo` (e.g. inside a PR title literally containing "add-R flag support"). That lookbehind does **not** address the more realistic residual case: `-R`/`--repo` appearing as a legitimately space-separated token *inside a quoted `--subject`/`--body` value*, e.g.:

```
gh pr merge 42 --subject "see -R other/repo for context"
```

Here `-R other/repo` is preceded by a space (from `"see -R"`), so the lookbehind doesn't exclude it — the regex has no way to know it's inside a quoted string rather than a top-level flag.

## Why this is a separate issue, not part of #616

A true fix requires quote-aware tokenization of the merge invocation's argument span (masking or skipping quoted substrings before running `_REPO_FLAG_RE`) — a materially larger change than a regex tweak, and one that would need to be applied consistently across all four (arguably five, including `post-tool-use.py`'s `extract_repo_flag`) sibling repo-flag checks. This is exactly the kind of "richer parsing" scope PR #623 deliberately declined to port from `post-tool-use.py`, since it's unrelated to the `-R`-recognition gap #616 reported.

**Important:** this is a **pre-existing** limitation, not newly introduced by #623 — the original `--repo`-only regex was *already* an unanchored raw-text search with the identical quoted-value blind spot. #623 only marginally widens the practical exposure by adding a second, shorter trigger literal (`-R`, 2 chars) alongside the original `--repo` (6 chars), which is somewhat more likely to appear by coincidence.

## Suggested fix

Investigate reusing (or extending) this repo's existing quote-aware `split_top_level`/`scan_top_level` engine (`_hookio.py`, ADR-050 Amendments 5/7) — built for statement-level splitting, but the same quote/heredoc/subshell-tracking core could plausibly mask quoted substrings within a single segment before a repo-flag regex runs over it. Scope to all repo-flag check sites in one PR, per this ADR's established "grep for the shape, not just the named instances" convention (ADR-050 Amendment 14's own general lesson).

## Severity

Low likelihood (requires a `-R`/`--repo`-shaped substring to appear coincidentally inside quoted flag-value prose), low-to-moderate impact (self-limiting "skip" no-op in most of the affected hooks; `post-pr-merge-pull.py`'s unscoped whole-command search is the most exposed, where a false match could redirect a post-merge local fetch to the wrong repo's clone).

brownm09:
## Summary

Fixes the residual quote-awareness gap in the `_REPO_FLAG_RE` family that PR #623 (Amendment 14) deliberately left as a follow-up. Amendment 14's `(?<!\S)` lookbehind stops a mid-word `-R` match (e.g. `xx-R`) but can't distinguish "whitespace inside a quoted value" from "whitespace between top-level tokens" — a `--subject`/`--body` value like `"see -R other/repo for context"` still contains a legitimately space-separated `-R other/repo` substring that falsely matches.

**Fix:** a new `mask_quoted_spans(command)` helper in `_hookio.py` — an independent stack-based walker (mirrors `split_top_level`'s state transitions, reuses the already-shared `_find_heredoc_end`) that blinds single/double-quoted spans, `$()` subshells, and heredoc bodies to `#` before a repo-flag regex ever runs against the text. Wired into all four sibling checks:

| File | Function | Masking scope |
|---|---|---|
| `post-pr-merge-project.py` | `extract_repo_from_command` | masks the `_MERGE_ARGS_RE`-scoped `args`, only for the `_REPO_FLAG_RE` check — the PR-URL fallback stays unmasked (a quoted PR URL is a legitimate, already-tested shape) |
| `pr-merge-reminder.py` | `_effective_merge_repo` | masks the whole `command` (no args-prescoping exists here) |
| `posttooluse-inert-advisory.py` | `_devenv_merge_pr` | masks `args`, only for the `_REPO_FLAG_RE` check — the PR-URL match (`url_m`) stays unmasked since it's reused for the PR-number fallback |
| `post-pr-merge-pull.py` | `extract_repo` | masks the whole `command`, only for the first (repo-flag) search — the second (URL) search stays unmasked. This file's check was the most exposed of the four (no args-scoping, no named regex constant at all) |

**Deliberately independent of `split_top_level`, not a refactor of it.** `split_top_level` has ~30 existing tests and several callers, hardened across Amendments 5 and 7 — Amendment 7's own postmortem ("`/review`'s adversarial pass, not the hand-written test suite, is what caught this") is a direct caution against generalizing its internals for an unrelated need. `mask_quoted_spans` mirrors the same state rules and reuses `_find_heredoc_end`, but `split_top_level` itself has zero lines changed. Because the two walks are independent, `test_hookio.py` gains an *enforced* cross-consistency test (`test_mask_quoted_spans_agrees_with_split_top_level`) rather than a prose "keep in sync" comment — direct precedent from Amendment 11's own lesson.

**Excluded from this PR:** `post-tool-use.py`'s `extract_repo_flag` — dev-env#626 flagged it as "arguably" a fifth site, but its regex already supports a legitimately *quoted* flag value as part of its own capture group (a shape none of the other four ever supported), so blanket masking would break that legitimate case rather than just fix the false-match one. Its docstring already discloses the narrower residual gap as an accepted, conservative-by-design tradeoff, further bounded by its caller requiring independent confirmation from a real sibling checkout's own config. See ADR-050 Amendment 15 for the full rationale.

**Also excluded, filed as a follow-up:** auditing the family surfaced three more related-but-distinct instances (a naive `.split()`-based tokenizer with the identical hijack in `pre-merge-findings-gate.py`, which also transitively covers `pre-auto-merge-checkpoint-gate.py` via a dynamic import of the same function; `stop-tile-enumeration-gate.py`'s own `_REPO_FLAG_RE`, missing *both* the Amendment 14 lookbehind and quote-awareness; and the PR-URL-regex analog in all four files fixed here). None are named in #626 itself and each needs its own fix shape — filed as [#634](https://github.com/brownm09/dev-env/issues/634) rather than expanded here, per this ADR's own "Follow-up, not fixed here" precedent (Amendment 11 → 12).

Closes #626

## Testing

Ran the required test suites for every touched file, the repo-wide hook-script syntax check, and the repo-wide crude-command-substring-check regression gate:

```
SYNTAX OK

Tests: 70 passed, 0 skipped, 0 failed   (57 pre-existing + 13 new: 12 direct mask_quoted_spans tests + 1 cross-consistency test)

Tests: 30 passed, 0 skipped, 0 failed   (28 pre-existing + 2 new)

Tests: 52 passed, 0 skipped, 0 failed   (50 pre-existing + 2 new)

Tests: 31 passed, 0 skipped, 0 failed   (unchanged count -- 2 new cases added as assertions inside the existing test_devenv_merge_pr_direct, matching this file's own bundled-assertions convention)

Tests: 18 passed, 0 skipped, 0 failed   (16 pre-existing + 2 new)

```

Every new test is pinned against the exact dev-env#626 repro shape (`gh pr merge 42 --subject "see -R other/repo for context"`, adapted per file) plus a "real flag survives alongside an irrelevant quoted decoy" case. All five existing suites re-run in full and pass unchanged — in particular the existing quoted-PR-URL (`test_repo_from_cross_repo_url`) and `--repo`-flag-precedence-over-quoted-URL tests confirm the masking-scope decision (never mask a fallback regex whose match is reused elsewhere) is correct, not just intended.

Suppression / test-integrity pre-PR greps (`git diff origin/main`) both returned no matches.

- **[maintainability] Imprecise `#`-placeholder docstring claim** — fixed in `267d3c5`.
- **[maintainability] Unenforced masking-scope invariant** — fixed in `267d3c5` (module-comment note documenting the invariant and why a unified mask+search helper was considered and rejected for this PR).


brownm09:
## Review: [fix] Mask quoted spans before repo-flag regex checks

### Summary
Ready to merge. This is a well-scoped, carefully-reasoned fix: the new `mask_quoted_spans` walker in `_hookio.py` is independently verified correct (traced against `split_top_level`'s real state transitions, empirically tested against the primary target case plus heredoc/subshell/unterminated/glued-quote adversarial inputs by an independent reviewer), the masking-scope decision per call site was checked against every existing test that depends on the "don't mask the URL fallback" behavior, and two candidate "correctness regressions" raised during review (a quoted `--repo "owner/repo"` value, and backtick command substitution) were independently verified to be pre-existing, unchanged behavior — not introduced by this diff. Two non-blocking maintainability findings surfaced (an imprecise docstring claim, and an unenforced masking-scope invariant); both are fixed in this PR. No blocking findings. Diff is 626 lines across 12 files, appropriately scoped to the one bug (three related-but-distinct discoveries were correctly deferred to a separate follow-up issue rather than bundled in).

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/_hookio.py` (mask_quoted_spans docstring)
The docstring justified the `#` placeholder choice by claiming it "appears in none of this file's repo-flag/PR-number/URL regexes' character classes" — imprecise on two counts: those regexes don't live in `_hookio.py` (they live in the four caller files), and one of them (`posttooluse-inert-advisory.py`'s `_REPO_FLAG_RE = (?<!\S)(?:--repo|-R)[=\s]+(\S+)`) captures via `\S+`, which `#` *does* satisfy. The actual safety property still holds (the anchor literal `--repo`/`-R` can't be synthesized from `#` characters, so a masked span can only flip a match from present-but-wrong-value to failing the caller's equality check — never absent-to-falsely-present), but a future maintainer relying on the stated invariant when wiring a fifth caller would be relying on an inaccurate premise.
**Fix:** Corrected in `_hookio.py` — the docstring now names the actual caller regexes, calls out the looser `\S+` capture explicitly, and states the real (narrower but still sound) safety property instead of the blanket claim.

**[maintainability]** `claude/scripts/_hookio.py` (module comment) / all four call sites
The correctness of the whole fix rests on an invariant enforced only by per-site prose comments: mask *only* the string fed to the vulnerable repo-flag regex, never a fallback (URL/PR-number) regex whose match is reused elsewhere. It's hand-wired at four sites with two different masking scopes (an `args` sub-region vs. the whole `command`), with nothing structural stopping a future edit from masking the wrong string — and `_hookio.py`'s own module docstring records this exact hook family already drifting this way once before (dev-env#380, the wrong-field-read bug independently reintroduced across four sibling hooks).
**Fix:** Added a module-comment note in `_hookio.py` making the invariant and the considered-and-rejected shared-helper alternative explicit (a unified "mask+search" helper would need immediate parameterization over per-site scope differences no current caller asks for — the same premature-convergence risk ADR-050 Amendment 5's scope decision already warns against), and pointing to the existing `*_survives_alongside_quoted_decoy` regression test at each of the four call sites as the enforced (not just prose) check that the hand-wiring is correct today.

---

### Findings Disposition

- **[maintainability] Imprecise `#`-placeholder docstring claim** — fixed in `_hookio.py` (commit follows this review).
- **[maintainability] Unenforced masking-scope invariant** — fixed in `_hookio.py` module comment (commit follows this review); structural alternative (shared mask+search helper) considered and explicitly documented as rejected for this PR, not silently dropped.

---

PR #647, opened 2026-07-09
[fix] Mask quoted spans/prose-flag values at the three remaining repo-flag/PR-URL sites

brownm09:
## Summary

Extends dev-env#626's quote-awareness fix (ADR-050 Amendment 15) to the three sites its own audit surfaced but did not fix, per dev-env#634:

1. **`pre-merge-findings-gate.py::_parse_merge_target`** — tokenized the `gh pr merge` tail via naive `tail.split()`, not a regex at all. A `--subject`/`--body` value like `"see -R other/repo for context"` splits into several whitespace tokens (no quote-awareness), so the decoy `-R`/`other/repo` inside it was indistinguishable from a real `--repo`/`-R` flag + value — confirmed live repro: `_parse_merge_target('gh pr merge 42 --subject "see -R other/repo for context"')` returned `('42', 'other/repo')` instead of `('42', None)`. **Fix:** `tokens = mask_quoted_spans(tail).split()` — masking collapses the whole decoy into one contiguous token, consumed whole by the existing `_VALUE_FLAGS` handling.
2. **`pre-auto-merge-checkpoint-gate.py`** — no code change; dynamically imports `_parse_merge_target` from file 1 (same function object), so site 1's fix transitively covers it. Re-ran its full test suite to confirm.
3. **`stop-tile-enumeration-gate.py`'s own `_REPO_FLAG_RE`** — never received PR #623's `(?<!\S)` lookbehind (it already recognized `-R` before that PR, so it fell outside that audit) or Amendment 15's masking — a strictly larger, pre-existing gap than dev-env#626 itself. **Fix:** added the lookbehind, and `_explicit_repo` now runs the flag check against a `mask_quoted_spans`-masked copy of its input (the `_PR_URL_RE` fallback stays unmasked, matching the established four-site pattern — and is explicitly out of dev-env#634's own scope).
4. **The PR-URL-regex analog** in all four Amendment 15 files (`_PR_URL_REPO_RE` / `_DEVENV_PR_URL_RE` / inline URL searches) — identical quoted-value blind spot, but for a URL-shaped decoy instead of a flag-shaped one (e.g. `--subject "see https://github.com/other/repo/pull/1 for context"`). **Fix:** new `mask_prose_flag_values()` helper in `_hookio.py` that masks *only* the value immediately following a `--subject`/`-t`/`--body`/`-b` flag (quoted **and** unquoted — see Review findings disposition below) — unlike blanket `mask_quoted_spans`, this preserves the legitimate bare-quoted-PR-URL-argument case (`gh pr merge "https://github.com/o/r/pull/N"`, `test_repo_from_cross_repo_url`) while still closing the decoy. Applied in `post-pr-merge-project.py`, `posttooluse-inert-advisory.py`, `post-pr-merge-pull.py`. `pr-merge-reminder.py` needs no change — its analogous function has no URL-fallback branch at all.

**Design note:** `mask_quoted_spans` is refactored to share a new private `_opaque_spans()` span-finder with `mask_prose_flag_values` — a byte-identical-output-preserving extraction (verified: all 13 pre-existing `mask_quoted_spans` tests pass unmodified), not a reversal of Amendment 15's "no premature parameterization" call (that was about the masking *policy*, not the underlying span-finding mechanism).

**Out of scope, not fixed here** (see ADR-050 Amendment 17 for full detail):
- `stop-tile-enumeration-gate.py`'s own `_PR_URL_RE` — dev-env#634's point 4 scoped the URL-regex analog fix to the four Amendment 15 files specifically, not this file's independent pair.
- A separate, more widespread characteristic noticed while implementing site 1: several files scope their "args" region via a raw split/negated-character-class match with no quote-awareness of its own (`pre-merge-findings-gate.py`'s `re.split(r"&&|\|\||;|\n", tail)[0]`, the `_MERGE_ARGS_RE` pattern) — a `--subject`/`--body` value containing a literal shell-separator character could truncate the scoped region early, potentially *hiding* a real trailing flag/URL rather than exposing a decoy (the opposite failure direction). Not demonstrated as live-exploitable, pre-existing across the whole family — flagged as a follow-up, not fixed here. Not yet filed as an issue.
- A **third decoy class** found during `/review` (see disposition below): a bare positional PR-number inside `--subject`/`--body` prose, unmasked at three sites. Filed as [dev-env#650](https://github.com/brownm09/dev-env/issues/650).

Closes #634

## Testing

Ran the required test suites for every touched file, the repo-wide hook-script syntax check, and the repo-wide crude-command-substring-check regression gate:

```
SYNTAX OK

Tests: 86 passed, 0 skipped, 0 failed   (74 pre-existing + 12 new mask_prose_flag_values tests, incl. 3 added post-review for the unquoted-value fix)

py -3 claude/scripts/tests/test_pre_merge_findings_gate.py
Tests: 8 passed, 0 skipped, 0 failed   (unchanged -- _parse_merge_target coverage lives in the shell test below, per this file's own convention)

bash claude/scripts/tests/test-merge-findings-gate.sh
Results: 11 passed, 0 failed   (9 pre-existing + step 8's 2 new dev-env#634 cases)

py -3 claude/scripts/tests/test_pre_auto_merge_checkpoint_gate.py
Tests: 30 passed, 0 skipped, 0 failed   (unchanged -- re-run to confirm transitive fix)

bash claude/scripts/tests/test-auto-merge-checkpoint-gate.sh
Results: 15 passed, 0 failed   (unchanged -- re-run to confirm transitive fix)

py -3 claude/scripts/tests/test_stop_tile_enumeration_gate.py
Tests: 81 passed, 0 skipped, 0 failed   (76 pre-existing + 5 new _explicit_repo direct tests)

Tests: 32 passed, 0 skipped, 0 failed   (30 pre-existing + 2 new)

Tests: 20 passed, 0 skipped, 0 failed   (18 pre-existing + 2 new)

Tests: 31 passed, 0 skipped, 0 failed   (unchanged count -- 3 new dev-env#634 cases added as assertions inside the existing test_devenv_merge_pr_direct, matching this file's own bundled-assertions convention)

Tests: 52 passed, 0 skipped, 0 failed   (unchanged -- sanity check; no code touched in this file)

```

Every new test is pinned against the exact dev-env#634 repro shapes (per-file adaptation of `gh pr merge 42 --subject "see -R other/repo for context"` and its URL-decoy analog) plus a "real signal survives alongside an irrelevant quoted decoy" case. All previously-passing suites re-run in full and pass unchanged — in particular the existing quoted-PR-URL tests (`test_repo_from_cross_repo_url` and siblings) confirm the `mask_prose_flag_values` scoping decision (mask only prose-flag values, never a bare positional URL argument) is correct, not just intended.

Suppression / test-integrity pre-PR greps (`git diff origin/main -- . | grep '^+'` filtered to added lines only) both returned no matches — no new suppressions or test-integrity violations introduced.

`/review` ran two parallel adversarial subagents (correctness/security; reliability/performance/maintainability). Four of the reliability subagent's findings were false positives — traced to an incomplete prompt (it was fed only the source-file diff, not the test-file diffs) — independently verified via grep that the claimed-missing tests all exist, and discarded those findings rather than act on them.

- **[correctness] blocking — unquoted-value masking gap in `mask_prose_flag_values`** — fixed in `5088c90`. An unquoted `--subject`/`--body`/`-t`/`-b` value could itself be a decoy with no surrounding prose needed (`--body https://github.com/other/repo/pull/1`, no quotes). Confirmed live before the fix; re-verified after. Docstring, module comment, and ADR-050 Amendment 17 all updated; 3 new tests added (net, after replacing 1 now-incorrect test), full suite re-run clean.
- **[maintainability] non-blocking — stale `mask_quoted_spans` docstring clause** — fixed in `5088c90` (same commit), a leftover contradiction from the `_opaque_spans` extraction.
- **[maintainability] non-blocking — test-integrity language-scope note** — no fix needed; the JS/TS-oriented automated scanners don't natively apply to this Python/shell diff, so manually re-verified clean (no skip markers, no deleted tests, no lowered thresholds in the added lines).
- **Bare positional PR-number decoy (3 sites)** — filed as [dev-env#650](https://github.com/brownm09/dev-env/issues/650) (Impact: Medium), out of scope for this PR — a third, distinct decoy shape (neither a repo-flag nor a PR-URL regex) affecting a comparable number of sites to redo as its own investigation.


brownm09:
## Review: [fix] Mask quoted spans/prose-flag values at the three remaining repo-flag/PR-URL sites

### Summary

Solid, well-tested fix — merge with the one blocking issue already fixed below. The diff closes dev-env#634's three sites cleanly and its own test suite is thorough (confirmed by direct execution, not just reading). Review turned up one real gap in the PR's own new code (an unquoted-value masking blind spot in `mask_prose_flag_values`) that has already been fixed in this same PR as part of this review pass, plus one pre-existing, out-of-scope gap (a bare-number decoy) now filed separately as dev-env#650. At 725 diff lines this PR is appropriately scoped — it is exactly the three sites dev-env#634 named, not more.

---

### Blocking Findings

**[correctness]** `claude/scripts/_hookio.py` — `mask_prose_flag_values` (pre-fix)
The original implementation masked a `--subject`/`-t`/`--body`/`-b` flag's value only when that value began with a quote/`$(`/heredoc-opener (i.e., only when `_opaque_spans` had an entry starting exactly at the flag's value position). An **unquoted** single-token value was left completely unmasked, on the reasoning that "no internal whitespace means no decoy surrounded by prose." That reasoning missed the case where the decoy doesn't need surrounding prose at all — the flag's entire unquoted value can itself be the decoy.
Confirmed live before the fix: `extract_repo_from_command('gh pr merge 380 --body https://github.com/other/repo/pull/1 --squash')` returned `'other/repo'` instead of `None` — a `gh pr merge 380` (this repo's own PR) with an unquoted `--body` referencing an unrelated PR in another repo would have been misattributed to that other repo, exactly the class of bug this whole PR exists to close, just reached via an unquoted value instead of a quoted one.
**Fix:** Already applied in this PR — `mask_prose_flag_values` now also masks an unquoted value (walking from the flag's value position to the next whitespace or end of string) when no `_opaque_spans` entry starts there. Re-verified: `extract_repo_from_command(...)` on the repro above now returns `None`. Docstring and module comment updated to describe the corrected behavior; ADR-050 Amendment 17 updated with the finding and fix. Three new tests added to `test_hookio.py` (unquoted value masked, the concrete unquoted-URL-decoy repro, and an unquoted value at end-of-string with no crash); the pre-existing test asserting the old ("unquoted not masked") behavior was updated to assert the corrected behavior. Full suite re-run clean (`test_hookio.py`: 86 passed, up from 84; all seven other affected suites unchanged).

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/_hookio.py` — `mask_quoted_spans` docstring (pre-fix)
The docstring's closing paragraph read "...opacity rules **this function's state machine** mirrors (kept as an independent walk...)" immediately followed by a newly-added sentence noting the state machine now lives in `_opaque_spans`. The two clauses contradicted each other — `mask_quoted_spans` no longer contains a state machine of its own after the `_opaque_spans` extraction, so "this function's state machine" was stale within the same docstring that describes the extraction.
**Fix:** Already applied in this PR — reworded to attribute the walk to `_opaque_spans` throughout, removing the internal contradiction.

**[maintainability]** Test-integrity gate — language scope
Per the Test Integrity Gate's language-scope preamble: every file in this diff is Python (`.py`) or shell (`.sh`), so the JS/TS-oriented automated skip-marker/threshold scanners in this review process don't natively apply. Manually verified as part of this review: `git diff origin/main -- . | grep '^+'` (added lines only) for `pytest.mark.skip`/`pytest.skip(`/`unittest.skip`/shell-equivalent patterns returned no matches; no test files were deleted; no `describe`/`it`-block-equivalent test functions were removed from any file (confirmed via `git diff --diff-filter=D` and a check for removed `def test_` lines). No action needed — noting for the record since the automated scanners can't attest to this themselves for a Python/shell diff.

---

### Filed Separately (pre-existing, out of scope for this PR)

**dev-env#650** — A third decoy class found during this review (by a subagent that executed the code, not just read it): a **bare positional PR-number** inside `--subject`/`--body` prose (e.g. `--subject "resolves 42 items"`) is read as a real PR number by three sites — `posttooluse-inert-advisory.py::_devenv_merge_pr`, `post-pr-merge-project.py::extract_pr_number_from_command`, `stop-tile-enumeration-gate.py::_target_pr` — none of which are touched by this PR's masking (neither `mask_quoted_spans` nor `mask_prose_flag_values` were designed to protect bare-number extraction). Confirmed live against all three. This is squarely pre-existing (the affected lines are untouched by this diff) and a distinct decoy shape from what dev-env#634 scoped this PR to fix (repo-flag and PR-URL decoys, not bare numbers) — filed as its own follow-up rather than expanding this PR, consistent with the "grep for the shape, note what's out of scope, file it" discipline this ADR has used since Amendment 9.

---

### Questions for Author

None — the design tradeoffs (why `mask_quoted_spans` can't be reused for the URL-fallback case, why `_opaque_spans` extraction doesn't reopen Amendment 15's "no premature parameterization" call, why `pr-merge-reminder.py` needs no change) are all explained in the PR body and ADR text clearly enough that there's nothing genuinely ambiguous left to ask.

---

### Style Notes

None beyond what's already covered above — the extensive docstrings/inline rationale match this codebase's established convention (every prior ADR-050 amendment does the same), not an outlier for this PR.

---

### Findings Disposition

- **[correctness] Unquoted-value masking gap in `mask_prose_flag_values`** — fixed in this PR (see commit following this review; `_hookio.py`, `test_hookio.py`, and `docs/adr/050-shared-hookio-sibling-hook-fixes.md` all updated).
- **[maintainability] Stale `mask_quoted_spans` docstring clause** — fixed in this PR (same commit).
- **[maintainability] Test-integrity language-scope note** — no fix needed; manually verified clean, noted above.
- **Bare positional PR-number decoy (3 sites)** — filed as [dev-env#650](https://github.com/brownm09/dev-env/issues/650), out of scope for this PR.

---

PR #633, opened 2026-07-08
docs(claude-md): Glob/git status skip gitignored files (#632)

brownm09:
## Summary
- Add item 5 to the CLI Scripting Checklist (claude/CLAUDE.md): before declaring a directory empty/cleared/fully processed, verify with a raw listing or `git status --ignored` — Glob and plain git status/ls-tree silently skip gitignored files.
- Prevents a repeat of the career-playbook incident where a gitignored drop-zone directory was wrongly reported as empty, nearly masking a real batch-pipeline bug (career-playbook #664 / PR #665).

Closes #632. Related: #602 (same checklist section, different gotcha).

ADR-warrant check: not warranted — see docs/adr/011-adr-warrant-check.md. This adds one checklist line; no new rule/hook/skill/settings change, no claude/ directory restructure, no new cross-referenced workflow rule.

Testing: docs-only guard run (`grep -n 'date -u' claude/CLAUDE.md`) — zero matches, unaffected by this change.

## Review findings disposition
- **[correctness]** `git ls-tree --ignored` is not a valid flag; grouping `git ls-tree` with `git status --ignored` implied a nonexistent remediation — fixed in `9483763` (reworded to describe `git ls-tree`'s actual blind spot instead).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #625, opened 2026-07-08
fix(hooks): deliver journal-stop-check archive reminder via exit 2 + stderr

brownm09:
## Problem

`journal-stop-check.py` (a Stop hook) emitted its **archive reminder** — "Archive this session now: call `ccd_session_mgmt__archive_session` … Then stop." — via `print(...)` + `sys.exit(0)`. That reminder is **Claude-facing**: it instructs Claude to invoke an MCP tool only Claude can call, then stop.

Per the [Claude Code hooks reference](https://code.claude.com/docs/en/hooks), for a **Stop** hook, exit-0 **stdout** is *"written to the debug log but not shown in the transcript"* and is **not** added to Claude's context — only `UserPromptSubmit` / `UserPromptExpansion` / `SessionStart` get exit-0 stdout added to context. Only **exit 2 + stderr** feeds text back to Claude on a Stop hook. So the archive reminder never reached Claude and the intended session-archiving silently never happened.

This is the exact failure class [ADR-088](../blob/main/docs/adr/088-state-keyed-tile-enumeration-gate.md) identified and fixed for the tile gate ("an exit-0 advisory … Would be ignored — the exact mechanism-2 failure. Rejected."). Surfaced during the #612 / #613 Stop-hook parallelism investigation.

## What changed

- **`journal-stop-check.py`** — the **archive-reminder branch** now emits on **stderr + exit 2** (blocking the stop so Claude receives it), mirroring `stop-tile-enumeration-gate.py`. Extracted `archive_reminder_message()` (pure, ASCII/cp1252-safe) and `parse_stop_hook_active()`; parameterized `consume_stub_pushed_sentinel(sentinel=…)` for testing. **The stale-draft / unmerged-branch / orphan-removed advisories stay non-blocking (exit 0, stdout)** — they're user-facing and point at *later, dedicated-session* work (journal composition must never be triggered proactively; stale-PR merges are separate work).
- **Loop-safe**: the sentinel is consumed on the single block (one-shot), plus the consume is gated on `not stop_hook_active` (the documented anti-loop flag) so a continuation never consumes-without-delivering.
- **`stub-push-archive-reminder.py`** — corrected the "via stdout — the correct output channel" docstring (now stderr + exit 2).
- **ADR-091** (new) + `INDEX.md`; corrected ADR-088's now-stale "first and only blocking Stop hook" / "always exits 0" note.
[...]

PR #659, opened 2026-07-09
fix(hooks): gate stub-push archive reminder on a confirmed-resolved PR

brownm09:
## Summary

- `stub-push-archive-reminder.py` armed the archive-session sentinel on **any** stub push to
  `engineering-journal`, with no check on whether an associated PR was still open. Per this repo's
  own workflow, a stub is pushed immediately after `gh pr create` and again after each subsequent
  push — well before `/review` and `gh pr merge` — so ADR-091's exit-2 delivery fix (merged today
  via #625) made this pre-existing trigger-condition gap newly harmful: the archive instruction now
  actually reaches Claude mid-session, before review/merge, threatening to destroy the very
  worktree that work still needs.
- Adds `head_commit_has_unresolved_pr()` (+ `manifest_path_for_stub()`, `has_unresolved_open_pr()`):
  for each touched `.stub.md` still on disk, reads its **paired manifest's current on-disk content**
  — not gated on whether the triggering commit's own diff touched that manifest, since an ordinary
  mid-session stub-only push doesn't re-touch it — and checks for a `prs_opened` entry not also in
  `prs_closed`. Conservative on every ambiguity (missing/unreadable/unparseable manifest): fails
  toward **not** archiving.
- Confirmed against two real same-day sessions in `engineering-journal` before writing the fix (dev-env
  PR #635 and PR #633 — the latter specifically pins the case where a later stub-only push doesn't
  re-touch the manifest, which a design that only inspected the triggering commit's own diff would miss).
- `journal-stop-check.py` needed no changes — the fix lives entirely on the sentinel-writer side.
- Documentation kept in sync per this repo's convention: ADR-091 Amendment 1, `docs/adr/INDEX.md`,
  `CLAUDE.md` Testing items 16/41, `docs/REFERENCE.md`, `README.md`.

## Test plan

- [x] `py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py` — syntax OK
- [x] `py -3 claude/scripts/tests/test_journal_schema.py` — 29 passed
- [x] `py -3 claude/scripts/tests/test_stub_push_archive_reminder.py` — 31 passed (includes the new
      `head_commit_has_unresolved_pr` cases, including a direct pin of the PR #633 shape, plus 2
      review-driven cases added after the initial push — see disposition below)
- [x] `py -3 claude/scripts/tests/test_journal_stop_check.py` — 12 passed (unchanged file, sanity check)
- [x] `py -3 claude/scripts/tests/test_validate_manifest.py` — 28 passed (other `_journal_schema.py` consumer, sanity check)
- [x] `py -3 claude/scripts/tests/test_journal_shard_write_advisory.py` — 61 passed (other `_journal_schema.py` consumer, sanity check)
- [x] Suppression + test-integrity pre-PR greps: both clean (no matches)

**Tests: 161 passed, 0 skipped, 0 failed**

Per [/review](https://github.com/brownm09/dev-env/pull/659#issuecomment-4927574975) (2 parallel
Opus passes, 0 blocking findings, 3 non-blocking):

1. **[maintainability]** Manifest read bypassed the shared `decode_shard_bytes` BOM-handling
   helper — **fixed in `cd40515`** (+ regression test).
2. **[reliability]** Empty/whitespace-only manifest silently fell through to "safe to archive"
   instead of the intended conservative result — **fixed in `cd40515`** (+ regression test).
3. **[reliability]** A cross-session merge that updates the *original* stub in place never
   re-arms the reminder (safe-direction limitation, not a fix-in-PR-sized change) — **filed as
   [dev-env#666](https://github.com/brownm09/dev-env/issues/666)**, documented in the function's
   docstring.

Closes #651


brownm09:
## Review: fix(hooks): gate stub-push archive reminder on a confirmed-resolved PR

### Summary
Merge with minor changes already applied. The core fix is sound: two independent Opus review
passes (correctness/security; reliability/performance/maintainability) traced every path that
could produce the harmful outcome (arming the archive sentinel while a PR is genuinely still open)
and found the design fails safe throughout. Two real non-blocking findings surfaced in
`head_commit_has_unresolved_pr()` — both already fixed in commit `cd40515` with regression tests.
One accepted, safe-direction limitation was also found and filed as
[dev-env#666](https://github.com/brownm09/dev-env/issues/666) rather than fixed here, since a
proper fix is a design decision (not a small patch). Diff size (453 lines) is appropriately scoped
for a bug fix plus its required test/doc trail — no split needed.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/stub-push-archive-reminder.py` (`head_commit_has_unresolved_pr`)
The new manifest read used a bare `manifest_file.read_text(encoding="utf-8")` guarded by
`except OSError`, instead of the shared `_journal_schema.decode_shard_bytes(raw)` helper that this
module's other two consumers (`validate-manifest.py`, `journal-shard-write-advisory.py`) already
use to decode shard bytes.
This PR explicitly makes the hook a third consumer of `_journal_schema`, and `decode_shard_bytes`
exists precisely so every consumer decodes shard bytes identically (BOM-stripping, then still
validating). Diverging has two consequences: (a) a UTF-8-BOM-prefixed but actually-resolved
manifest fails JSON parse here and is treated as unresolved, where the other two consumers would
strip the BOM and read the real entry; (b) `read_text(encoding="utf-8")` raises
`UnicodeDecodeError` (a `ValueError`, not `OSError`) on non-UTF-8 bytes, so the local
`except OSError: return True` never catches it — the intended per-manifest conservative-True
contract is silently bypassed, surviving only because the module-level `except Exception` in
`__main__` happens to swallow it.
**Suggestion:** Read bytes and route through `decode_shard_bytes`, treating `None`-text as
conservative-True, matching the other two consumers.

**Disposition:** Fixed in `cd40515` — now reads via `manifest_file.read_bytes()` +
`decode_shard_bytes`, with a regression test pinning a UTF-8-BOM-prefixed-but-resolved manifest.

---

**[reliability]** `claude/scripts/stub-push-archive-reminder.py` (`head_commit_has_unresolved_pr`)
When a touched stub's paired manifest exists but is empty or whitespace-only,
`parse_manifest_text` returns `[]`, the entry loop never runs, and the function falls through to
`return False` — "no unresolved PR, safe to archive."
This is a fail-open path in the exact safety gate this PR adds, asymmetric with the *missing*
manifest case one branch above (which correctly returns `True`). A truncated or not-yet-fully
flushed manifest write for a session with a genuinely open PR would arm the archive reminder —
the precise harm this PR exists to prevent. It also contradicts the PR's own stated "conservative
on every ambiguity" contract; empty shards are already a recognized state elsewhere
(`journal-shard-write-advisory.py` reports `"empty shard (no JSON object)"`), so this isn't purely
theoretical, though likely rare in normal operation.
**Suggestion:** Treat a zero-entry parse result as conservative-`True`.

**Disposition:** Fixed in `cd40515` — an empty `entries` list now returns `True`, with a
regression test pinning an empty/whitespace-only manifest.

---

**[reliability]** `claude/scripts/stub-push-archive-reminder.py` (`head_commit_has_unresolved_pr`)
In the documented cross-session-merge workflow where a *later* session updates the *original*
opening session's stub in place (`claude/CLAUDE.md`: "New session: update the opening stub in
place for minor merges ... set `prs_closed:[N]` in **this session's** manifest shard"), the merge
commit touches the original session's stub but writes `prs_closed` into the *merging* session's
own, separate manifest — never the original session's. Because this function derives a manifest
from each touched stub's own path, it keeps reading the original session's stale
`prs_opened:[N],"prs_closed":[]` forever, so the archive reminder never re-arms for that specific
push, even though the merge genuinely completed.
This fails in the same safe direction as every other ambiguity here (a missed nudge, not a false
archive) and doesn't reproduce the bug this PR fixes — it's also narrower than the common
same-session flow the fix targets, both of which the reviewing agent confirmed arm correctly.
**Suggestion:** A proper fix likely means treating "this commit deletes
`open-prs/<N>.json`" as corroborating evidence, which would require reasoning about individual PR
numbers rather than a single bool — a design change, not a small patch.

**Disposition:** Filed as [dev-env#666](https://github.com/brownm09/dev-env/issues/666) — documented
in the function's docstring in this PR; not fixed here since the proper fix is a design decision
this PR's own ADR amendment already weighed (and rejected as a *primary* data source) for a
different reason.

---

### Findings Disposition

- Maintainability finding (decode_shard_bytes) — fixed in `cd40515`.
- Reliability finding (empty-manifest fail-open) — fixed in `cd40515`.
- Reliability finding (cross-session-merge stub-in-place gap) — filed [dev-env#666](https://github.com/brownm09/dev-env/issues/666).

---

*Reviewed by `claude-opus-4-8` (two parallel passes) via Claude Code*

PR #661, opened 2026-07-09
fix(hooks): guard engineering-journal canonical against branch hijack; fix dev-env-sync detached-HEAD gap

### LINKED ISSUE #619: dev-env-sync.py never detects or corrects a detached canonical HEAD
Sub-issue of #617.

## Problem

`claude/scripts/dev-env-sync.py` (the `UserPromptSubmit` hook that keeps the canonical dev-env checkout synced and warns when it's off `main`) determines the current branch like this:

```python
branch = run(["git", "symbolic-ref", "--short", "HEAD"])
if branch.returncode != 0:
    sys.exit(0)
current_branch = branch.stdout.strip()
if current_branch != "main":
    # ... thorough diagnose/auto-correct/warn logic (worktree topology,
    # squatter detection, dirty-drift preservation) ...
```

`git symbolic-ref --short HEAD` **fails** (non-zero exit) when HEAD is detached — there is no symbolic ref to resolve. When that happens, line 59-60 responds with a silent `sys.exit(0)`, **before ever reaching the off-main diagnostic block** (lines 62-125) that otherwise handles every other "canonical not on main" scenario thoroughly:
- auto-returns a clean canonical to `main`,
- warns by name when a sibling worktree is squatting `main`,
- warns without switching when the canonical has dirty drift.

None of that runs for a detached HEAD. The script's own docstring says it "warns if the repo has diverged and needs manual attention" — but a detached canonical produces **zero warning, on every single prompt of every session**, for as long as it stays detached, since this hook fires on every `UserPromptSubmit` across every concurrent session.

## Confirmed impact (from #617)

The canonical checkout sat with a detached HEAD from 2026-07-07T10:19:17Z until discovered by chance during unrelated post-merge cleanup roughly 13 hours later. Given this script runs on every prompt in every session touching dev-env in that window, it silently did nothing every single time — no stderr warning, unlike the equivalent wrong-branch cases which reliably print one.

## Recommendation

Detect detached HEAD explicitly and route it into the same diagnostic path already used for "on the wrong branch" — e.g.:

```python
branch = run(["git", "symbolic-ref", "--short", "HEAD"])
if branch.returncode != 0:
    # Detached HEAD - not the same as "command failed for some other reason".
    # A clean detached-at-main's-own-tip canonical is exactly the safe
    # auto-return case; anything else should warn like the wrong-branch path.
    ...
```

A clean, non-diverged detached HEAD (as in #617 — zero unique commits, matching `main`'s tip exactly) is exactly the "safe to auto-return" case `canonical_sync_action`'s `return-canonical` path already models for the wrong-branch scenario; it should reach that same function rather than exiting silently first.

Consider whether this warrants an amendment to ADR-058 (which documents this script's topology diagnosis/auto-correction design).

## Context

Filed from the investigation in #617 (canonical dev-env checkout detached-HEAD incident, 2026-07-07). Required reading `claude/scripts/dev-env-sync.py` directly to confirm — this is a read-from-source finding, not an inference from an error message.

### LINKED ISSUE #630: Scheduled-task worktree provisioning hijacks the engineering-journal canonical checkout's HEAD — reproduced 2 days running

brownm09:
## Summary

- New `UserPromptSubmit` hook `journal-canonical-guard.py` detects the engineering-journal canonical checkout hijacked onto a stray `claude/<slug>` branch or left detached (a scheduled-task worktree-provisioning defect, reproduced 2 mornings running via reflog evidence — see #630) and restores it to `main`. Unlike `dev-env-sync.py`, it does **not** treat "off `main`" as broken, since engineering-journal's canonical is legitimately on `draft/YYYY-MM-DD` for most of every working day — only the actual hijack signature (`is_hijacked_branch()`) triggers correction.
- While extending the shared `_worktree_topology.py` module for this, also closes the closely related #619: `dev-env-sync.py` never detected a detached canonical HEAD, because `git symbolic-ref` failing on detached HEAD caused a silent early exit before its own diagnose/auto-correct/warn block ever ran (documented as ADR-058 Amendment, 2026-07-09).
- Filed [#657](https://github.com/brownm09/dev-env/issues/657) for an adjacent, out-of-scope finding surfaced during design review: `biweekly-retro` and `weekly-memory-audit` still run the canonical-mutating `sync-routine-worktree` sequence against this same shared canonical — the exact hazard ADR-082 already removed from their sibling `daily-journal-compose`. Not fixed here (different files/mechanism).

Closes #630
Closes #619

## Design

Full design rationale, TOCTOU race analysis, and rejected alternatives are in [ADR-093](docs/adr/093-journal-canonical-hijack-guard.md) (new) and the [ADR-058 2026-07-09 amendment](docs/adr/058-worktree-squatting-main-detection-correction.md#amendment-2026-07-09--dev-env-syncpy-never-detected-a-detached-canonical-head-dev-env619) (dev-env#619 fix). Short version:

- **Why not fix this in `daily-journal-compose`/`journal-compose`**: both deliberately avoid mutating the canonical for exactly this hazard class (ADR-082); the correction has to live outside them, as its own mechanism.
- **Why not "off main" as the gate**: that's `dev-env-sync.py`'s invariant, not engineering-journal's — its canonical is intentionally off `main` most of the day.
- **TOCTOU fix**: the hijack check is re-verified against a freshly-read `git worktree list` immediately before the corrective `git checkout main`, since a concurrent stub-writing session could have already legitimately moved the canonical between the cheap first read and the more expensive second one. Caught during a Plan-agent design review before implementation, not observed live.
- **Non-destructive**: `git checkout main` never deletes the hijacked branch or its commits.

## Testing

No `## Testing` section exists yet for automated test suites beyond the per-file convention this repo already uses (pure-helper-offline tests). Ran the required commands:

```
py -3 claude/scripts/tests/test_worktree_topology.py

Tests: 7 passed, 0 skipped, 0 failed   (confirms all 43 settings.json hook entries, incl. the new one, resolve + parse)

py -3 -c "import ast,sys,glob; [ast.parse(open(f,encoding='utf-8').read(),f) for f in glob.glob('claude/scripts/*.py')]"
SYNTAX OK
```

**Manual fixture verification** for `journal-canonical-guard.py`'s orchestration (no dedicated test file — it has zero local pure logic, same shape as `dev-env-sync.py`, which also has none; matches this repo's established convention for topology-diagnosing orchestration scripts, e.g. `## Testing` item 22/26/30). Created a throwaway git repo, ran the real script against it via the new `JOURNAL_CANONICAL_GUARD_REPO_PATH` env-var override, confirmed all four scenarios live:
1. `claude/*` hijack branch + clean → restored to `main`, non-destructive (hijack branch still exists).
2. Detached HEAD + clean → restored to `main` (verified the success message doesn't offer a bogus `checkout <detached>` recovery command — caught and fixed during this verification pass).
3. Legitimate `draft/2026-07-09` branch → left untouched, silent no-op (the specific false-positive this predicate exists to avoid).
4. `claude/*` hijack + uncommitted changes → warned to stderr, did **not** switch (drift preserved).

Suppression / test-integrity / deleted-test-file / coverage-threshold pre-PR greps: all clean.

Re-ran the full test suite and all four fixture scenarios after the review-driven fixes (commit `fc80077`) — no regression.

Full review: https://github.com/brownm09/dev-env/pull/661#issuecomment-4927632323 (2 blocking, 2 non-blocking, all addressed)

- **[correctness] TOCTOU residual window** (git status + checkout still inside the exposed race window) — fixed in `fc80077`
- **[reliability] Unguarded subprocess exceptions** (violated the "Exit 0 always" docstring contract) — fixed in `fc80077`
- **[maintainability] Hardcoded `<detached>` sentinel** (4 call sites, drift risk) — fixed in `fc80077`
- **[maintainability] Duplicated correction flow** between the two canonical-guard hooks — documented as an accepted two-consumer trade-off in ADR-093, fixed in `fc80077`

## Doc reconciliation

- `README.md`: new hooks-table row for `journal-canonical-guard.py`; updated the `_worktree_topology.py` shared-module prose (line ~57) to name the fourth consumer and the module's two distinct invariants.
- `docs/REFERENCE.md`: new hooks-section entry; updated `dev-env-sync.py`'s entry to mention the detached-HEAD fix; added a cross-reference from the "Draft branch recovery" runbook's closing sentence.
- `_worktree_topology.py` module docstring: documents the two distinct invariants the module now hosts and names the fourth consumer.
- dev-env `CLAUDE.md` `## Testing`: extended item 22 (not a new item — the new script has no local pure logic, same as `dev-env-sync.py`) with the two new helpers and their coverage.

## ADR-warrant

Warranted — new hook mechanism, new shared-module invariant, workflow-rule-adjacent. [ADR-093](docs/adr/093-journal-canonical-hijack-guard.md) (new, cites ADR-086 as closest structural precedent) + [ADR-058 amendment](docs/adr/058-worktree-squatting-main-detection-correction.md) (dev-env#619 fix). `docs/adr/INDEX.md` updated.

## Scope note

Deliberately scoped to engineering-journal only — the one repo with confirmed, reproduced incidents. Career-playbook (`nightly-cover-letters`, identical 07:00 fire time) and research-notes (`nightly-research`) carry the same theoretical exposure but are unconfirmed/uninstrumented; stated explicitly in ADR-093 rather than left implicit.


brownm09:
## Review: fix(hooks): guard engineering-journal canonical against branch hijack; fix dev-env-sync detached-HEAD gap

### Summary

Merge-ready. The design is sound and well-precedented (reuses ADR-058's shared topology module, correctly scopes the hijack gate narrower than dev-env-sync's "off main" invariant since engineering-journal's canonical is legitimately on `draft/YYYY-MM-DD` most of the day). Diff was 649 lines (over the 400-line single-pass threshold), so this went through parallel Correctness/Security and Reliability/Performance/Maintainability subagent passes rather than one pass. Both surfaced real, substantive findings — no rubber-stamping. The highest-severity finding was a TOCTOU race in the original two-read design that could, in rare cases, have caused the guard itself to yank a concurrent session's legitimate branch back to `main`. All findings (2 blocking, 2 non-blocking) have already been fixed in-PR; see Findings Disposition below.

---

### Blocking Findings

**[correctness]** `claude/scripts/journal-canonical-guard.py:101-130` (pre-fix)

The hijack check was re-verified once against a fresh `git worktree list` read before acting, but the subsequent `git status --porcelain` call and the `git checkout main` mutation itself remained inside the exposed window. Engineering-journal's canonical is moved between `main` and `draft/YYYY-MM-DD` constantly by many concurrent stub-writing sessions — unlike dev-env's own canonical, which is never legitimately off `main` at all — so this residual window is meaningfully more likely to be hit here than the ADR's "same residual as dev-env-sync" framing implied. A concurrent session completing its own branch switch inside this window would have its work yanked back to `main` by the guard meant to prevent exactly that harm.

**Fix:** Added a third, final `git symbolic-ref` re-check immediately before the mutating `git checkout main` call, after the `git status` read. This narrows the residual to a single subprocess-call window, matching dev-env-sync.py's actual residual rather than just asserting parity with it. ADR-093 updated to document the two-round TOCTOU tightening explicitly.

---

**[reliability]** `claude/scripts/journal-canonical-guard.py:169-170` (pre-fix)

The script's docstring promises "Exit 0 always — never block the user's prompt," but nothing enforced this against an uncaught subprocess exception (`TimeoutExpired`, `FileNotFoundError` if `git` were ever off `PATH`, or another `OSError`) — any of the four git calls in `main()` could have propagated a traceback with a non-zero exit. The closest sibling hook guarding this same repo on the same event, `new-day-journal-check.py`, already wraps every `subprocess.run` call in `try/except Exception`, making this a deviation from the established fail-open convention rather than a faithful copy of `dev-env-sync.py`'s (also imperfect) pattern.

**Fix:** Wrapped the `if __name__ == "__main__":` entry point in `try/except Exception: sys.exit(0)`, matching the established convention.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/_worktree_topology.py` / `claude/scripts/journal-canonical-guard.py:151` (pre-fix)

The `"<detached>"` sentinel was a bare string literal repeated at 4 call sites across 2 files (`parse_worktree_porcelain`, `main_squatter`, `resolve_current_branch`, `is_hijacked_branch`, plus the consumer's recovery-message check). If the spelling ever changed at one site and not another, the consumer would silently diverge and could emit a bogus recovery command.

**Fix:** Extracted a single `DETACHED = "<detached>"` module-level constant in `_worktree_topology.py`, referenced by all five sites.

---

**[maintainability]** `claude/scripts/journal-canonical-guard.py:108-164` vs `claude/scripts/dev-env-sync.py:92-130`

The correction *flow* (git-call sequence, three-way `warn-squatter`/`warn-dirty`/`return-canonical` dispatch) is near-duplicated between the two hooks — only the *decision* logic is properly shared via `_worktree_topology.py`. A future change to the correction protocol needs editing in both places and can drift.

**Fix:** Documented as an explicit, accepted trade-off in ADR-093's Consequences section (same shape as the ADR's existing "don't generalize to other repos yet" decision, one level lower) — acceptable for two consumers; extract a shared `correct_canonical(repo, gate_predicate, message_set)` orchestrator when a third canonical-guarding hook appears, rather than copying the flow a third time.

---

*One performance finding (registering a 12th unconditional `UserPromptSubmit` hook adds a per-prompt process-spawn cost machine-wide) was raised and explicitly self-assessed by its reviewer as not warranting a change — it mirrors `dev-env-sync.py`'s already-accepted always-on cost, and no cheaper correct shape exists without merging two hooks with different invariants, which isn't warranted. Dropped rather than recorded, per this review's own guidance not to log findings that don't meet the fix-or-file bar.*

---

### Findings Disposition

- **[correctness] TOCTOU residual window** — fixed in `fc80077`
- **[reliability] Unguarded subprocess exceptions** — fixed in `fc80077`
- **[maintainability] Hardcoded `<detached>` sentinel** — fixed in `fc80077`
- **[maintainability] Duplicated correction flow** — documented as accepted trade-off in ADR-093, fixed in `fc80077`

---

*Reviewed by claude-opus-4-8 (parallel subagents) via Claude Code*

PR #665, opened 2026-07-09
[fix] Mask the bare positional PR-number regex family (dev-env#650)

brownm09:
## Summary

Extends dev-env#626/#634's quote-awareness fix (ADR-050 Amendments 15/17) to a third, distinct
decoy shape flagged by dev-env#650: a bare positional integer inside `--subject`/`--body` prose
(e.g. `--subject "resolves 42 items"`) was read as the real PR/issue number by four sites' identical
`(?<!\S)(\d+)(?=\s|$)`-shaped regex, none of them protected by `mask_quoted_spans` or
`mask_prose_flag_values` — both were built for flag/URL extraction, not bare-number extraction.

1. **`posttooluse-inert-advisory.py::_devenv_merge_pr`** — `num_m = _MERGE_POS_NUM_RE.search(args)`
   now runs against a `mask_quoted_spans`-masked copy of `args`, shared with the pre-existing
   `repo_m` check (one fewer redundant masking walk over the same string).
2. **`post-pr-merge-project.py::extract_pr_number_from_command`** — same fix for its own
   `(?<!\S)(\d+)(?=\s|$)` inline regex. Its `_PR_URL_RE` fallback is untouched (see "Out of scope" below).
3. **`stop-tile-enumeration-gate.py::_target_pr`** — named in the issue.
4. **`stop-tile-enumeration-gate.py::_closed_issue_number`** — **not** named in the issue, found by
   grepping the repo for the exact `(?<!\S)(\d+)(?=\s|$)` pattern per dev-env#650's own closing
   suggestion ("worth grepping... before assuming these three are exhaustive"). Same file, same
   compiled `_POS_NUM_RE` object, same vulnerability shape — confirmed live against the pre-fix code
   before writing the fix.

Each site's own PR-URL/issue-URL fallback regex stays on unmasked text, matching this hook family's
established per-site masking-scope discipline (mirrors `_explicit_repo`'s own documented scope
decision immediately above `_target_pr` in the same file).

A repo-wide grep for the exact pattern after this fix confirms no other `claude/scripts/*.py` file
matches it — the sweep is exhaustive as of this PR.

**Out of scope, filed as a follow-up:** auditing `post-pr-merge-project.py::extract_pr_number_from_command`
for this fix surfaced a structurally distinct, previously-undocumented gap in the *same function* — its
`_PR_URL_RE` fallback (a decoy PR **URL**, not a bare number) still runs unmasked. Confirmed live:
[...]

PR #662, opened 2026-07-09
[fix] pr-merge-reminder.py: resolve --repo/-R flag for gh pr create, not just merge

brownm09:
## Summary

- `pr-merge-reminder.py`'s `is_create` branch always reported `cwd` in the journal reminder, with
  no `--repo`/`-R` flag resolution at all — unlike the `merge_ok` branch's `_effective_merge_repo`
  (dev-env#470/#616). A cross-repo `gh pr create --repo other/repo` run from an unrelated cwd (e.g.
  dev-env instructions edited from a project-cwd session) risked writing the journal stub and
  open-PR shard under the wrong project's journal — live repro in dev-env#644 (2026-07-08), caught
  only because the session independently knew better.
- Adds `_effective_create_repo(command, cwd)`, structurally identical to `_effective_merge_repo`:
  an explicit `--repo`/`-R` flag (via the same `_REPO_FLAG_RE`, run against a
  `mask_quoted_spans`-masked copy of `command`) takes precedence, falling back to `cwd` otherwise.
  Unlike the merge path, there's no cd-chain-aware dir to fall back to — this branch has only ever
  reported plain `cwd`, so an unflagged create command's reminder is unchanged.
- `_build_messages`'s `is_create` branch now prints both `cwd:` and `repo:` lines and says "Identify
  the project journal path from the repo above" — mirroring the `merge_ok` branch's existing display
  and phrasing exactly.
- Docs: `_hookio.py`'s `mask_quoted_spans` comment gains a new call-site paragraph (following
  Amendment 17's own precedent of adding rather than rewriting), `CLAUDE.md` Testing item 28 and
  `docs/REFERENCE.md`'s hook description are updated, and `docs/adr/050-shared-hookio-sibling-hook-fixes.md`
  gains Amendment 18 (this ADR has an amendment for every prior fix in this exact bug family —
  14, 15, 16, 17 all touch `_REPO_FLAG_RE`/masking/create-detection).

## Test plan

- [x] Syntax check: `py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py` — OK
- [x] `py -3 claude/scripts/tests/test_pr_merge_reminder.py` — Tests: 56 passed, 0 skipped, 0 failed (up
      from 52 — 4 new `_effective_create_repo` tests: explicit flag overrides cwd (the dev-env#646 repro),
[...]

PR #668, opened 2026-07-09
[fix] Mask the gh pr merge args-region boundary, not just the search within it

brownm09:
## Summary

Fixes dev-env#660: the `gh pr merge` args-region **boundary**-finding step in
4 sites was quote-unaware, even though Amendments 15/17 (ADR-050) already made
the **search within** that region quote-aware. A `--subject`/`--body` value
containing a bare `&`, `|`, `;`, or `\n` — including an entirely ordinary
commit subject like `"R&D tracking"`, no deliberate crafting needed —
truncated the region before a later real `--repo` flag or PR number was ever
seen, silently dropping it. Opposite failure direction from #626/#634's
decoy-hijack (a fake flag mistaken for real): here a **real** flag is lost.

Confirmed live in 3 sites:
- `pre-merge-findings-gate.py::_parse_merge_target` (ADR-028/039 merge gate)
- `post-pr-merge-project.py::extract_pr_number_from_command` /
  `extract_repo_from_command` (board-mutation hook — reintroduces the
  dev-env#559 wrong-repo-board-mutation shape via a different vector)
- `posttooluse-inert-advisory.py::_devenv_merge_pr`

A 4th site, `pre-auto-merge-checkpoint-gate.py::_merge_tail`, has the
identical pattern (explicitly documented as "mirrors `_parse_merge_target`'s
own tail-extraction") but is **not** a live bypass — its `wants_auto_merge()`
already fails closed via a `shlex.split()` `ValueError` on the resulting
unterminated quote (dev-env PR #588's fallback). Fixed for consistency, not
because a live gap was confirmed.

`pr-merge-reminder.py` and `post-pr-merge-pull.py` are unaffected — both mask
the whole command directly before searching, with no separate pre-truncation
step.

## Fix

Same technique at all 4 sites: `mask_quoted_spans` the whole relevant text
**before** the boundary-finding split/regex runs, then use the match's
**span offsets** (length-preserving masking) to slice the original unmasked
text. Every existing stop-character-set is preserved exactly — this is a
quote-awareness fix, not a behavior change to which characters bound the
region.

Full writeup: [ADR-050 Amendment 20](docs/adr/050-shared-hookio-sibling-hook-fixes.md#amendment-20-2026-07-09--quote-aware-args-region-boundary-not-just-the-search-within-it-dev-env660).

**Amendment numbering note:** this PR's amendment was renumbered twice after two independent,
concurrently-developed PRs each claimed an earlier number first: PR #662 (dev-env#646) merged as
"Amendment 18", then PR #665 (dev-env#650) merged as "Amendment 19" — both while this PR was open.
Rebased through both (plus two unrelated docs-only `origin/main` moves) and settled on Amendment
20. Both PRs' fixes compose correctly with this one at the code level (verified directly, not
assumed) — they fix a different, complementary layer of the same functions (within-region search
vs. this PR's region-boundary detection).

## Testing

```
py -3 claude/scripts/tests/test_pre_merge_findings_gate.py
bash claude/scripts/tests/test-merge-findings-gate.sh
py -3 claude/scripts/tests/test_pre_auto_merge_checkpoint_gate.py
```

Tests: 8 + 13 + 40 + 32 + 33 + 86 + 13 passed, 0 skipped, 0 failed (225 total
across the touched-file suites; `test_hookio.py` unchanged in behavior — only
a documentation-only docstring addition, plus additional call sites of its
existing, unmodified `mask_quoted_spans`; `test_post_pr_merge_project.py`'s
40 includes 2 tests that landed via PR #665's own rebase-through, not this
PR's diff). New coverage from this PR: 6 tests in
`test_post_pr_merge_project.py`, 1 bundled test (4 assertions) in
`test_posttooluse_inert_advisory.py`, 3 tests in
`test_pre_auto_merge_checkpoint_gate.py`, and `test-merge-findings-gate.sh`
step 9 (2 assertions covering 4 `print()` cases) — including, added during
`/review` on this PR, a case combining a quoted separator-decoy WITH a real
trailing `&&` chain in the same command at all 4 sites (see Findings
Disposition below). All pre-existing tests re-run and pass unmodified —
confirms no regression to the already-fixed within-region search behavior
(Amendments 15/17/19) or any other existing decision path, and that this
PR's boundary-fix composes correctly with PR #665's within-region bare-number
masking fix in the two files both PRs touched.

Also re-ran `test_stub_push_archive_reminder.py` / `test_journal_schema.py`
(PR #659), `test_pr_merge_reminder.py` (PR #662), `test_worktree_topology.py`
(PR #661), and `test_stop_tile_enumeration_gate.py` (PR #665) — all merged
after this branch was first cut and rebased through since — as a courtesy
check post-rebase. All unaffected, pass unchanged.

## Risk-dimension audit

1. **Testing** — see above.
2. **Observability** — N/A; no hook I/O contract changed (same exit codes, same stderr/stdout usage).
3. **Security** — this *is* the security-relevant dimension: closes a silent-data-loss gap in a security-relevant merge-findings gate (ADR-028/039) and a board-mutation hook. No new attack surface; strictly more correct, no existing check weakened.
4. **Resilience/failure modes** — all existing fail-open/fail-closed behaviors preserved unchanged (e.g. site 4's `ValueError` fallback still exists as a backstop). Revert = revert this one PR; no migration/state involved.
5. **Performance** — negligible: one extra linear `mask_quoted_spans()` scan per hook invocation on short CLI command strings; not a hot path (hooks run once per Bash tool call).
6. **Data integrity & migrations** — N/A.

## Documentation Maintenance

N/A — no skill/hook/routine added, removed, or renamed; no `hook-config.json`
schema change; no skill invocation-syntax change. README.md / docs/REFERENCE.md
are unaffected. `docs/adr/INDEX.md` updated (new tags on ADR-050's row).

## ADR-warrant check

Warranted and done — this amends ADR-050 (Amendment 20), since the change
modifies documented hook behavior in the `gh pr merge` hook family.

Both non-blocking findings from `/review` fixed in this PR (no filed issues):

1. **[test-coverage]** Missing test combining a quoted separator-decoy with a
   real trailing `&&` chain in the same command — added to all 4 sites'
   suites (`test_post_pr_merge_project.py`, `test_posttooluse_inert_advisory.py`,
   `test_pre_auto_merge_checkpoint_gate.py`, `test-merge-findings-gate.sh` step 9).
2. **[documentation]** The newline-preservation residual limitation was
   documented in the ADR/issue but not in source — added a paragraph to
   `mask_quoted_spans`'s docstring in `_hookio.py` (documentation only, no
   behavior change).

Closes #660


brownm09:
**Renumbered ADR-050 Amendment 18 -> 19** after pushing: PR #662 (dev-env#646) independently claimed "Amendment 18" and merged first (2026-07-09T16:58:38Z), while this branch was still open. Rebased twice more (PR #661 also landed in between) and renumbered this PR's own amendment plus all cross-references (CLAUDE.md testing items 14/18/38, docstrings in all 4 fixed hook scripts, their test files) from 18 to 19. Full test suite re-run clean after each rebase.

This is exactly the numbering-collision class `pre-merge-numbering-check.py` (ADR-074, dev-env#516) exists to catch at merge time -- caught proactively here via a routine pre-push freshness check instead.

brownm09:
## Review: [fix] Mask the gh pr merge args-region boundary, not just the search within it

### Summary
Merge-ready with two small, cheap non-blocking improvements worth taking in-PR rather than filing. The fix itself is sound: I independently re-derived the offset-preservation argument for the mask-then-reslice technique and verified it against the actual code (not just the PR's own claims) for all 4 sites, including one combined edge case not covered by the existing test suite (see below) — it resolves correctly. No blocking findings.

---

### Non-Blocking Findings

**[test-coverage]** `claude/scripts/tests/test_post_pr_merge_project.py` (and the sibling suites for the other 3 sites)
None of the new tests combine a quoted separator-decoy (e.g. `--subject "a && b"`) with a REAL trailing top-level `&&` chain in the *same* command. Each shape is tested separately (decoy-only in the new dev-env#660 tests; real-chain-only in `test_merge_args_chained_command_still_excluded_after_fix`), but the fix's correctness actually depends on the boundary-finder picking the *first unmasked* separator when both are present — verified this manually against the live code (`'gh pr merge 42 --subject "a && b" --repo brownm09/dev-env && rm -rf /'` correctly resolves to `('42', 'brownm09/dev-env')`, not swallowing the real chain), and it passes, but nothing in the suite pins it.
**Suggestion:** Add one combined-case test (decoy quote followed by a real `&&` chain) to at least `test_post_pr_merge_project.py` and `test-merge-findings-gate.sh`'s step 9 — this is exactly the interaction a future refactor of the masking/boundary logic is most likely to silently break.

**[documentation]** `claude/scripts/_hookio.py` (not touched by this diff) — `mask_quoted_spans`'s docstring
The ADR-050 Amendment 19 writeup and the GitHub issue both candidly document a known residual limitation (an embedded literal newline inside a still-open quoted value or heredoc body defeats the new boundary-finding fix, since `mask_quoted_spans` deliberately preserves newlines unmasked). None of the 4 new call-site docstrings (`_parse_merge_target`, `_merge_args`, `_devenv_merge_pr`, `_merge_tail`) or `mask_quoted_spans` itself mention this. A future maintainer reading the source (not archaeologizing through ADR-050's amendment history) has no way to discover this trade-off.
**Suggestion:** Add a short paragraph to `mask_quoted_spans`'s docstring in `_hookio.py` noting the newline-preservation caveat for boundary-finding callers specifically (this PR is the first consumer to use it that way) — centralizing it there once, rather than repeating it in 4 call sites.

---

### Findings Disposition
- test-coverage (combined decoy+chain case): fixing in this PR.
- documentation (mask_quoted_spans newline caveat): fixing in this PR.

---

brownm09:
**Renumbered again: Amendment 19 -> 20.** PR #665 (dev-env#650) merged and had *also* independently claimed "Amendment 19" (the prior comment's own prediction of a possible 3rd collision came true). Rebased through it -- both PRs' fixes compose correctly at the code level in the two files both touched (`post-pr-merge-project.py`, `posttooluse-inert-advisory.py`): this PR's boundary-detection fix and PR #665's within-region bare-number-decoy masking fix are complementary layers of the same functions, verified directly by reading the auto-merged code rather than assumed.

Also caught and fixed a real mistake from the *first* renumbering pass: an earlier blanket find-and-replace on `CLAUDE.md` had accidentally changed an unrelated PR #662 (dev-env#646) cross-reference from "Amendment 18" to "19" -- reverted back to 18 during this pass's careful line-by-line audit.

PR body and ADR-050 updated to Amendment 20. Full test suite re-run clean, including the sibling test files from all concurrently-merged PRs (#659, #661, #662, #665) as a courtesy check.

PR #671, opened 2026-07-09
fix(journal-compose): mechanically enforce the today-guard

### LINKED ISSUE #631: journal-compose today-guard is a prompt-level instruction, not a mechanical gate — agent reasoned its way past it (confirmed via transcript)

brownm09:
## Summary

Closes #631. Converts `/journal-compose`'s today-guard (ADR-017) from a prose-only instruction
into a two-part mechanical gate, per the issue's proposed fix — mirroring this repo's ADR-028 →
ADR-039 precedent (a policy ADR later hardened by a dedicated enforcement ADR).

A 2026-07-08 scheduled-task transcript confirmed an agent noticed the today-guard should block it,
reasoned "this is an automated scheduled task... so I should treat it as having `--force=true`
semantics," and proceeded — without the guard's own refusal text ever actually being emitted or
read. Nothing mechanically enforced the guard; it was prose an agent could reason past.

**Part 1 — deterministic `FORCE` resolution.** New `claude/scripts/journal-compose-force-resolve.py`
runs as the literal first Bash action of `/journal-compose` Step 0.6, using the harness-substituted
`$ARGUMENTS` text verbatim (before any of the agent's own reasoning). It regex-matches a
whitespace-bounded `--force` token and writes a same-day marker file recording the resolution.

**Part 2 — mechanical gate.** New `PreToolUse(Bash)` hook
`claude/scripts/pre-tool-use-journal-compose-force-guard.py` (mirroring
`pre-merge-findings-gate.py`/`pre-tool-use-canonical-mutate-guard.py`'s structure) blocks a git
`worktree`/`commit`/`push` command that references a `draft/<today>` or `compose[-/]<today>`
target unless a fresh, `force=true` marker exists. A commit message's `-m`/`--message`/`-F`/`--file`
*value* is excluded from the scan, so a commit merely discussing "draft/2026-07-09" as prose (this
repo's own commits legitimately do) can never false-trigger. Deliberately **fails closed** on a
missing/corrupt/stale marker (a reversal of this hook family's usual fail-open convention — justified
by the hook's narrow, already-rare trigger) and ships with **no override token** (an override would
just be a second, easier route to reason past — the exact vulnerability this closes).

**Part 3 — explicit prohibition.** `claude/routines/daily-journal-compose/SKILL.md` and the live
`daily-journal-compose-local` scheduled task (updated in the same change via `update_scheduled_task`
— its own text already warned it drifted from canonical once before, dev-env#615) now both state:
never pass `--force` from this routine, and never reason that `--force` is implied; a guard refusal
is success, not a problem to work around.

Full rationale, alternatives considered, and limitations: [ADR-096](docs/adr/096-journal-compose-mechanical-force-guard.md)
(renumbered twice while this branch was in flight — first 093→094 when origin/main claimed 093 for
the sibling dev-env#630 fix, then 094→096 when origin/main separately claimed both 094 and 095 for
two more concurrently-merged PRs; `CLAUDE.md` Testing items were similarly renumbered 51–53 → 53–55
when origin/main claimed 51–52 for a third PR. All three collisions were caught proactively via a
pre-push `origin/main` fetch, never by the merge-time `pre-merge-numbering-check.py` gate itself.)

## Changes

- `claude/scripts/_journal_compose_force.py` (new) — shared pure module: `resolve_force`,
  `marker_path_for`, `build_marker`, `write_marker`/`read_marker`, `is_marker_fresh`.
- `claude/scripts/journal-compose-force-resolve.py` (new) — skill-invoked CLI script.
- `claude/scripts/pre-tool-use-journal-compose-force-guard.py` (new) — the enforcement hook.
- `claude/settings.json` — registers the new hook under `PreToolUse`/`Bash`.
- `claude/skills/journal-compose/SKILL.md` — Step 0.6 now resolves `FORCE` mechanically as its
  first action; today-guard prose now notes the hook backstop and its documented residual gap.
- `claude/routines/daily-journal-compose/SKILL.md` + live `daily-journal-compose-local` task —
  explicit "never pass --force" prohibition.
- `docs/adr/096-journal-compose-mechanical-force-guard.md` (new) + `docs/adr/INDEX.md`.
- `README.md`, `docs/REFERENCE.md`, `CLAUDE.md` — doc reconciliation (new hook rows, updated
  `/journal-compose` entry, new numbered Testing items 53–55).

## Testing

Tests: 102 passed, 0 skipped, 0 failed (three test files covering the shared module, the resolve
script, and the guard hook — 44 + 32 + 6 across the three) + 13 passed, 0 skipped, 0 failed
(repo-wide `test_no_crude_command_substring_checks.py` gate) + 7 passed, 0 skipped, 0 failed
(`test_pyw_stdio.py`, confirming the new hook correctly needs no `_winsubp` import since it spawns
no subprocess).

```bash
py -3 claude/scripts/tests/test_journal_compose_force.py                        # 32 passed
py -3 claude/scripts/tests/test_journal_compose_force_resolve.py                # 6 passed
py -3 claude/scripts/tests/test_pre_tool_use_journal_compose_force_guard.py     # 44 passed
py -3 claude/scripts/tests/test_no_crude_command_substring_checks.py            # 13 passed
py -3 claude/scripts/tests/test_pyw_stdio.py                                    # 7 passed
py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py  # syntax OK
```

Also manually smoke-tested the real hook end-to-end (no-marker block, fresh-marker allow, past-day
allow, commit-message-prose non-match) before writing the formal test suite, and empirically verified
every self-run `/review` finding against the live hook both before and after each fix (not just
trusting the subagent findings' descriptions).

No `package.json` touched (lockfile checks N/A). No suppressions, skip markers, or deleted tests
introduced (greps clean).

## Risk-dimension notes

- **Security:** this PR *is* the security-boundary hardening — closing an agent-reasoning bypass of
  a guard meant to prevent premature journal merges. The no-override-token / fail-closed design is
  the core decision; documented in ADR-096. Self-review found and closed a real bypass (a `git -c`
  flag silently escaping the gate entirely) before merge — see Review findings disposition below.
- **Resilience:** fails closed only on the hook's narrow, already-rare same-day-compose trigger;
  fails open on payload-parse errors (standard convention). 4-hour marker freshness window is
  deliberately generous to cover a realistic multi-project compose with subagent research. A
  timezone-aware marker previously crashed the hook open instead of failing closed — fixed and
  regression-tested.
- **Blast radius:** the hook's trigger condition (git `worktree`/`commit`/`push` + a same-day
  `draft`/`compose` token) never fires for a past-day compose (the nightly routine's normal path)
  or any unrelated git/gh command. Since this hook is globally registered on every Bash call, a
  cheap `today not in command` pre-filter was added to keep that overhead negligible.

Self-run `/review` (2 parallel subagents against the 1407-line diff, correctness/security +
reliability/performance/maintainability) found 3 blocking + 5 non-blocking findings. All empirically
verified against the live hook (not just accepted from the subagents' descriptions) both before and
after fixing. Full review: https://github.com/brownm09/dev-env/pull/671#issuecomment-4928435287

**Blocking (all fixed in `5ea6715`, renumbering carried in `9301eb2`):**
1. `git -c <name>=<value>` (e.g. `-c core.hooksPath=x`) silently bypassed the gate entirely (its
   config value was mistaken for the verb) — fixed via a new `_SKIP_VALUE_FLAGS` set in
   `_verb_and_scan_tokens`.
2. A timezone-aware marker `resolved_at` crashed `is_marker_fresh` with an uncaught `TypeError`,
   failing the hook OPEN instead of closed — fixed by widening the `except` clause.
3. `git commit -am "..."` and glued `-m<value>`/`-F<value>` forms bypassed the message-value
   exclusion, false-blocking innocent commits — fixed via `_MESSAGE_GLUED_RE`/`_MESSAGE_CLUSTER_RE`.

**Non-blocking:**
4. Marker-forgery gap (no provenance binding) — fixed: ADR-096's Limitations section now names it
   explicitly, and `SKILL.md`'s "cannot be reasoned past" claim is qualified accordingly.
5. No cheap pre-filter before the interpreted parse (this hook runs on every Bash call) — fixed:
   `today not in command` short-circuit added.
6. `write_marker`'s shared `.tmp` filename could race two concurrent writers — fixed: per-PID temp
   filename.
7. Triplicated git-tokenization helpers vs. the sibling `pre-tool-use-canonical-mutate-guard.py`
   (the direct root cause of finding #1) — **deliberately not extracted into a shared module in
   this PR.** Fixed the local bug directly instead; a full shared-module extraction would touch the
   well-tested, unrelated canonical-mutate-guard hook for a proportionality-questionable risk/scope
   increase in a PR about a different hook.
8. Hook behavioral guarantees restated across ~6 prose locations with no single source of truth —
   **filed, not fixed in-PR**: [dev-env#683](https://github.com/brownm09/dev-env/issues/683). This
   is a repo-wide convention affecting ~50 hooks, not specific to this PR; fixing one hook's instance
   would not address the actual complaint and a repo-wide documentation refactor is well outside
   this PR's scope.


brownm09:
## Review: fix(journal-compose): mechanically enforce the today-guard

### Summary
Needs revision before merge. The design and test discipline are strong, but empirical testing against the actual hook surfaced a **complete, silent bypass**: any `git commit`/`push`/`worktree` command carrying a `-c <name>=<value>` git-level flag (e.g. `git -c core.hooksPath=x -C <compose-worktree> commit -m "..."`) is misparsed and skips the gate entirely — exit 0, no marker check performed at all, confirmed via direct invocation of the real hook. That defeats the PR's central claim ("cannot be reasoned past") without requiring any reasoning or marker forgery, just an ordinary flag. Two more confirmed defects (a fail-open crash on a valid-but-tz-aware marker, and a false-positive block on the common `git commit -am`/glued `-m<msg>` forms) round out the blocking list. All three have small, mechanical fixes and the existing test suite's structure (pure classification + e2e subprocess layers) already provides the right place to add regression coverage for each.

---

### Blocking Findings

**[correctness]** `claude/scripts/pre-tool-use-journal-compose-force-guard.py` (`_verb_and_scan_tokens`)
`_verb_and_scan_tokens` only recognizes `-C`, `--git-dir`, and `--work-tree` as value-taking flags that precede the git subcommand. A `-c <name>=<value>` (git's per-invocation config override — e.g. `-c core.hooksPath=x`, `-c commit.gpgsign=false`, both ordinary) is not recognized: `-c` is dumped into `scan_tokens`, and the *next* token (the config value, e.g. `core.hooksPath=x`) is then taken as the verb since nothing else claims it. The real verb (`commit`/`push`/`worktree`) is never reached, so `verb not in _GATED_VERBS` and the whole segment is judged not-gated — even though its `-C` value legitimately points at today's compose worktree.
Confirmed by direct invocation with **no marker written at all**: `git -c core.hooksPath=x -C <today's compose-worktree> commit -m "msg"` exits **0** (allowed) where an equivalent command without `-c` correctly exits 2 (blocked, no marker). This is a full, silent bypass of the mechanism this entire PR builds — no reasoning, no marker forgery required, just a common flag. The sibling hook `pre-tool-use-canonical-mutate-guard.py` already handles this exact case via `_parse_git_prefix`'s `_GIT_LEVEL_FLAG_WITH_VALUE = {"-c"}`; this hook reimplemented a weaker, independent parser instead of reusing it.
**Fix:** Handle `-c <value>` (and other valueless git-level flags like `--no-pager`) before verb detection — ideally by extracting the shared git-prefix-parsing logic into `_hookio.py` so both hooks use one implementation (see the related maintainability finding below). Add a regression test asserting a `-c`-prefixed same-day commit is still gated with no marker present.

**[correctness]** `claude/scripts/_journal_compose_force.py:118` (`is_marker_fresh`)
`age = (now - resolved_at).total_seconds()` sits *outside* the `try/except ValueError` that wraps `fromisoformat`. `now` (from `datetime.datetime.now()`) is always naive. A marker whose `resolved_at` is a **timezone-aware** ISO string (e.g. `"...+00:00"` — a completely standard `.isoformat()` output shape, not an exotic input) parses successfully via `fromisoformat` but then raises `TypeError: can't subtract offset-naive and offset-aware datetimes` on the subtraction, which `except ValueError` does not catch.
Confirmed end-to-end: piping a payload at the real hook with such a marker present produces an **uncaught traceback and exit code 1** — not exit 2. Claude Code treats a non-zero-non-2 hook exit as a non-blocking error, so the gated command **proceeds**. This directly inverts the PR's own central, explicitly-stated invariant ("fails CLOSED on a missing/corrupt/stale marker") for exactly the "corrupt marker" case the invariant is supposed to cover. `test_e2e_corrupt_marker_fails_closed` only exercises invalid-JSON (caught by `read_marker`'s `except (OSError, ValueError)` → `None` → block); it does not exercise a valid-JSON, unparseable-timedelta marker.
**Fix:** Move the subtraction inside the `try` and catch `TypeError` alongside `ValueError` (or normalize tz-awareness before subtracting). Add a test with a tz-aware `resolved_at` asserting `is_marker_fresh` returns `False` and the e2e hook still blocks.

**[correctness]** `claude/scripts/pre-tool-use-journal-compose-force-guard.py` (`_verb_and_scan_tokens`, message-value exclusion)
The exclusion of a commit message's *value* from the date-pattern scan only matches the exact tokens `-m`/`--message`/`-F`/`--file` and the `--message=`/`--file=` prefixes. It does not recognize git's glued short-option form (`-m<value>`, `-F<value>`, no space) or the combined-short-option form (`-am <value>`) — all ordinary ways to invoke `git commit`.
Confirmed: `git commit -am "Documented the draft/2026-07-09 today-guard fix"` → `segment_targets_today_compose` returns `True` (blocks, exit 2) even though this is precisely the "commit message merely mentions the pattern as prose" case the hook's docstring, the ADR, and the test suite all state can "never" trigger a match. Because the hook fails closed with **no override token**, this is a real, no-clean-recovery false positive — any commit anywhere on the machine (not just journal-compose's own) whose message happens to discuss today's date and uses `-am`/glued `-m`/glued `-F` gets hard-blocked with a misleading recovery message (it tells the agent to re-run the force-resolve script, which doesn't apply to an unrelated commit).
**Fix:** Extend the message-flag exclusion to recognize glued (`-m<value>`, `-F<value>`) and combined (`-a`+`m` as `-am`) short-option forms. Add regression tests for each (`-am "..."`, `-mvalue`, `-Fvalue`) alongside the existing `-m`/`--message`/`--message=` tests.

---

### Non-Blocking Findings

**[documentation]** `docs/adr/094-journal-compose-mechanical-force-guard.md` (Limitations section) / `claude/skills/journal-compose/SKILL.md`
The marker file has no provenance or integrity binding to the legitimate resolve script — `read_marker` just does `json.load` on a well-known path with no signature or origin check. Nothing stops an agent from authoring `{"force": true, "resolved_at": "<now>"}` at that exact path directly via the `Write` tool or a non-git Bash command (e.g. `python -c "...json.dump..."`) — the guard hook only pattern-matches `git worktree/commit/push` invocations, so neither route is inspected at all. This is a materially easier and less visible bypass than the "hand-typed `--force` fed to the resolve script" route the ADR's own Limitations section already documents and dismisses as low-risk *specifically because* it "requires authoring command text that contradicts the literal `$ARGUMENTS` sitting right above it" — a direct marker write has no such contradicting context to make it visible in review.
**Suggestion:** Extend the Limitations section to name this route explicitly, and soften SKILL.md's "that hook cannot be reasoned past" claim (e.g. "cannot be reasoned past via the documented invocation path") so the stated guarantee matches the actual one.

**[performance]** `claude/scripts/pre-tool-use-journal-compose-force-guard.py` (`command_targets_today_compose`)
This hook is globally registered and runs on **every** Bash call in every project on the machine. `main()` unconditionally runs the full `split_top_level` stack-parse plus per-segment tokenization/regex, even though a match is only possible when today's literal date string appears somewhere in the command (`_DATE_TOKEN_RE` requires an exact match against `today`).
**Suggestion:** Add a cheap `if today not in command: return False` substring pre-check at the top of `command_targets_today_compose` (not inside `segment_targets_today_compose`, so the prose-exclusion tests — which do contain the date — still exercise the real path). Skips the interpreted parse for the overwhelming majority of Bash calls with no behavior change.

**[reliability]** `claude/scripts/_journal_compose_force.py` (`write_marker`)
Uses a fixed `path + ".tmp"` temp filename rather than a per-process-unique one. Two `journal-compose-force-resolve.py` invocations racing for the same date can collide on the shared temp path (one process's `os.replace` firing after the other already renamed the same `.tmp` away, or content tearing). The atomic-rename *does* protect readers (the guard hook never observes a partial write) — the gap is concurrent *writers*. ADR-094's Limitations section already accepts that concurrent same-day composes "share one marker" but frames it as benign, not as a potential resolve-script crash.
**Suggestion:** Use a per-process-unique temp name (e.g. `path + f".{os.getpid()}.tmp"`, or `tempfile.mkstemp` in the same directory) before the atomic rename.

**[maintainability]** `claude/scripts/pre-tool-use-journal-compose-force-guard.py` (module-level git-tokenization helpers)
`_strip_leading_env`, `_first_line`, `_tokenize`, `_git_rest_tokens`, and `_GIT_INVOCATION_RE` are near-verbatim copies of the same-named members in `pre-tool-use-canonical-mutate-guard.py` (the docstring for `_first_line` says as much). This third, independent copy has already diverged and is weaker than the original — it's the direct root cause of the `-c` bypass above, since the sibling's `_parse_git_prefix` already handles `-c <value>` correctly. This repo's dominant convention is aggressive extraction into shared, tested `_*.py` modules (`_hookio`, `_journal_schema`, `_worktree_topology`, `_worktree_canon`); `split_top_level` (the one piece already shared between these two hooks) shows there's already a natural home for this.
**Suggestion:** Extract the shared git-tokenization helpers (at minimum `_git_rest_tokens`, `_tokenize`, `_strip_leading_env`, `_GIT_INVOCATION_RE`, and ideally a shared `_parse_git_prefix`) into `_hookio.py` and import them in both hooks — this removes the triplication and fixes the `-c` bug in one move.

**[maintainability]** `docs/adr/094-journal-compose-mechanical-force-guard.md` / `CLAUDE.md` (Testing entry 53) / `docs/REFERENCE.md`
The same behavioral guarantees (message-flag exclusion, verb gating, fail-closed/4-hour window) are independently restated across roughly six prose locations (hook docstring, `_verb_and_scan_tokens` docstring, ADR Decision + Limitations sections, the REFERENCE.md table row, CLAUDE.md's Testing entry 53) with no single source of truth. Consistent with this repo's existing hook-documentation convention, so not a regression — but it adds another instance of a cost the canonical-mutate-guard's own inline comments already flag about itself, and the three new CLAUDE.md Testing entries (51–53) alone add ~55 dense lines.
**Suggestion:** Not blocking — consider treating the ADR as the single narrative home and trimming REFERENCE.md/CLAUDE.md to short pointers, as a repo-wide pattern rather than something to fix in this PR specifically.

---

### Findings Disposition
Per ADR-028, each finding above must be fixed in this PR or filed as a tracked, linked issue before merge. Disposition will be recorded in a "Review findings disposition" section of the PR body before merge.

---

PR #663, opened 2026-07-09
docs(rules): session-boundary summaries + tile-table/issue-per-tile (PR1 of #652)

brownm09:
Closes #653. Sub-issue of #652 (top-level).

## What

PR1 of 3 — the behavioral backbone. Adds a global **Session Summaries & Tile Tracking** section to `claude/CLAUDE.md` and records the two ADRs behind it. This makes three of the four requested behaviors live *by rule* immediately:

1. **Substantive-stop summary** — close each real stop with **Completed / Context-or-ask / Remaining**; skip trivial exchanges.
2. **Issue-per-tile + end-of-session tile table** — every genuine `spawn_task` tile also gets a tracking issue (durable link + open/closed status); the table renders under the stable `### Tiles spawned this session` heading. Deliberately **overrides** ADR-046's tile≠issue default.
3. **Idle refresher** — documented here; the `UserPromptSubmit` hook that triggers it lands in **PR2 (#655)**. Until then the rule is dormant (no cue → no refresher), not wrong.

Also extends the existing *Capture follow-ups as tiles* bullet with the issue-per-tile requirement.

## ADRs

- **ADR-094** — Tile Tables and Issue-Per-Tile (records the ADR-046 override + the ADR-092 no-loop interaction: `enumeration_recorded` is satisfied session-globally once any `spawn_task` runs, so tile-tracking issues never re-trigger "tile this issue").
- **ADR-095** — Session-Boundary Summaries and Idle Refresher.
- `docs/adr/INDEX.md` updated.

> **Renumber note:** originally drafted as 093/094; **093** was claimed on `main` by #661 (Engineering-Journal Canonical Hijack Guard) while this PR was open, so the branch was rebased onto `origin/main` and the session-boundary ADR renumbered **093 → 095** (tile-tables stays 094). Final: 094 = tile-tables, 095 = session-boundary.

## Testing

Docs-only change (CLAUDE.md rules + 2 ADRs + INDEX; no scripts touched). dev-env has no runtime; verification per the project `## Testing` docs-only guard:

- `grep -n 'date -u' claude/CLAUDE.md` → no matches (clean; nothing in stub/branch-name context).
- All ADR internal links resolve (027, 046, 088, 092, 094, 095 referenced; all present).
- INDEX rows 094/095 added after main's 093; new section present at the intended location.
- No `.py`/shell scripts changed → hook syntax / pyw-stdio / shellcheck suites N/A.
[...]

PR #673, opened 2026-07-09
feat(hooks): idle-refresher UserPromptSubmit hook (PR2 of #652)

brownm09:
## Summary

PR2 of #652 (top-level) — implements the **idle-refresher** hook (sub-issue #655), the one behavior in the initiative that cannot work by rule alone: Claude cannot observe elapsed idle time, so a `UserPromptSubmit` hook measures the wall-clock gap from the transcript and injects an `additionalContext` cue.

On the user's return after an idle gap over the threshold (default **60 min**; per-project override `idle_refresher_minutes` in `.claude/hook-config.json`), `idle-refresher.py` tells Claude to open its reply with a brief refresher — what we were working on, current state, pending to-dos/tiles — before addressing the new prompt. This is the code half of the "Open with a refresher after a long idle gap" rule in the global `claude/CLAUDE.md` **Session Summaries & Tile Tracking** section added by PR1 ([ADR-095](https://github.com/brownm09/dev-env/blob/main/docs/adr/095-session-boundary-summaries-and-idle-refresher.md)).

### Design notes
- **Gap anchor = last assistant turn, not "last record."** The just-submitted user prompt (and a preceding `queue-operation` record) are appended to the transcript around submit time, so "the last record" would be ~now → gap ~0 → the hook would never fire. Anchoring on the last **assistant** record (end of Claude's previous turn) is immune to that, and its absence cleanly means "first prompt / no prior turn → skip." A *resumed* session after a long break fires correctly — a primary intended case. (Verified against a real transcript that assistant records carry a top-level `"timestamp":"...Z"`.)
- **Skips:** automated/XML-prefixed prompts; the first prompt of a session; gap ≤ threshold.
- **Fail-open, stateless, ASCII-only cue** (survives the cp1252 hook-stdout pipe). Mirrors `session-mode-prompt.py` (additionalContext / exit-0) and reuses `_hookutil.load_records` / `find_transcript`.

## Files
- `claude/scripts/idle-refresher.py` — the hook
- `claude/scripts/tests/test_idle_refresher.py` — 14 pure-helper tests
- `claude/settings.json` — registered under `UserPromptSubmit` (after `session-mode-prompt.py`)
- `README.md` + `docs/REFERENCE.md` — Hooks tables + `idle_refresher_minutes` config key
- `CLAUDE.md` — `## Testing` item 51

## Testing
- `py -3 claude/scripts/tests/test_idle_refresher.py` → **Tests: 14 passed, 0 skipped, 0 failed**
  (ISO parse incl. `Z`/naive/microseconds; last-assistant anchor via reverse-scan, incl. falling back past a corrupted trailing timestamp; first-prompt `None`; gap calc; strict-`>` threshold boundary; `idle_refresher_minutes` override + missing/absent-key/malformed/non-positive → 60 default; automated-prefix skip; cp1252-safety of the injected cue.)
- Item 1 syntax gate over `claude/scripts/*.py` — clean (all scripts parse).
- Item 39 crude-command-substring regression (required for any `*.py` change) → **13 passed**; 57 files scanned, 0 unexpected/stale/duplicated.
- **Manual end-to-end smoke of `main()`** (8 cases, all pass): 3h idle → emit; 1 min → silent; automated prompt → silent; first prompt (no assistant record) → silent; `idle_refresher_minutes:5` + 10-min gap → emit; 10-min gap at default 60 → silent; corrupted trailing timestamp with a valid 3h-old earlier one → emit; `idle_refresher_minutes:0` → falls back to default, 10-min gap → silent.
- Suppression / test-integrity greps (incl. pytest idioms) — clean; no new suppressions, skip markers, or deleted tests. Coverage gate: the new behavior is fully covered by the new test.
- N/A: `baseline_test_failure_tracking` (off for dev-env — no Jest); lockfile-sync (no `package.json` change).

## ADR / docs
- **No new ADR** — this implements the decision already recorded in **ADR-095** (PR1, #663). Doc-reconciliation satisfied: a new event-driven hook → README + REFERENCE Hooks tables + a `## Testing` item, per the Documentation Maintenance table.

`/review --post-comment` ([comment](https://github.com/brownm09/dev-env/pull/673#issuecomment-4928419070)) ran two parallel opus subagents (correctness/security; reliability/performance/maintainability). 0 blocking, 4 non-blocking findings — both subagents independently converged on the same top issue (the mtime fallback), verified empirically against a real transcript before fixing:

- **[maintainability]** mtime fallback was dead/misleading code (its own premise implied mtime ≈ now at hook-fire time) — **fixed in `64ad1f4`**, removed along with the now-unused `has_prior_assistant_turn` helper.
- **[performance]** `last_activity_epoch` did a full forward scan instead of a reverse scan with early return — **fixed in `64ad1f4`**, same semantics, less work.
- **[correctness]** `load_threshold_minutes` had no lower-bound validation, so a configured `0`/negative `idle_refresher_minutes` would fire on nearly every prompt — **fixed in `64ad1f4`**, falls back to the 60-min default.
- **[performance]** `_hookutil.load_records` (a shared module used by 4 other hooks) reads+parses the entire transcript on every call — genuinely out of scope for this PR (fixing it well means either diverging from the shared convention or a separate PR touching the shared module) — **filed [dev-env#679](https://github.com/brownm09/dev-env/issues/679)**.

Two new regression tests added for the first three fixes (14 total, up from 12); all re-verified via the full test suite, syntax gate, and an expanded 8-case end-to-end smoke run.

Closes #655


brownm09:
## Review: feat(hooks): idle-refresher UserPromptSubmit hook (PR2 of #652)

### Summary

Merge with minor changes — actually already applied. Two parallel opus subagents (correctness/security; reliability/performance/maintainability) independently converged on the same top finding: the mtime fallback in `main()` could never usefully fire, because the module's own design premise (the user's prompt is written to the transcript around submit time) implies the file's mtime is already ~now at hook-fire time — the identical "gap collapses to zero" problem the primary design already avoids by anchoring on the last assistant record instead of "the last record." I verified this empirically against a real transcript before fixing rather than taking the subagents' word for it (the `user` record lands within ~100ms of the preceding `queue-operation`). All four findings below were low-severity (fail-safe, no crash/data-loss/security impact) and have already been fixed in commit `64ad1f4` or filed as a follow-up issue. No blocking or security findings. DIFF_SIZE (460 lines across two commits) is appropriately scoped for a single hook + its tests.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/idle-refresher.py` (mtime fallback, since removed)
The original code fell back to `os.path.getmtime(path)` as a "last-resort proxy" for idle time when an assistant record existed but its timestamp was unparseable. By the module's own documented premise — the transcript is written at ~submit time, which is *why* the design anchors on the last assistant record instead of "the last record of any type" — the file's mtime at hook-fire time is itself ~now, so this fallback's computed gap would always collapse to ~0 and could never cross the threshold. It failed safe (a missed refresher in an already-rare corrupted-timestamp edge case, never a spurious fire), but it was dead, misleading code a future maintainer would trust at face value.
**Fix:** Removed the fallback and its now-unused `has_prior_assistant_turn` helper in commit `64ad1f4`; `main()` now simply skips when no assistant record has a parseable timestamp. Two regression tests added.

**[performance]** `claude/scripts/idle-refresher.py` — `last_activity_epoch`
Scanned all records front-to-back, overwriting the running answer each time it saw a valid assistant timestamp, to end up with the final (chronologically last) one. Transcripts are append-only chronological, so the answer is almost always at the tail; a full forward scan does maximal work for a result reverse iteration finds in O(1) in the common case.
**Fix:** Reverse-scan with early return in commit `64ad1f4` — identical semantics (falls back to an earlier valid timestamp if the trailing assistant record's own timestamp is corrupted), same result, less work. Locked in with a new regression test (`test_last_activity_skips_corrupted_trailing_timestamp`).

**[correctness]** `claude/scripts/idle-refresher.py:123` (now ~124) — `load_threshold_minutes`
No lower-bound validation on the configured `idle_refresher_minutes`. A `0` or negative value makes `threshold_seconds <= 0`, so `should_refresh`'s `gap > threshold_seconds` check fires on essentially every prompt — turning an occasional orientation cue into per-turn injected noise. Mirrors `turn-count-hook.py`'s equally-unvalidated `turn_threshold` (a pre-existing convention, not a regression), but the cheap fix is worth it since it prevents a realistic typo from degrading every turn.
**Fix:** `load_threshold_minutes` now falls back to `DEFAULT_MINUTES` for any parsed non-positive value, in commit `64ad1f4`. Two new test cases (`0` and `-5`) plus an end-to-end smoke case.

**[performance]** `claude/scripts/idle-refresher.py` — `_read_records` → `_hookutil.load_records`
On every user prompt submission this hook reads the session's entire JSONL transcript into memory and parses every line, even though it only needs the last assistant record's timestamp. This is inherent to the shared `_hookutil.load_records` utility (already used by 4 other hooks), not something this PR's design introduces independently — fixing it well means either diverging from the shared convention for this one caller, or modifying the shared module itself, which has a much larger blast radius and needs its own dedicated PR and test coverage.
**Fix:** Filed [dev-env#679](https://github.com/brownm09/dev-env/issues/679) rather than fixed in-PR — genuinely out of scope per the ~75% scope-growth guard.

---

### Findings Disposition

- mtime fallback (dead/misleading code) — fixed in `64ad1f4`
- `last_activity_epoch` forward-scan perf — fixed in `64ad1f4`
- `load_threshold_minutes` no lower-bound validation — fixed in `64ad1f4`
- `_hookutil.load_records` full-transcript-read cost — filed [dev-env#679](https://github.com/brownm09/dev-env/issues/679)

---

*Reviewed by `claude-opus-4-8` (two parallel subagent passes) via Claude Code*

PR #674, opened 2026-07-09
feat(hooks): add tile-table trigger to stop-tile-enumeration-gate.py (PR3 of #652)

brownm09:
## Summary

PR3 of [dev-env#652](https://github.com/brownm09/dev-env/issues/652) — sub-issue [#656](https://github.com/brownm09/dev-env/issues/656).

Extends `claude/scripts/stop-tile-enumeration-gate.py` with a **third**, fully independent Stop-hook trigger alongside the existing merged-PR ([ADR-088](docs/adr/088-state-keyed-tile-enumeration-gate.md)) and dangling-created-issue ([ADR-092](docs/adr/092-dangling-issue-tile-enumeration-gate.md)) triggers:

- **Trigger 3 (ADR-094 addendum):** if `spawn_task` tiles were spawned this session but no assistant message carries the stable heading `### Tiles spawned this session` (the marker convention PR1/[#663](https://github.com/brownm09/dev-env/pull/663) introduced in `claude/CLAUDE.md` → Session Summaries & Tile Tracking), block the stop (exit 2) with a reminder to post the table.
- Reuses the gate's existing machinery: `_SPAWN_TASK_RE`/spawn detection, `_content_items`, `iter_bash_calls`, the once-per-session sentinel, and the `stop_hook_active` loop guard.
- **Key asymmetry, by design:** a spawned tile satisfies `enumeration_recorded` (so it silently *resolves* triggers 1/2), but does **not** by itself satisfy trigger 3 — the table is a stricter, separate bar. A session that merges a PR, spawns a tile, and never emits the table sees trigger 1 resolve quietly while trigger 3 still fires.
- Fail-open (exit 0 on any error); ASCII-safe reminder text (cp1252 pipe); pre-filter extended with a third OR-branch (bare `"spawn_task"` substring, matching the detector's own namespace-agnostic philosophy — see review findings below).

## Test plan

- [x] `py -3 claude/scripts/tests/test_stop_tile_enumeration_gate.py` — **112 passed, 0 skipped, 0 failed** (88 pre-existing + 24 new: 14 pure-helper, 4 interaction, 6 e2e)
- [x] `py -3 -c "import ast,sys; [ast.parse(...) for f in sys.argv[1:]]" claude/scripts/*.py` — hook-script syntax check, OK
- [x] `py -3 claude/scripts/tests/test_no_crude_command_substring_checks.py` — AST substring-check regression gate, 13 passed
- [x] Suppression grep (`ts-ignore`/`!`/etc.) — clean, no matches
- [x] Test-integrity grep (skip markers, deleted tests, lowered thresholds) — clean, no matches
- [x] Two **pre-existing** e2e tests (`test_e2e_merged_with_enum_allows`, `test_e2e_dangling_issue_with_enum_allows`) updated: a bare spawn no longer reaches exit 0 once trigger 3 exists, so both now also emit the table heading — their original intent (proving triggers 1/2 resolve on enumeration) is preserved via a genuinely fully-compliant session, not by weakening the new trigger.
- [x] Post-review: added a dedicated regression test (pure + e2e) using a fabricated non-standard MCP namespace, pinning the pre-filter/detector soundness fix below.

No automated tests were skipped or deleted; no coverage thresholds touched. No `package.json` changed (no lockfile-sync check needed).

## Documentation

- `README.md` Stop-hook row and `docs/REFERENCE.md` hook entry + narrative — updated to describe all three triggers.
- `CLAUDE.md` → `## Testing` item 48 — extended with the third trigger's test-coverage description, including the review-driven fixes.
- ADR: appended a dated addendum to `docs/adr/094-tile-tables-and-issue-per-tile.md` (the ADR whose own Decision section forward-referenced #656) documenting the detection design, the enumeration-vs-table asymmetry, the pre-filter fix, and an accepted-limitations note. No `docs/adr/INDEX.md` change needed (ADR-094 already indexed).

Full review: https://github.com/brownm09/dev-env/pull/674#issuecomment-4928413899

- **Pre-filter narrower than the detector it gates** (correctness, latent soundness gap independently found by two review passes) — **fixed** in `735c7c1`: switched the pre-filter from the exact fully-qualified tool name to the same bare `"spawn_task"` substring `_SPAWN_TASK_RE` matches; added regression tests using a fabricated non-standard namespace.
- **`session_spawned_tiles`/`enumeration_recorded` duplicated tool_use-detection loop** (maintainability) — **fixed** in `735c7c1`: `enumeration_recorded` now delegates to `session_spawned_tiles`.
- **Shared once-per-session sentinel can mask trigger 3 for a tile spawned in a later, separate turn** (reliability) — **filed** [dev-env#677](https://github.com/brownm09/dev-env/issues/677); documented as an accepted, known limitation in the `docs/adr/094-tile-tables-and-issue-per-tile.md` addendum (the same class of tradeoff ADR-088 already documents for the two existing triggers; a proper fix would restructure shared gating logic with correctness implications beyond this PR's scope).

Closes #656


brownm09:
## Review: feat(hooks): add tile-table trigger to stop-tile-enumeration-gate.py (PR3 of #652)

### Summary

Merge with the fixes already applied in this PR. The diff adds a third, well-isolated trigger to `stop-tile-enumeration-gate.py`, mirroring the existing merged-PR/dangling-issue trigger shape closely and backed by 112 tests (0 failed). Two parallel review passes (correctness/security; reliability/perf/maintainability) independently converged on the same root-cause finding — a real, latent soundness gap between the new pre-filter and the detector it gates — which has been fixed and regression-tested in a follow-up commit on this branch, along with a related DRY finding. One architectural finding (the shared once-per-session sentinel can mask a later, separate-turn trigger fire) was judged out of scope for this PR — it's the same class of limitation ADR-088 already documents and accepts for the two existing triggers, and properly fixing it means restructuring shared gating logic with correctness implications for those existing triggers, not just the new one — so it's documented as an accepted limitation in the ADR-094 addendum and tracked as [dev-env#677](https://github.com/brownm09/dev-env/issues/677).

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/stop-tile-enumeration-gate.py` — pre-filter was narrower than the detector it gates (FIXED)
The new pre-filter checked the exact fully-qualified tool name `mcp__ccd_session__spawn_task`, while the real detector (`_SPAWN_TASK_RE`) is a deliberately namespace-agnostic bare `spawn_task` match — its own pre-existing comment says "Bare verb so any namespacing hits." The fully-qualified check was a strict subset of what the detector can match: a spawn recorded under any other MCP namespace would satisfy `session_spawned_tiles` but be silently skipped by the narrower pre-filter before `evaluate_tile_table` ever ran. Not currently triggerable (all recorded spawns in this environment use the standard prefix today), but a real, latent gap that would silently disable trigger 3 for good if the MCP namespace hosting `spawn_task` were ever renamed or rehosted — independently identified from two different angles (a correctness lens and a maintainability lens) by the two review passes.
**Fix:** Switched the pre-filter to the same bare `"spawn_task"` substring `_SPAWN_TASK_RE` matches, accepting the same bounded extra-reparse cost the pre-existing `"merged"` pre-filter branch already carries. Added a dedicated regression test (pure `session_spawned_tiles`/`enumeration_recorded` case + a full end-to-end subprocess case) using a fabricated non-standard namespace, so the soundness property is pinned, not just restored.

**[maintainability]** `claude/scripts/stop-tile-enumeration-gate.py` — `session_spawned_tiles` duplicated `enumeration_recorded`'s tool_use loop (FIXED)
`session_spawned_tiles` re-implemented, line for line, the tool_use half of `enumeration_recorded`'s existing loop (`for rec... type=="assistant"` / `for item in _content_items` / the same `_SPAWN_TASK_RE.search(...)` check). The two could silently drift, which is exactly what the finding above shows already almost happened between the detector and its own pre-filter.
**Fix:** `enumeration_recorded` now delegates its tool_use check to `session_spawned_tiles` (`if session_spawned_tiles(records): return True`, then only the text-marker loop remains), collapsing the duplication to one source of truth. A dedicated test exercises the delegation directly through a non-standard-namespace spawn.

**[reliability]** `claude/scripts/stop-tile-enumeration-gate.py` — the shared once-per-session sentinel can mask trigger 3 for a tile spawned in a later, separate turn (FILED, documented as accepted limitation)
All three triggers share one sentinel. If trigger 1/2 fires or resolves at an earlier Stop (e.g., a PR merges with no tile spawned yet), the sentinel is set and suppresses re-evaluation of *all three* triggers for the rest of the session — including trigger 3, whose own condition (a later, separate-turn tile spawn with no table) hadn't happened yet when the sentinel was set. This does not affect the common compliant flow, where the reminder text prompts an inline spawn-and-table response within the same blocked-Stop continuation (governed by `stop_hook_active`, checked before the sentinel) — the gap is specifically a genuinely later, separate turn.
**Disposition:** This is architecturally the same class of limitation ADR-088 already documents and accepts for triggers 1/2 ("session-global enumeration... not re-flagged for the second [merge]"), and the task that produced this PR explicitly instructed reusing the gate's existing once-per-session sentinel machinery. Splitting per-trigger sentinels is a materially larger change to `main()`'s shared gating logic, with correctness implications for the two *existing* triggers, not just the new one — judged out of scope for this PR rather than a rushed redesign. Documented as an accepted, known limitation in the `docs/adr/094-tile-tables-and-issue-per-tile.md` addendum. Filed: [dev-env#677](https://github.com/brownm09/dev-env/issues/677).

---

### Findings Disposition

- Pre-filter/detector soundness gap — **fixed** in `735c7c1`
- `session_spawned_tiles`/`enumeration_recorded` duplication — **fixed** in `735c7c1`
- Shared-sentinel cross-trigger masking — **filed** [#677](https://github.com/brownm09/dev-env/issues/677), documented as an accepted limitation in the ADR-094 addendum

---

PR #676, opened 2026-07-09
[docs] Fold cross-repo worktree tool scoping gaps into Git Workflow rules

### LINKED ISSUE #627: Document EnterWorktree cwd-scoping + Bash cwd non-persistence after worktree tool calls in CLAUDE.md
## Context

Discovered during the dev-env#616 session (2026-07-08, PR #623): the session's primary working directory was `lifting-logbook`, but the task was entirely in `brownm09/dev-env`.

## Gap 1 — EnterWorktree only ever targets the session's own cwd repo

Calling `EnterWorktree({name: "..."})` from a `lifting-logbook` session created a worktree **of lifting-logbook**, not of dev-env, even though the intent was clearly to work in dev-env. `EnterWorktree` has no parameter to target a different repo than the session's current working directory — this isn't documented anywhere in the global `CLAUDE.md`'s worktree guidance, which currently assumes the session is already operating in the correct repo.

Recovery: `ExitWorktree({action: "remove"})` the mistaken worktree, then fall back to a plain `git worktree add <path> -b <branch> origin/main` issued directly against the target repo's absolute path.

## Gap 2 — Bash tool cwd does not reliably persist a `cd` after worktree tool calls

After using `EnterWorktree`/`ExitWorktree` at least once in a session, a plain `cd <path>` in one Bash tool call did not carry over to the next Bash call — observed directly as `Shell cwd was reset to <original session cwd>`. Every subsequent command touching the manually-created worktree needed its own explicit `cd <path> &&` prefix for the rest of the session.

## Suggested fix

Add a short note to the global `claude/CLAUDE.md`, near the existing "Worktree Bash commands must not `cd` to the main repo" bullet in the Git Workflow section:

1. Before calling `EnterWorktree`, confirm the task's target repo matches the session's own primary working directory (shown in the system prompt's "Primary working directory" line). If it doesn't, skip `EnterWorktree` and use `git worktree add` directly against the target repo's path instead.
2. Once any worktree tool has been invoked in a session, don't rely on Bash cwd persistence — prefix every subsequent Bash command touching a worktree with an explicit `cd`, even for back-to-back commands.

Full write-up and the originating incident: session memory `feedback_cross_repo_worktree_tool_scoping.md` (private, not committed) and the dev-env#616 session's engineering-journal stub (`sessions/dev-env/2026-07-08_110125.stub.md` in `brownm09/engineering-journal`, PR history for `draft/2026-07-08`).

brownm09:
## Summary

Closes #627 by folding its three documented gaps into the global `claude/CLAUDE.md` Git Workflow section:

1. **Gap 1 — `EnterWorktree` targets the session's primary repo only.** Already had a dedicated bullet from #610 (a separate, earlier-filed issue covering the same mechanic). This PR extends that bullet with an explicit proactive check: compare the task's target repo against the session's `Primary working directory` line *before* calling `EnterWorktree`, rather than discovering the mismatch after the fact.
2. **Gap 2 — Bash cwd does not reliably persist a plain `cd` once a worktree tool has fired.** New content: after `EnterWorktree`/`ExitWorktree` has run at least once in a session, every subsequent Bash command touching a manually-created worktree needs its own explicit `cd`/`-C` — cwd persistence across separate Bash calls can no longer be assumed.
3. **Gap 3 — `-C`/`cd` does not chain across `&&`/`;`/`||`.** New content (added to #627 via comment on 2026-07-09): in a multi-command Bash invocation, a leading `-C`/`cd` scopes only the command it's attached to. `git -C <repo> fetch origin && git worktree add ...` silently runs the second command against the *actual* shell cwd, not the `-C` target, if they differ — each command needs its own explicit scoping.

Both new/extended bullets also recommend sanity-checking the result of any `worktree add`/`checkout`/`clone` (`git remote -v`, or a commit-hash comparison against the target repo's `origin/<default-branch>` tip) before trusting it — the exact verification this PR's own drafting session performed on its own worktree.

**Note on Gap 1:** while writing this PR I discovered `#610` (closed, merged via PR #608) had already independently documented Gap 1's mechanic under its own issue number, from a different incident. Rather than duplicate that bullet, I extended it in place and added the new bullet for Gaps 2–3 immediately after it. See the commit message for the full rationale, including why no ADR was added (ADR-066 covers a different, single-repo concern; #610 set precedent for a plain bullet + issue link).

## Test plan

- [x] Docs-only guard (dev-env `## Testing` item 4): `grep -n 'date -u' claude/CLAUDE.md` — zero matches, so nothing to review.
- [x] `git diff` reviewed — only `claude/CLAUDE.md` touched, two bullets (one extended, one new).
- [x] ADR-warrant check: N/A — extends existing #610 precedent (bullet + issue link, no ADR); ADR-066 is scoped to a different (single-repo) concern.
- [x] Doc-reconciliation checkpoint: N/A — no skill/hook/script/routine added, removed, renamed, or behavior-changed.

Tests: no automated test suite applies to a docs-only `CLAUDE.md` change; see the docs-only guard above.

Per `/review` ([comment](https://github.com/brownm09/dev-env/pull/676#issuecomment-4928485353)): 0 blocking, 1 non-blocking.

- **[documentation] `claude/CLAUDE.md:98-99`** — the new bullet's "leading `cd <path> &&`" example read in tension with the immediately preceding bullet's "never `cd` there." Fixed in `f2bbe9d` — added a clause distinguishing the single-invocation-scoped `cd &&` (dies with its chained command) from the persistent `cd` the bullet above warns against (lingers across separate Bash calls).


brownm09:
## Review: [docs] Fold cross-repo worktree tool scoping gaps into Git Workflow rules

### Summary

Small (3-line), well-scoped, docs-only change that folds the three gaps documented in #627 into
`claude/CLAUDE.md`'s Git Workflow section. Content is accurate and traceable to the originating
incidents — verified the Gap 3 example command, the `Primary working directory` system-prompt
citation, and the claim that `-C` is a per-invocation git flag (not a shell-level scope) all hold
up. One non-blocking clarity nit below: the new bullet's "leading `cd <path> &&`" concession reads
in tension with the immediately preceding bullet's "never `cd` there," even though the two are
describing different things. Ready to merge once that's tightened.

---

### Non-Blocking Findings

**[documentation]** `claude/CLAUDE.md:98-99`
Line 98 ends with an unqualified imperative — "interact with it via explicit paths or `git -C
<path>` — never `cd` there." Line 99, immediately below, then endorses "a leading `cd <path> &&`"
as an equally valid scoping mechanism alongside `-C`. These aren't actually in conflict — line 98's
"never `cd` there" means don't let a bare `cd` persist across separate Bash calls (the split-brain
failure mode ADR-066 exists for); line 99's `cd <path> &&` is a single-invocation-scoped prefix,
functionally equivalent to `-C` for that one command. But a session skimming under time pressure —
exactly the failure mode this whole section exists to prevent — could read the two back-to-back and
either avoid the legitimate `cd &&` idiom unnecessarily, or miss which `cd` pattern is actually
dangerous.
**Suggestion:** Add a short clause to line 99 distinguishing "a bare `cd` that outlives this command"
(line 98's target) from "a `cd <path> &&` prefix scoped to a single chained command" (line 99's own
example) — or simplify line 99 to recommend `-C` exclusively and drop the parenthetical `cd`
alternative, since `-C` already covers the same case without needing the distinction spelled out.

---

### Findings Disposition

Per ADR-028, the one non-blocking finding above must be fixed in this PR or filed and linked before
merge. Will fix in-PR immediately following this review (small wording addition to line 99) and
record the disposition here before merge.

---

PR #691, opened 2026-07-10
docs(claude-md): remove stale test-count tally from Testing item 48

brownm09:
## Summary
- Removes the stale "76 tests total, up from 39 pre-ADR-092" tally line from `CLAUDE.md`'s Testing item 48 (covering `test_stop_tile_enumeration_gate.py`)
- The count had already drifted stale once before this PR (dev-env#650 added ~12 tests without updating it), and the line wasn't even at the end of the item — a third trigger's worth of tests (the ADR-094 addendum, dev-env#656) was documented *after* it with no update either
- Verified the actual current count by running the test file directly: **112**, not 76
- Removed the tally rather than correcting it to 112, per the issue's own suggested alternative — a hand-maintained running total buried mid-paragraph in a heavily and frequently edited item has now proven unmaintainable twice over; the item's prose already describes each addition's scope in detail, so the count added no information not already present

## Testing
Ran the specific test file this documentation item describes, from the repo root:
```
py -3 claude/scripts/tests/test_stop_tile_enumeration_gate.py
```
Tests: 112 passed, 0 skipped, 0 failed (6.2s)

Docs-only change (no code/script modified). Suppression-policy and test-integrity greps were run against `git diff origin/main` and are both clean. `baseline_test_failure_tracking` is not enabled for this repo (no `.claude/hook-config.json` baseline flag) — N/A.

## Risk-dimension audit
1. **Testing** — N/A, docs-only; verified via the test run above.
2. **Observability** — N/A, no runtime behavior changed (dev-env has no application runtime; see project `## Observability`).
3. **Security** — N/A, no code touched.
4. **Resilience/failure modes** — N/A.
5. **Performance** — N/A.
6. **Data integrity/migrations** — N/A.

Closes #681

PR #690, opened 2026-07-09
[docs] Clarify what counts as a qualifying citation in /review Step 2f

brownm09:
## Summary

`/review` Step 2f (ADR-Warrant Check) lets a PR satisfy `adr_warrant=written` by citing a prior PR
that "already covers the decision" — but never defined what "covers" means. dev-env#676 hit this
ambiguity directly: it cited PR #608 as precedent, but #608 covered a *different* gap in the same
issue family, not #676's actual new content. The reviewing session resolved it via one-off
judgment (precedent for the same class of change was treated as sufficient), with no codified rule
to make that call consistent next time.

This PR adds that rule to `claude/skills/review/SKILL.md` Step 2f: a citation satisfies `written`
only if the cited PR establishes precedent for the **same class of change** (same file/section,
same category of rule) **and** the current PR body states *why* that precedent applies — surface
similarity alone (e.g. same file, different category of rule) does not qualify.

## Where the clarification lives

Step 2f (and its `written`-via-citation language) was originally authored by
[ADR-083](docs/adr/083-auto-merge-checkpoint-gate.md)'s Decision section, not ADR-011 (which
defines the ADR-warrant-check *concept* but predates Step 2f entirely — grepped to confirm ADR-011
never mentions the citation mechanism). So the clarification is recorded as a new dated addendum on
ADR-083, following this repo's own established convention on that exact file (4 prior dated
addenda already there: 2026-07-06, 07-07, 07-08 ×2) — not a new ADR number.

## Doc reconciliation (Step 2b)

Touching `claude/skills/review/SKILL.md` requires README.md or docs/REFERENCE.md in the same diff.
Both are included — and while reconciling, found `docs/REFERENCE.md`'s `/review` section was
already stale independent of this change: it said "three pre-checks" and only described Steps
2b–2e, omitting Step 2f (ADR-Warrant Check) entirely. Fixed the count and added the missing bullet
in the same edit; made the same one-line addition to README.md's Skills table for consistency.

## ADR-warrant check (Step 2f, dogfooding this PR's own rule)

Criterion 1 applies (touches `claude/skills/review/SKILL.md`, under `claude/`). Satisfied via
direct inclusion, not citation: the ADR-083 addendum above is a qualifying amended ADR present in
this diff. `adr_warrant=written`.

## Testing

dev-env's `## Testing` section in `CLAUDE.md` itemizes 50 required checks, all scoped to
[...]

PR #645, opened 2026-07-09
docs(claude-md): document PostToolUse hook firing on gh pr create

brownm09:
## Summary
- The "## GitHub Project" section's Workflow/Fallback paragraphs described `post-tool-use.py` as firing only on `gh issue create`. It fires identically on `gh pr create` (confirmed directly in `post-tool-use.py`: `is_issue_create_command`/`is_pr_create_command` lines 91-98 feed a single shared gate in `main()` lines 542-555, and the add-to-project + `format_reminder` flow at lines 590-599 is byte-identical between the two — the only difference anywhere in the script is one interpolated word, `item_type = "Issue" if is_issue_create else "PR"` at line 566). This is also the documented, accepted design in [ADR-023](docs/adr/023-generic-required-fields-issue-hook.md).
- Updates the Workflow paragraph, Fallback paragraph, and the fallback code block's comment/placeholder to mention `gh pr create` alongside `gh issue create`, so a future session isn't caught off guard by PR-creation hook output or skips setting Impact/Why on a PR's project item.
- No behavior change — `post-tool-use.py` itself is correct and unchanged; only the CLAUDE.md prose needed to catch up.

Closes #643

## ADR-warrant check
Not warranted — see `docs/adr/011-adr-warrant-check.md`. Checked against all four criteria:
1. Changes a rule/hook/skill/settings value documented in `claude/` — No, corrects prose to match existing, unchanged hook behavior.
2. Introduces or restructures a directory under `claude/` — No.
3. Establishes or changes a workflow rule other CLAUDE.md files reference — No, no new rule.
4. Rationale hard to recover from `git log` alone — No, fully recoverable: the hook already treats both identically per the existing, accepted ADR-023.

Same reasoning pattern as the precedent PR #633 for an analogous one-line CLAUDE.md prose addition.

## Doc reconciliation
Checked `README.md` (line 82) and `docs/REFERENCE.md` (line 239) — both already describe the hook's trigger generically as "issues/PRs" / "`gh issue create` or `gh pr create`". No updates needed there; the staleness was isolated to the root `CLAUDE.md` GitHub Project section.

## Testing
[...]

PR #693, opened 2026-07-10
fix(hooks): split the tile-enumeration gate's shared sentinel per trigger

brownm09:
## Summary

- Fixes dev-env#677: `stop-tile-enumeration-gate.py`'s three triggers (merged-PR, dangling-issue,
  tiles-without-a-table) shared ONE once-per-session sentinel, so whichever trigger fired or resolved
  FIRST silently suppressed evaluating the other two for the rest of the session — including one
  whose own condition (e.g. a tile spawned in a later, separate turn) hadn't even occurred yet.
- Splits the sentinel into three independent per-trigger files (`-pr-`/`-issue-`/`-table-` suffixes on
  the same `SENTINEL_PREFIX`). `main()` now checks all three up front — skipping the transcript read
  entirely only when all three are already set — and evaluates only the still-open triggers, marking
  each independently.
- The cheap pre-filter is also gated per trigger (added during review — see disposition below): a
  transcript signal like `"merged"` never disappears once written, so without this gate a session
  with an early merge would force a full reparse on every remaining Stop even after that trigger
  resolved.
- No pure evaluator function (`evaluate`, `evaluate_issues`, `evaluate_tile_table`,
  `session_merged_prs`, etc.) changed — only `main()`'s sentinel bookkeeping and the two small I/O
  helpers it uses.
- Adds ADR-097 documenting the design (forward-referenced by ADR-094's addendum as this exact
  follow-up — dev-env#677 was filed during the PR #674 review that landed the third trigger).

## Testing

Ran `py -3 claude/scripts/tests/test_stop_tile_enumeration_gate.py` from the repo root. All 112
pre-existing tests pass **unmodified** (traced the three existing "sentinel suppresses a second fire"
e2e tests by hand first — each fixture only ever satisfies one trigger's precondition, so the other
two evaluate to their own no-op regardless of sentinel state). Four new tests cover the fix directly:
a two-turn simulation reproducing the dev-env#677 bug itself (trigger 1 fires in turn 1; a tile
spawned with no table in turn 2 must still fire trigger 3); that a partial (merge-only) session sets
ONLY the `pr-` sentinel file on disk, not `issue-`/`table-`; that a fully-compliant session sets all
three sentinel files and stays allowed on a second Stop; and a three-turn sequence (added during
review) proving a resolved trigger's stale signal text doesn't block detection of a later, distinct
trigger.

```
Tests: 116 passed, 0 skipped, 0 failed
```

Also ran, required for any `claude/scripts/*.py` change:
- Hook-script syntax check (`ast.parse` over `claude/scripts/*.py`) — clean.
- `py -3 claude/scripts/tests/test_no_crude_command_substring_checks.py` — 13 passed, 0 failed (60
  files scanned, 0 unexpected/stale/duplicated offenses).

Suppression-policy and test-integrity-policy greps against `origin/main` — both clean, no matches.
(The standard test-integrity greps target JS/TS idioms; this diff's test changes are Python — see the
review's language-scope finding and its disposition below for the manual verification performed.)

## Documentation

- New ADR: `docs/adr/097-per-trigger-tile-gate-sentinels.md`, added to `docs/adr/INDEX.md`.
- Updated dev-env `CLAUDE.md` → `## Testing` item 48 to describe the fix and the four new tests.
  (Per PR #691's just-established convention, this PR does **not** add a hand-maintained "N tests
  total" tally back into item 48 — an earlier draft of this PR briefly did, before rebasing onto
  origin/main picked up #691 and this was corrected.)
- Updated `docs/REFERENCE.md`'s tile-checkpoint section with a "Per-trigger sentinels" note.

Reviewed via `/review` (two parallel Opus subagents — correctness/security found nothing; reliability/
performance/maintainability found two issues, both addressed) plus the skill's own Step 2e
test-integrity language-scope check. Full review: see PR comment.

1. **Test-integrity language-scope caveat** (Step 2e) — verified manually; no violation found (no
   skip markers, no deleted tests, no config thresholds in this repo's Python convention, no
   test-input-shaped implementation branches). No further action needed.
2. **Pre-filter not gated per trigger** — fixed in `08c93a4`.
3. **Four parallel per-trigger structures in `main()`** — filed as
   [dev-env#696](https://github.com/brownm09/dev-env/issues/696); a fully generic loop was drafted and
   rejected as a net complexity wash given the triggers' heterogeneous shapes (see the issue and
   ADR-097's Alternatives section for the full reasoning). Deferred by design, not currently blocking.

## Note on canonical checkout staleness (unrelated, flagging separately)

While starting this work, the canonical dev-env checkout (`~/Git/dev-env`) was found to be ~45 merged
PRs behind `origin/main` (local tip PR #645 vs. origin's PR #690), despite the session-start sync hook
reporting success. This PR's work was done in a worktree branched directly from `origin/main`, so it
is unaffected. Filed [dev-env#694](https://github.com/brownm09/dev-env/issues/694) to investigate
separately.

Closes #677


brownm09:
## Review: fix(hooks): split the tile-enumeration gate's shared sentinel per trigger

### Summary
Ready to merge. This is a well-scoped, well-tested bug fix: `stop-tile-enumeration-gate.py`'s three Stop-hook triggers previously shared one sentinel file, so any trigger firing/resolving first silently suppressed evaluating its siblings for the rest of the session. The fix splits the sentinel per trigger with no change to any pure evaluator function, and all 112 pre-existing tests pass unmodified. Two non-blocking findings surfaced during review (both already addressed below — no outstanding blockers). Diff is appropriately scoped for a single, focused fix (436 lines, entirely within one hook, its test file, and documentation).

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/stop-tile-enumeration-gate.py` (Step 2e — test-integrity language-scope check)
The diff's test changes are entirely in Python (`test_stop_tile_enumeration_gate.py`), while the standard suppression/test-integrity greps in `claude/CLAUDE.md` target JS/TS idioms (Jest/Vitest/Mocha). A clean result from those patterns is not itself evidence the integrity policy was satisfied for this Python diff.
**Suggestion:** Manually verified instead: no skip markers were added (no test was excluded from the `tests = [...]` registration list), no existing test function or list entry was deleted (test count went from 112 → 116, additive only, confirmed via full suite run before and after), no config/threshold files exist in this repo's Python test convention to lower, and no implementation branch is shaped around a specific test's input value (the per-trigger sentinel mechanism is generic over any `session_id`). No action needed beyond this note.

**[performance]** `claude/scripts/stop-tile-enumeration-gate.py` — pre-filter (fixed in this PR)
The cheap pre-filter (`"merged"` / `gh issue create` / `spawn_task` substring/regex check) was not gated on each trigger's own resolved state. Because a transcript signal never disappears once written, a session with an early PR merge — the single most common trigger — would carry `"merged"` in its transcript for the rest of the session, forcing a full reparse on *every* remaining Stop even after that trigger resolved. This is the common case (merge, then keep working), not the "narrow window" the ADR's first draft described.
**Fix:** Gated each pre-filter clause on `already_done[trigger] or <original check>`, restoring the parse-skipping fast path the instant every trigger with a live transcript signal is either resolved or genuinely absent. Added `test_e2e_stale_merged_text_does_not_block_a_later_distinct_trigger` (a 3-turn simulation) proving a resolved trigger's stale signal text doesn't block detection of a later, distinct trigger. ADR-097's rationale, Consequences, and Limitations sections were corrected to match.

**[maintainability]** `claude/scripts/stop-tile-enumeration-gate.py:main()` — four parallel per-trigger structures (filed, not fixed in-PR)
Each trigger's identity is now spread across four sites that must stay hand-synchronized: the `_TRIGGERS` tuple, the evaluate block's hardcoded "nothing to resolve" skip-default tuples (`(None, False)` for PR/issue, `(False, False)` for table), the sentinel-marking loop's `fired` test (a silent `is not None` vs. bare-truthiness asymmetry), and the message-building block's per-trigger formatter call.
**Fix:** A fully generic loop was drafted and rejected during this review — the three evaluators don't share a return-shape or formatter-signature convention, so genericizing would need a 5th parallel element (per-trigger formatter dispatch via lambdas), trading four simple, explicit structures for one denser, more abstract one — not a clear net gain. Filed [dev-env#696](https://github.com/brownm09/dev-env/issues/696) to reconsider only if a 4th trigger is ever proposed.

---

### Findings Disposition

1. Test-integrity language-scope caveat — verified manually in this review; no violation found, no further action needed.
2. Pre-filter not gated per trigger — fixed in `08c93a4`.
3. Four parallel per-trigger structures — filed as [dev-env#696](https://github.com/brownm09/dev-env/issues/696); deferred by design (see finding above), not currently blocking.

---

PR #701, opened 2026-07-10
fix(hooks): route dev-env-sync.py advisories to stdout, not stderr

brownm09:
## Summary

Investigates and fixes dev-env#694: the sync hook reported "Pulled 0 commits" while the
canonical checkout was actually ~45 merged PRs behind `origin/main`.

**Root cause, confirmed live during this investigation, not just theorized:** `dev-env-sync.py`
is a `UserPromptSubmit` hook that always exits 0, but routed every warning (failed
fast-forward pull, diverged `main`, worktree squatting `main`) to **stderr**. Per the
[Claude Code hooks reference](https://code.claude.com/docs/en/hooks), only `UserPromptSubmit`'s
exit-0 **stdout** is added to Claude's context — stderr is not. At investigation time, the
canonical checkout (`~/Git/dev-env`) had an uncommitted, unrelated modification to
`claude/skills/sources.md` (leftover `/research` content from an earlier session, never
committed). Reflog pinned the timeline exactly: local `main` last successfully advanced
2026-07-09 00:06:01; 34 minutes later, PR #649 merged touching that same file; from that
moment on, **every** `git pull --ff-only` attempt failed identically — reproduced live,
36+ hours and 21+ commits later. The hook already detected this correctly and printed a
`WARNING` — to stderr, where it vanished. Direct proof from this very session: the hook fired
on this conversation's first turn (confirmed via git status matching the exact expected
post-fetch state), yet no `[dev-env-sync]` text of any kind appeared anywhere in the model's
context.

The exact literal "Pulled 0 commits" text could not be deterministically reproduced from the
code as a single path (that branch only executes after `local != remote` is already confirmed)
— most plausibly a TOCTOU race between two sessions' `dev-env-sync.py` invocations against the
same shared, unlocked canonical checkout. Full analysis in
[ADR-098](docs/adr/098-dev-env-sync-advisories-to-stdout.md).

A separate session earlier the same day independently traced this identical dirty-`sources.md`
condition from a different angle (an engineering-journal canonical-hijack investigation) and
filed [dev-env#697](https://github.com/brownm09/dev-env/issues/697) with the direct remediation
(commit the pending content) plus a spawned tile. That issue and this PR are complementary:
#697 fixes *this occurrence* of the drift; this PR fixes the hook's *general* inability to ever
surface such an occurrence, so the next one doesn't again require an hours-long investigation.

**Update:** #697 was closed in this same follow-on session — the "pending content" turned out
to be byte-identical to what PR #649 had already merged upstream (confirmed via matching git
blob hash), so it was discarded rather than re-committed, and the canonical fast-forwarded
cleanly. This PR (opened by an earlier, now-ended session) was then picked up, reviewed, and
its findings fixed in the same follow-on session — see Review findings disposition below.

## Changes

- All warning paths in `dev-env-sync.py` now print to **stdout** (never stderr).
- The non-cp1252-safe `⚠️` emoji is replaced with plain `WARNING:` text — these messages are
  now actually decoded (previously silently discarded on stderr regardless of encoding), so the
  encoding-safety requirement is live for the first time.
- Fast-forward-related messages (pulled / diverged / pull-failed) now state local/remote short
  SHAs and a commit-behind count — dev-env#694's own suggested follow-up.
- The success message flags a pre/post commit-count mismatch as a likely concurrent-process
  race, directly targeting the unreproduced "0 commits" ambiguity — refined during `/review` to
  not misattribute a failed measurement (see Review findings disposition).
- `main()` is wrapped in `try`/`except Exception: sys.exit(0)`, matching the fail-open
  convention PR #661 established for this file's siblings (`journal-canonical-guard.py`,
  `new-day-journal-check.py`) — refined during `/review` to print a fail-open notice rather than
  fail silently, since this file's whole point is eliminating invisible failures.
- Corrected `claude/CLAUDE.md`'s `## Observability` section, which stated a blanket
  "diagnostics go to stderr" rule — backwards for exactly this hook class (advisory,
  always-exit-0, `UserPromptSubmit`); this is the same misconception dev-env-sync.py's bug
  embodied.

## Scope note — journal-canonical-guard.py

That sibling hook (merged one day earlier in PR #661) has the **identical** stderr-on-exit-0
defect in all its warning paths. Not fixed here (different file, different motivating incident
— would have roughly doubled this PR's footprint). Filed as
[dev-env#699](https://github.com/brownm09/dev-env/issues/699) instead, citing ADR-098 directly.

## Testing

```
py -3 claude/scripts/tests/test_dev_env_sync.py
Tests: 14 passed, 0 skipped, 0 failed   (was 11 before /review fixes added 3 more cases)

py -3 claude/scripts/tests/test_worktree_topology.py   (regression check, unchanged file)

py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py claude/scripts/tests/*.py
SYNTAX OK
```

Pure message-formatting helpers (`_plural`, `_count_from`, `format_sync_note`,
`format_pull_failure_message`, `format_diverged_message`, `format_pulled_message`) are unit
tested offline, including 3 new cases added during `/review` (a zero-measurement case, an
ahead-only-not-diverged case, and a cp1252-encodability pin across all 5 formatters). The
git/subprocess orchestration is not separately tested — it has no local pure logic beyond the
formatters above, matching this repo's established convention for topology-diagnosing
orchestration scripts (`## Testing` items 22/26/30).

Suppression / test-integrity pre-PR greps: both clean (no output), re-run after the review
fixes too.

`/review` ([comment](https://github.com/brownm09/dev-env/pull/701#issuecomment-4940269321))
found 2 blocking + 3 non-blocking findings. All fixed in commit `3f27031`:

1. **[correctness, blocking]** `format_pulled_message` accepted `local`/`remote` but never used
   them — the success message never actually showed SHAs, contradicting this PR's own claim
   that all fast-forward messages do. **Fixed** — wired `local[:8]`/`remote[:8]` into the
   success message.
2. **[correctness, blocking]** Same function's mismatch-note could misattribute a failed
   `rev-list` measurement (fail-open to `0`) as "a concurrent process moved origin/main" — a
   real measurement is always ≥ 1 at that call site, so `0` can only mean the measurement
   failed. **Fixed** — `behind_count == 0` now reports "could not be measured" instead.
3. **[correctness, non-blocking]** `format_diverged_message` labeled a local-ahead-only state
   (`behind == 0`) as "diverged" alongside "0 commits behind" — internally contradictory.
   **Fixed** — distinct, non-contradictory wording for the ahead-only case.
4. **[maintainability, non-blocking]** No test asserted cp1252-encodability despite this PR's
   entire premise being a cp1252 fix. **Fixed** — added `.encode("cp1252")` pins across all 5
   formatters.
5. **[reliability, non-blocking]** The outer `except Exception: sys.exit(0)` swallows all
   output silently — confirmed this matches an existing, unmodified pattern in
   `journal-canonical-guard.py`/`new-day-journal-check.py` (not a regression). **Fixed** for
   this file specifically (prints one pure-ASCII fail-open notice first) since its whole
   purpose is eliminating invisible failures; the two sibling files are left as-is
   (out of scope, same precedent as the #699 deferral above).

## Doc reconciliation

- `docs/REFERENCE.md`: updated `dev-env-sync.py`'s entry to describe the stdout fix and new
  diagnostics.
- `README.md`: updated `dev-env-sync.py`'s hooks-table row similarly.
- `claude/CLAUDE.md`: new Testing item 56 (updated during `/review` to describe the 3 new test
  cases); corrected the Observability section's stream-choice guidance.
- `docs/adr/INDEX.md`: new ADR-098 row.
- `docs/adr/098-dev-env-sync-advisories-to-stdout.md`: updated during `/review` to describe the
  three refinements (SHA wiring, measurement-failure distinction, ahead-only wording, cp1252
  test pin, fail-open notice).

## ADR-warrant

Warranted — touches a documented hook's behavior (ADR-058, ADR-006) and corrects a
cross-cutting convention statement in `claude/CLAUDE.md` itself. New
[ADR-098](docs/adr/098-dev-env-sync-advisories-to-stdout.md), citing ADR-027/ADR-058/ADR-091 as
precedent. `docs/adr/INDEX.md` updated.

Closes #694


brownm09:
## Review: fix(hooks): route dev-env-sync.py advisories to stdout, not stderr

### Summary
The core fix — routing `dev-env-sync.py`'s advisories from stderr to stdout, since `UserPromptSubmit`'s exit-0 stdout is the only channel Claude Code actually surfaces — is correct and well-precedented (ADR-091's mirror-image case). Doc reconciliation, ADR-warrant, and test coverage for the new pure helpers are all present and accurate. However, two of the new formatter functions have real logic/functional bugs that undercut the PR's own stated goal of making failures "self-diagnosing": the success-message SHA parameters are dead code (never actually surfaced, contradicting the PR body's explicit claim), and the mismatch-detection note can misattribute a measurement failure as a "concurrent process" race. Recommend fixing both before merge — needs revision, not a straight merge.

---

### Blocking Findings

**[correctness]** `claude/scripts/dev-env-sync.py` — `format_pulled_message(local, remote, behind_count, pulled_lines)`
The function signature takes `local` and `remote` (both full SHAs) but never references either in its body — only `behind_count` and `pulled_lines` are used. Every sibling formatter (`format_sync_note`, `format_pull_failure_message`, `format_diverged_message`) actually uses `local[:8]`/`remote[:8]` to show short SHAs, and the PR body explicitly claims "Fast-forward-related messages (**pulled**/diverged/pull-failed) now state local/remote short SHAs and a commit-behind count" — but the success ("Pulled N commits") message never shows them. This is a functional gap that contradicts the PR's own stated behavior, and `docs/REFERENCE.md`'s updated entry repeats the same now-inaccurate claim.
**Fix:** Wire `local[:8]`/`remote[:8]` into the "Pulled N commits" message (e.g. append `format_sync_note(local, remote, behind_count)` or similar), matching the other two messages' pattern, and add a test asserting the SHAs appear in the success-path output.

**[correctness]** `claude/scripts/dev-env-sync.py` — `format_pulled_message`'s mismatch note (the `if count != behind_count:` block)
At the point this fires, the code has already established `base == local` and `local != remote`, so a *correctly measured* `behind_count` is always ≥ 1. `_count_from` fails open to `0` on any `rev-list` failure (documented as "advisory-only diagnostic"). So `behind_count == 0` reaching this branch can only mean the measurement itself failed — yet if the pull then succeeds and pulls ≥ 1 commit, the mismatch note fires and blames it on "a concurrent process," which is simply wrong in this case. Since this file runs on every prompt against a shared, multi-session canonical checkout where transient git failures (or a concurrent `git gc`) are realistically in scope, this isn't a hypothetical: it's a plausible real path to a misleading diagnostic in the exact function whose entire purpose is accurate self-diagnosis.
**Fix:** Guard the note on a successful measurement, e.g. `if behind_count > 0 and count != behind_count:`, or give the `behind_count == 0`-but-commits-were-pulled case its own distinct wording ("commit count could not be measured before pulling") rather than conflating it with the concurrent-process case.

---

### Non-Blocking Findings

**[correctness]** `claude/scripts/dev-env-sync.py:format_diverged_message`
The `base != local` branch fires both for genuine divergence and for a local-ahead-only state (`base == remote`, i.e. someone committed directly onto the canonical with nothing new on origin). In the ahead-only case, the new counts render as `"...is N commits ahead and 0 commits behind origin/main ... has diverged"` — internally contradictory wording (0 behind isn't divergence), newly visible now that this PR adds explicit counts to what was previously a vaguer message.
**Suggestion:** Either distinguish ahead-only from true divergence in the message, or accept the imprecise wording as a known simplification (the underlying advice to investigate is still correct either way).

**[maintainability]** `claude/scripts/tests/test_dev_env_sync.py`
None of the 11 new tests assert that the formatters' output is encodable for Claude Code's cp1252 hook-output pipe on Windows, even though this repo has an established convention of pinning exactly that property on advisory-emitting scripts (Testing items 18/40/46/50 in `CLAUDE.md`), and this PR's entire premise is fixing a cp1252 failure (the `⚠️` emoji). I verified by hand that the current output is cp1252-safe (the only non-ASCII character is the em dash, U+2014 / cp1252 0x97), but nothing in the suite would catch a regression — e.g. someone adding a "✓" or smart-quote character later sails through green and silently reintroduces this exact bug class.
**Suggestion:** Add a `.encode("cp1252")` assertion (not `.isascii()` — the em dash is non-ASCII but cp1252-safe) for each of the five new formatters' output, matching the sibling tests' convention.

**[reliability]** `claude/scripts/dev-env-sync.py` — the `if __name__ == "__main__": try: main() except Exception: sys.exit(0)` block
This swallows all output on any unexpected failure — verified this matches the exact pre-existing pattern in `journal-canonical-guard.py:172` and `new-day-journal-check.py:273` (both identical `except Exception: sys.exit(0)` with no message), so it's not a regression this PR introduces. Still worth flagging here specifically: this file's whole purpose is eliminating invisible failures, and a `UnicodeEncodeError` from a future non-cp1252 character (see the finding above) would hit exactly this handler and vanish completely — silently reconstructing dev-env#694's original symptom.
**Suggestion:** Consider a minimal ASCII fail-open notice to stdout before `sys.exit(0)` (e.g. `print("[dev-env-sync] WARNING: sync check failed unexpectedly — skipped this prompt.")`). Given the identical pattern already exists unmodified in two other merged files, I'm not blocking this PR on it — filing a separate follow-up covering all three files is more appropriately scoped than expanding this PR's footprint.

---

### Style Notes
No style issues beyond what's already covered above — naming, docstrings, and formatting are consistent with the rest of the file.

---

### Findings Disposition
_To be added after fixes are applied in this same session._

---

*Reviewed by claude-opus-4-8 (subagent analysis) via Claude Code*

PR #705, opened 2026-07-10
fix(hooks): route journal-canonical-guard.py advisories to stdout, not stderr

brownm09:
## Summary

Fixes dev-env#699: `journal-canonical-guard.py` is a `UserPromptSubmit` hook that always exits
0, but every one of its four warning paths (worktree-list-unreadable, warn-squatter,
warn-dirty, auto-return-checkout-failed) routed to **stderr**. Per the
[Claude Code hooks reference](https://code.claude.com/docs/en/hooks), only a `UserPromptSubmit`
hook's exit-0 **stdout** is added to Claude's context -- stderr is not. This is the identical
defect [ADR-098](docs/adr/098-dev-env-sync-advisories-to-stdout.md) fixed in the sibling
`dev-env-sync.py` ([PR #701](https://github.com/brownm09/dev-env/pull/701), open, not yet
merged at the time of this PR) -- that PR's author found this hook's parallel defect by
inspection and deliberately scoped it out, filing dev-env#699 instead of doubling that PR's
footprint. This PR is that follow-up, mirroring the fix exactly.

## Changes

- All four warning paths in `journal-canonical-guard.py` now print to **stdout** (never
  stderr).
- No emoji-to-`WARNING:` substitution needed here (unlike `dev-env-sync.py`) -- every warning
  already read plain ASCII `WARNING:` text; confirmed by scanning the full file for non-ASCII
  characters (only em dashes in docstrings/comments, never inside a `print(...)` argument, and
  cp1252-representable regardless).
- No SHA/commit-behind diagnostics added (unlike ADR-098's `dev-env-sync.py` enrichment) --
  dev-env#699 explicitly scoped this out: this hook's warnings are about worktree/branch
  topology, not commit counts, so there's no analogous ambiguity to resolve.
- Module docstring gains a short paragraph documenting the stdout contract, mirroring
  `dev-env-sync.py`'s equivalent addition.
- New `claude/scripts/tests/test_journal_canonical_guard.py`: a narrow end-to-end test suite
  (6 tests) driving the real hook via subprocess against a disposable throwaway git repo, using
  the pre-existing `JOURNAL_CANONICAL_GUARD_REPO_PATH` test seam (previously unused by any
  test). Written test-first -- confirmed to fail on exactly the two warning paths this PR
  touches (warn-dirty, warn-squatter) against the pre-fix code. This is a deliberate, narrow
  departure from [ADR-093](docs/adr/093-journal-canonical-hijack-guard.md)'s "no dedicated test
  file" precedent for this exact file -- that decision was reasoned about topology-*decision*
  correctness (unaffected by this change, still covered by `test_worktree_topology.py`), not
  the orthogonal stream-*routing* axis this new suite covers.
- New [ADR-099](docs/adr/099-journal-canonical-guard-advisories-to-stdout.md), mirroring
  ADR-098's structure, citing ADR-091/ADR-093/ADR-098 and explaining the departure from
  ADR-093's testing precedent.

## Testing

```
py -3 claude/scripts/tests/test_journal_canonical_guard.py
Tests: 6 passed, 0 skipped, 0 failed

py -3 claude/scripts/tests/test_worktree_topology.py   (regression check, unchanged file)

py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py claude/scripts/tests/*.py
SYNTAX OK
```

The new test file has no pure-function layer -- this fix adds no new logic, only a
stream-routing change, so there's nothing to unit test offline beyond the end-to-end proof
above. Two of the four warning paths (worktree-list-unreadable, auto-return-checkout-failed)
are deliberately not exercised end-to-end -- see the test file's docstring for why each is
fragile to construct reliably, matching this repo's established convention for
hard-to-construct git-failure-injection paths (`## Testing` items 22/26/30).

Suppression / test-integrity pre-PR greps: both clean (no output). No test files deleted, no
coverage thresholds touched (N/A, this is a Python repo).

- **[reliability]** `test_warn_dirty_lands_on_stdout_not_stderr` relied on an untracked file's
  visibility in `git status --porcelain`, which depends on the ambient
  `status.showUntrackedFiles` git config -- fixed in `429a480` by staging the file instead, so
  the dirty-detection is deterministic regardless of that setting. Re-ran the suite green
  (6/6) after the fix.

Full review: https://github.com/brownm09/dev-env/pull/705#issuecomment-4940257887

## Doc reconciliation

- `docs/REFERENCE.md`: updated `journal-canonical-guard.py`'s entry to describe the stdout fix.
- `README.md`: updated `journal-canonical-guard.py`'s hooks-table row similarly.
- `CLAUDE.md`: new Testing item 57 (deliberately skipping 56, claimed by the still-open
  PR #701, to avoid a live known numbering collision; a gap there is non-blocking per
  `pre-merge-numbering-check.py` and will close naturally once #701 merges).
- `docs/adr/INDEX.md`: new ADR-099 row (099 chosen for the same reason -- 098 is claimed by
  PR #701; the resulting numbering gap is a non-blocking advisory, not a collision).
- Deliberately **not** touching `CLAUDE.md`'s `## Observability` section's stream-choice
  wording -- PR #701 already corrects that general blanket-rule text as part of its own
  scope; duplicating it here would risk an unnecessary merge conflict between the two PRs on
  the same lines.

## ADR-warrant

Warranted -- touches a documented hook's behavior (ADR-093) and extends the same
observability-contract failure class ADR-091/ADR-098 already document. New
[ADR-099](docs/adr/099-journal-canonical-guard-advisories-to-stdout.md), citing
ADR-091/ADR-093/ADR-098 as precedent. `docs/adr/INDEX.md` updated.

Closes #699


brownm09:
## Review: fix(hooks): route journal-canonical-guard.py advisories to stdout, not stderr

### Summary
This PR is ready to merge with one minor fix. It correctly and completely removes `file=sys.stderr` from all four warning `print()` calls in `journal-canonical-guard.py` — a grep confirms zero `sys.stderr` references remain, and the success-path message was already on stdout, so no site was missed. The new end-to-end test suite (`test_journal_canonical_guard.py`) genuinely proves the fix: its assertions were checked against the real decision logic in `_worktree_topology.py` and correctly reach `warn-squatter`/`warn-dirty`/`return-canonical` per the actual `canonical_sync_action` branch order, and both regression tests would fail against the pre-fix code (confirmed in the PR's own TDD description). One non-blocking reliability finding in the new test fixture's hermeticity.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/tests/test_journal_canonical_guard.py:146` (`test_warn_dirty_lands_on_stdout_not_stderr`)
The test establishes a "dirty" canonical by writing an **untracked** file (`(canonical / "dirty.txt").write_text(...)`) and never staging it, relying on the hook's own `git status --porcelain` call to report it as dirty.
Whether an untracked file appears in `--porcelain` output depends on the ambient git config `status.showUntrackedFiles` — a real, if uncommon, setting (e.g. `no`, sometimes set on large-repo machines). If a developer's machine has that set globally, `git status --porcelain` returns empty for this fixture, `clean` evaluates `True` in the hook, `canonical_sync_action` takes the `return-canonical` branch instead of `warn-dirty`, and the test fails with a confusing assertion mismatch instead of testing what it says it tests. The fixture already neutralizes `init.templateDir`/`core.hooksPath` specifically to keep ambient machine config from leaking in — this is the one remaining gap in that same hermeticity discipline.
**Suggestion:** Stage the file before invoking the hook (`git -C <canonical> add dirty.txt`) — staged changes always appear in `--porcelain` output regardless of `status.showUntrackedFiles`, making the dirty-detection deterministic across machines.

---

### Findings Disposition

_To be added by the author after addressing the finding above (fixed in `<sha>` or filed as a tracked issue), per ADR-028/ADR-039._

---

*Reviewed by claude-opus-4-8 (2 parallel subagents: correctness/security, reliability/performance/maintainability) via Claude Code*

PR #706, opened 2026-07-10
feat(hooks): add stop-journal-stub-checkpoint Stop hook for report/analysis sessions

brownm09:
Closes #702

## Summary

Mechanizes the previously prose-only **"Report / analysis generated"** journal trigger ([ADR-062](docs/adr/062-journal-report-analysis-trigger.md)). A new global **Stop** hook, `stop-journal-stub-checkpoint.py`, blocks the stop (exit 2, reminder on stderr) when a report/analysis/verification session did substantive work but is ending with no engineering-journal stub — and opened no PR (those already nudge), isn't a `/review` session, and carries no "skip journal" override.

- **Delivery:** exit-2 + stderr, blocking-once. Per [ADR-091](docs/adr/091-journal-stop-check-archive-reminder-blocking.md) that's the only Stop delivery that reaches Claude; a once-per-session sentinel + `stop_hook_active` loop guard + a one-line dismissal keep it advisory in spirit. Fires in background/SDK sessions ([ADR-053](docs/adr/053-posttooluse-hooks-inert-in-background-sessions.md)/[055](docs/adr/055-reliable-event-inert-posttooluse-advisory.md)), unlike a PostToolUse hook.
- **Detection:** user-prompt keyword intent (scoped to genuine user text) **AND** >=5 substantive tool calls **AND** no PR **AND** no stub **AND** not `/review` **AND** no skip override.

Design: [ADR-100](docs/adr/100-stop-journal-stub-checkpoint-hook.md). Motivating incident: a lifting-logbook session verifying PR #770's production fix that wrote no stub until the user asked (2026-07-10).

## Numbering

Authored as ADR-098 / item 56, renumbered to **ADR-100 / item 58** (commit `73072ae`) after confirming open PR #701 claims ADR-098 + item 56 and PR #705 claims ADR-099 + item 57 (both unmerged, invisible on `main`). Gaps are non-blocking advisories in `pre-merge-numbering-check.py`.

## Testing

- **Suite** — `py -3 claude/scripts/tests/test_stop_journal_stub_checkpoint.py` → **Tests: 64 passed, 0 skipped, 0 failed** (55 original + 9 added in review). Pure helpers (`report_intent` incl. every false-positive guard + the narrowed-keyword cases; `substantive_tool_count`; `opened_or_merged_pr`; `wrote_stub` incl. read-vs-write Bash; `is_review_only_session` incl. `/review-*` boundary; `skip_override`; `evaluate`; `format_reminder` cp1252 incl. non-ASCII cwd; malformed-record robustness) + a subprocess e2e layer (fire→exit-2/empty-stdout; every suppression/loop-guard/below-threshold/no-intent path→exit-0; sentinel-suppresses-refire; empty-stdin / non-dict-payload fail-open).
- **Hook-script syntax (item 1):** all `claude/scripts/*.py` parse. **Crude-substring AST gate (item 39):** 13 passed, 0 offenses. **settings.json:** valid JSON.
- **Coverage gate:** all new behavior covered. **Suppression grep:** clean. **Test-integrity grep** (JS + Python idioms): clean. Python test file manually verified free of skip idioms.

Reviewed by `claude-opus-4-8` (comment above). 1 blocking + 9 non-blocking; all addressed:

- **[reliability] non-cp1252 `cwd` → UnicodeEncodeError after sentinel burned (silent self-disable)** — fixed in `064a7a3` (`format_reminder` sanitizes cwd to ASCII; +test).
- **[correctness] over-broad keywords false-fire on dev prompts** — fixed in `064a7a3` (`analy...` excludes `analytics`; bare `deploy` dropped; +tests). Residual incidental-keyword class documented in ADR-100 Limitations.
- **[correctness] pre-filter not a strict superset (JSON-escaped whitespace)** — fixed in `064a7a3` (atomic-stem `_PREFILTER_RE`).
- **[correctness] `wrote_stub` Bash branch trips on `ls`/`cat` reads** — fixed in `064a7a3` (write-scoped `_STUB_BASH_WRITE_RE`; +tests).
- **[correctness] `_REVIEW_CMD_RE` over-matches `/review-*`** — fixed in `064a7a3` (`(?!-)` boundary; +test).
- **[correctness] sentinel-once / fires-at-first-qualifying-turn (not session end)** — documented in `064a7a3` (ADR-100 Limitations + hook docstring); inherent to a non-terminal Stop trigger, accepted.
- **[performance] pre-filter low selectivity / re-parse cost** — documented in `064a7a3` (ADR-100 Consequences corrected; cost bounded, equals the tile gate).
- **[maintainability] `_is_synthetic_user` docstring inaccurate** — fixed in `064a7a3` (reworded).
- **[maintainability] e2e fail-open paths untested** — fixed in `064a7a3` (empty-stdin / non-dict-payload e2e added).
- **[maintainability] share `_user_message_texts`/`_is_synthetic_user` with the tile gate via `_hookutil`** — filed [#710](https://github.com/brownm09/dev-env/issues/710) (out of scope: touches the sibling hook + its ~76 tests).


brownm09:
## Review: feat(hooks): add stop-journal-stub-checkpoint Stop hook for report/analysis sessions

### Summary
Merge with minor changes. The hook is correctly wired, fail-open, honors `stop_hook_active`, writes the sentinel before emit, and ships 55 passing tests plus a subprocess e2e layer. Security is clean (read-only transcript scan). The highest-severity finding is a **reliability** bug: a non-cp1252 `cwd` interpolated into the reminder raises `UnicodeEncodeError` at the stderr write — caught by the outer guard, but only *after* the sentinel is written, so the hook silently self-disables for the whole session. All findings below are being addressed in-PR.

---

### Blocking Findings

**[reliability]** `claude/scripts/stop-journal-stub-checkpoint.py` — `format_reminder` interpolates the free-form payload `cwd` into the exit-2 reminder. Claude Code pipes Stop-hook stderr as cp1252 on Windows; a project path under a Unicode-named folder makes `sys.stderr.write` raise `UnicodeEncodeError`. It's caught by `main()`'s outer `except Exception: sys.exit(0)` (fail-open holds, no traceback), but `_mark_fired` already ran — the sentinel is burned and the reminder is never delivered, silently disabling the hook for that session. The `format_reminder` ASCII discipline (and its cp1252 test) only cover the static text; this is the family's first Stop reminder to interpolate an unbounded string.
**Fix:** Sanitize `cwd` to ASCII before interpolation; add a `format_reminder(<non-ASCII cwd>)` cp1252 test.

---

### Non-Blocking Findings

**[correctness]** Intent keyword lists match ordinary implementation prompts, not just report requests. Verified FIRE-ing prompts include "add analytics tracking" (`analy\w*` matches *analytics*), "the deploy broke, fix the Dockerfile" (bare `deploy\w*`), "add a reports tab", "comparison helper", "verify this compiles then commit". Because a Stop hook fires at every turn-end, the fire lands at turn-1-end (before any PR exists), so the "no PR" gate doesn't screen in-progress dev sessions.
**Suggestion:** Narrow the clearest over-matches (`analy\w*`→exclude analytics; drop bare `deploy\w*`, keep verify/production/check-the-deploy contexts); document the residual incidental-keyword class in ADR-100. (Fixing in-PR.)

**[correctness]** The once-per-session sentinel is consumed at that first (often premature) qualifying Stop, so a multi-turn report session can have its single nudge spent early and the true-session-end Stop (report done, still no stub) suppressed.
**Suggestion:** Document that the nudge fires at the first qualifying turn, not session end. (Documenting in ADR-100 + docstring.)

**[correctness]** The `main()` pre-filter runs the intent regexes (with `\s+` compound branches) over the raw JSON transcript; a compound keyword whose inter-word whitespace is JSON-escaped (`\t`/`\n`) fails the raw scan but matches the parsed detector, so `main()` can fast-exit a fire-worthy session — contradicting the "can never fast-exit" claim. Safe direction (missed advisory), but the claim is wrong.
**Suggestion:** Make the pre-filter an atomic-stem superset (no `\s+` dependency). (Fixing in-PR.)

**[correctness]** `_REVIEW_CMD_RE` (`<command-name>\s*/?review\b`) over-matches a future `/review-followups`-style command, silently exempting the session. Latent (only `/review` exists today).
**Suggestion:** Anchor the terminator (`review\b(?!-)`). (Fixing in-PR.)

**[correctness]** `wrote_stub`'s Bash branch matches any `.stub.md` mention, so the journal workflow's own `ls sessions/…/*.stub.md | sort | tail -1` (a read) or `cat …stub.md` sets `wrote_stub=True` and suppresses the nudge.
**Suggestion:** Restrict the Bash match to write-ish operations (`git add`/`git commit`/redirect/`mv`/`cp`). (Fixing in-PR.)

**[performance]** The pre-filter's keywords (`report`, `analy*`, `summar*`, `verif*`, `deploy*`, …) are common in assistant prose and tool output, so on substantive sessions the full `read_text` + `_parse_records` runs on every Stop for the rest of the session (no sentinel is written in the `(False, False)` no-intent path). ADR-100's "common case is a single regex scan, no parse" overstates selectivity.
**Suggestion:** Correct the ADR framing; the cost is bounded (equals the tile gate's worst case) and accepted. (Documenting in ADR-100.)

**[maintainability]** `_is_synthetic_user`'s docstring describes checking for a `<local-command-*>` marker, but the body only checks `isMeta`/`isCompactSummary` flags.
**Suggestion:** Reword to describe the flag check. (Fixing in-PR.)

**[maintainability]** `_user_message_texts`/`_is_synthetic_user` re-implement user-text extraction + synthetic-record filtering that `stop-tile-enumeration-gate.py` does inline in `skip_override` — the same drift risk `_hookutil` (ADR-090) was created to prevent.
**Suggestion:** Promote both to `_hookutil` and have the tile gate consume them. Out of scope here (touches the tile gate + its 76 tests) — filing as a follow-up.

**[maintainability]** The subprocess e2e layer never exercises the top-level fail-open paths (empty stdin, non-dict payload, missing transcript). Safe by the outer guard, but cheap insurance for an every-Stop hook.
**Suggestion:** Add e2e cases for empty stdin / non-dict payload → exit 0, empty output. (Fixing in-PR.)

**[maintainability]** Language-scope: the diff's test file is Python; the JS/TS integrity scanners don't cover it. Manually verified: no `@pytest.mark.skip`/`pytest.skip(`/skip idioms (plain function suite, no pytest), and the PR body carries the `Tests: N passed…` summary.

---

### Findings Disposition
Recorded in the PR body's "Review findings disposition" section before merge (ADR-039). All blocking + non-blocking findings are fixed in-PR except the `_hookutil` promotion (filed as a follow-up).

---

PR #712, opened 2026-07-10
fix(hooks): add bounded tail reader to _hookutil, use it in idle-refresher

brownm09:
## Summary

`_hookutil.load_records` always reads and parses a transcript's entire JSONL file. `idle-refresher.py` fires on every `UserPromptSubmit` but only ever needs the last assistant record's timestamp, so it was paying a full parse of a potentially multi-MB transcript on every prompt — flagged independently by two adversarial `/review` subagents reviewing PR #673.

- Add `iter_records_reverse(transcript_path, chunk_size=...)` to `claude/scripts/_hookutil.py` — a generator that reads the file from the end in bounded 64 KiB chunks, yielding JSON-object records most-recent-first. A caller stops consuming (`break`/`next()`) as soon as it finds a match; the unread remainder is never touched. Chunk boundaries are found on raw bytes before UTF-8 decoding, so multi-byte characters split across chunk reads decode correctly. A not-yet-terminated line's bytes accumulate in a list across chunk reads and are joined once, not re-concatenated every iteration (see Review findings below — this was a real bug caught pre-merge).
- `load_records` is unchanged — this is purely additive, kept for callers that need the whole transcript (`stop-tile-enumeration-gate.py`'s session-wide scan), per the issue's explicit guidance.
- `idle-refresher.py`'s `last_activity_epoch` now takes an already most-recent-first iterable (dropped the internal `reversed()`) instead of a materialized forward list — that's what actually lets it consume the new generator lazily; keeping `reversed()` would force full materialization regardless of what was passed in.
- Verified via grep that the other three callers named in the issue (`posttooluse-inert-advisory.py`, `stop-tile-enumeration-gate.py`, `reconcile-open-prs.py`) don't call `load_records`/`_parse_records` on a path this change touches.

Closes #679.

## Testing

Ran the full set of affected suites from `CLAUDE.md`'s `## Testing` list, both before and after the review-driven fix:

```
py -3 claude/scripts/tests/test_hookutil.py           # item 27 (extended: 21 -> 33 tests)
py -3 claude/scripts/tests/test_idle_refresher.py      # item 51 (extended: 14 -> 15 tests)
py -3 claude/scripts/tests/test_posttooluse_inert_advisory.py   # item 18 (unchanged, regression check)
py -3 claude/scripts/tests/test_stop_tile_enumeration_gate.py   # item 48 (unchanged, regression check)
py -3 claude/scripts/tests/test_reconcile_open_prs.py            # item 19 (unchanged, regression check)
py -3 -c "import ast,sys; [ast.parse(open(f,encoding='utf-8').read(),f) for f in sys.argv[1:]]" claude/scripts/*.py  # item 1
```

Tests: 33+15+32+116+15 = 211 passed, 0 skipped, 0 failed (all suites green; hook-script syntax check clean).

New `test_hookutil.py` coverage: `_record_from_line` (valid/blank/malformed/non-object), `iter_records_reverse` basic reverse order, a property check matching `reversed(load_records(...))` across 9 chunk sizes (1..4096) forcing lines to reassemble across chunk boundaries, multi-byte UTF-8 characters surviving small chunk sizes, no-trailing-newline, blank/malformed-lines-skipped, empty file, `FileNotFoundError` on a missing path, `ValueError` on a non-positive `chunk_size`, a `builtins.open` read-call-counting laziness test with a `>= 1` floor assertion (mirroring `test_prune_merged_worktrees.py`'s established mocking precedent) proving a ~4000-line fixture yields its tail record after a single chunk read, and a timing-based regression test for the quadratic-growth fix (~2MB single line, ~15600 chunk reads, 5s bound; completes in ~0.08s post-fix).

New `test_idle_refresher.py` coverage: the three `last_activity_epoch` fixtures reordered newest-first for the new contract, plus `test_last_activity_epoch_consumes_lazily` — a hand-rolled generator that raises `AssertionError` if pulled past its first match.

No suppressions, no test-integrity violations, no deleted test files (pre-PR greps both clean, re-verified after the review fix).

## Coverage gate

New behavior (`iter_records_reverse`, `_record_from_line`, the `last_activity_epoch` contract change) is fully covered per above — no deferral.

## ADR

Added [Amendment 1](https://github.com/brownm09/dev-env/blob/main/docs/adr/090-shared-transcript-readers-hookutil.md) to ADR-090 (the ADR that established `_hookutil`'s transcript-record readers) documenting the new bounded tail reader, including the quadratic-growth finding and fix, and updated its `docs/adr/INDEX.md` row (amended date + new tags).

## Documentation

Extended `CLAUDE.md` Testing items 27 (`_hookutil` shared-helper test) and 51 (idle-refresher test) in place to describe the new coverage — no new numbered item, no renumbering. Updated `README.md` / `docs/REFERENCE.md`'s `_hookutil.py` prose to mention `iter_records_reverse` (caught as a blocking finding in `/review` Step 2b/2c, fixed in this PR).

`/review` ran with two parallel opus subagents (correctness/security; reliability/perf/maintainability). Correctness/security came back clean. 1 blocking + 5 non-blocking findings from the other pass + Step 2b/2c doc-reconciliation check — **all fixed in this PR**, commit [`e986100`](https://github.com/brownm09/dev-env/pull/712/commits/e986100):

1. **[blocking][documentation]** `README.md`/`docs/REFERENCE.md` `_hookutil.py` prose missing `iter_records_reverse` — fixed, both files updated.
2. **[non-blocking][performance]** `iter_records_reverse` had O(line_length²/chunk_size) quadratic buffer re-concatenation, hitting idle-refresher.py's own hot path (a large pasted prompt as the record adjacent to the scan target) — fixed, rewritten to accumulate-and-join-once (O(line_length)); added a dedicated timing-based regression test.
3. **[non-blocking][maintainability]** Laziness test could vacuously pass if a future refactor stopped routing through `open()` — fixed, added `>= 1` floor assertion.
4. **[non-blocking][maintainability]** `_record_from_line` docstring inaccurately claimed UTF-8 error handling "matches" `_parse_records`'s contract — fixed, reworded to state it's more lenient than `load_records`, not identical.
5. **[non-blocking][maintainability]** `last_activity_epoch` docstring didn't document the silent-wrong-answer failure mode for a caller passing forward-chronological order — fixed, added explicit warning sentence.
6. **[non-blocking][maintainability]** `chunk_size` validation raises on first iteration, not at call time — fixed, documented explicitly in the docstring (a full restructure was judged disproportionate to the low severity).

Full suite (211 tests) re-ran green after all fixes.


brownm09:
## Review: fix(hooks): add bounded tail reader to _hookutil, use it in idle-refresher

### Summary
Merge-ready. The highest-severity finding was a real **O(line_length^2 / chunk_size) quadratic buffer-re-concatenation bug** in `iter_records_reverse` (caught by the reliability/perf subagent) — the first implementation's `buf = f.read(read_size) + tail` re-copied the growing not-yet-terminated line fragment on every chunk read, and the subagent identified this hits the PR's own hot path directly: the record right before whatever the generator scans past is often the transcript's newest entry (a user's just-submitted prompt on `idle-refresher.py`'s `UserPromptSubmit` path), whose size a large paste puts under the user's control. Empirically, a 2MB paste added ~30ms of overhead vs. the `load_records` call this PR replaces, a 10MB paste ~540ms — the PR's own stated motivation (multi-MB transcripts) could have made this a regression, not an improvement. **This has already been fixed** in a follow-up commit (`e986100`), along with every other finding below. Diff size (475 lines) is appropriately scoped for a single well-defined issue (dev-env#679) — no split needed.

---

### Blocking Findings

**[documentation]** `README.md:55`, `docs/REFERENCE.md:169`
Both files carry a prose paragraph enumerating `_hookutil.py`'s exposed function surface (`cleanup_stale_sentinels`, `sentinel_path`, `find_transcript`, `load_records`, `_parse_records`, `iter_bash_calls`, `_result_text`, `_content_items`), but the new `iter_records_reverse` was not added to either enumeration in the original diff.
A future reader orienting from these repo-level docs (the established discoverability path for this shared module, since `_hookutil.py` has no directory-level README of its own — confirmed no `claude/scripts/README.md` exists) would not know the bounded tail reader exists.
**Fix:** Add `iter_records_reverse` to both prose paragraphs, describing its purpose and consumer.

---

### Non-Blocking Findings

**[performance]** `claude/scripts/_hookutil.py` (original `iter_records_reverse`)
`buf = f.read(read_size) + tail` re-concatenated the entire not-yet-terminated fragment from scratch every chunk iteration when no newline was found, costing `O(line_length^2 / chunk_size)` for a single line spanning many chunks.
This directly undermines the PR's stated purpose in a realistic scenario (see Summary) — not a rare pathological input.
**Suggestion:** (Already applied) Accumulate fragments in a list and join once when a terminator is found, making the cost `O(line_length)` regardless of how many chunks the line spans.

**[maintainability]** `claude/scripts/tests/test_hookutil.py` — `test_iter_records_reverse_is_lazy_stops_early`
The test asserted only `len(read_calls) <= 2`; if a future refactor moved `iter_records_reverse` off `open()` (e.g. to `Path.read_bytes()`), the `builtins.open` patch would silently stop intercepting, `read_calls` would stay empty, and the test would pass *vacuously* (the correctness assertion on the returned record would still hold from the real, unpatched read).
A laziness guarantee that can silently stop being tested is worse than not having the test, since it reads as continued coverage.
**Suggestion:** (Already applied) Add `assert len(read_calls) >= 1` so the mock's own interception is verified, not just assumed.

**[maintainability]** `claude/scripts/_hookutil.py` — `_record_from_line` docstring
The docstring claimed `errors="replace"` decoding "match[es] `_parse_records`'s skip-and-continue contract," but `_parse_records` never decodes (it receives already-`str` input) — the actual comparison point is `load_records`'s strict `open(..., encoding="utf-8")`, which *raises* `UnicodeDecodeError` on malformed bytes rather than skipping. The two readers genuinely diverge on a corrupt-UTF-8 transcript.
A maintainer relying on the docstring's "parity" claim to swap one reader for the other would be surprised by the actual divergence (though the new behavior is arguably more robust, not worse).
**Suggestion:** (Already applied) Reword to state it degrades corrupt bytes to a skip, which is *more lenient* than `load_records`, not identical to it.

**[maintainability]** `claude/scripts/idle-refresher.py` — `last_activity_epoch` docstring
The contract change (forward list + internal `reversed()` -> already-most-recent-first iterable) has exactly one live caller today, but a future caller passing forward-chronological order by mistake gets a **silently wrong** answer (oldest matching timestamp instead of newest) rather than an exception, since laziness means the function can't verify the order it's handed.
Undocumented silent-failure modes are the kind of thing that cost real debugging time when someone eventually hits them.
**Suggestion:** (Already applied) Add one explicit sentence to the docstring naming this failure mode.

**[maintainability]** `claude/scripts/_hookutil.py` — `iter_records_reverse`'s `chunk_size` validation
The `if chunk_size <= 0: raise ValueError(...)` guard is inside the generator body, so it only raises on first `next()`/iteration, not when `iter_records_reverse(path, chunk_size=0)` is called — a caller that constructs the generator in one place and consumes it elsewhere would see the error surface at a confusing, distant point.
Low severity — the current caller consumes immediately inside a `try`, so this is latent, not live.
**Suggestion:** (Already applied) Documented the first-iteration-raise timing explicitly in the docstring rather than restructuring into a validating wrapper, which would add complexity disproportionate to the severity.

---

### Findings Disposition

All 6 findings (1 blocking, 5 non-blocking) were fixed in commit [`e986100`](https://github.com/brownm09/dev-env/pull/712/commits/e986100) — see that commit's message for the itemized list. No findings were filed as separate issues; all were small, well-scoped, and addressed directly in this PR. Full test suite (`test_hookutil.py` 33, `test_idle_refresher.py` 15, `test_posttooluse_inert_advisory.py` 32, `test_stop_tile_enumeration_gate.py` 116, `test_reconcile_open_prs.py` 15 — 211 total) re-ran green after the fixes.

---

PR #714, opened 2026-07-11
fix(skills): route source-library writes through a worktree, not canonical

brownm09:
## Summary

`/research` and `journal-compose` Section 11 both appended new source-library citations
directly to `~/.claude/skills/sources.md` — a directory junction that always resolves to the
canonical dev-env checkout's working tree, regardless of which project the invoking session was
actually working in. Neither committed the write, so the canonical checkout went dirty and
blocked `dev-env-sync.py`'s `git pull --ff-only` for every session on the machine — twice
([PR #649](https://github.com/brownm09/dev-env/pull/649),
[issue #697](https://github.com/brownm09/dev-env/issues/697), the second time for ~41 hours /
21 commits before being noticed).

**Scope expansion (confirmed with the user before implementing, see the
[issue #708 comment](https://github.com/brownm09/dev-env/issues/708#issuecomment-4940791322)):**
a repo-wide grep for `sources.md` found `journal-compose/SKILL.md` Section 11 has its own inlined
copy of the identical write, invoked automatically every morning by the fully-unattended
`daily-journal-compose` scheduled routine — a stronger match for both incidents' "background
research subagent" framing than the interactive `/research` command. `journal-compose`'s own
worktree isolation ([ADR-082](docs/adr/082-journal-compose-worktree-isolation.md)) never covered
this write since it targets dev-env, a different repo than engineering-journal.

## Changes

- **New:** `claude/skills/queue-source-library-entry/SKILL.md` — a shared helper skill (mirroring
  the existing `sync-routine-worktree` pattern) that appends an entry to `claude/skills/sources.md`
  inside a dedicated, reused worktree (`.claude/worktrees/research-sources-queue`, branch
  `chore/research-sources-queue`) instead of through the canonical-resolving junction path.
  Commits inside the worktree and best-effort pushes for durability. **Does not auto-open a PR**
  — the queue branch accumulates entries; a human sweeps it into a PR when convenient, the same
  manual pattern already used twice. Hardened during `/review` — see Review findings disposition
  below.
- `claude/skills/research/SKILL.md` — Library feedback loop step now invokes the new helper
  instead of writing to `~/.claude/skills/sources.md` directly. Adds `Bash` to `allowed-tools`.
- `claude/skills/journal-compose/SKILL.md` — Section 11 Pass 2 now invokes the new helper the
  same way. Read-only grep/cache-lookup paths in both skills are unchanged.
- `docs/adr/101-source-library-writes-through-worktree.md` (new) + `docs/adr/INDEX.md` — design
  rationale, including the two rejected alternatives (auto-PR per entry; fresh worktree per
  invocation instead of one reused queue), plus a Review Hardening section documenting all
  `/review` findings and fixes.
- `docs/REFERENCE.md` — updated `/research` and `/journal-compose` entries to describe the new
  write path. No new top-level Skills-table row for the helper skill, matching how
  `sync-routine-worktree` (an existing internal-only helper) is already documented — inline
  mentions only, not a `### /queue-source-library-entry` heading.

## Test plan

Skill-markdown and documentation only — no `.py`/`.sh` files changed, matching
[ADR-082](docs/adr/082-journal-compose-worktree-isolation.md)'s own precedent for this class of
change. `Tests: N/A — no automated test suite covers skill-markdown content (dev-env project has
no test coverage for skill-markdown files; see dev-env's `## Testing` section scope)`.

Verification performed (all against disposable throwaway fixture repos under the session
scratchpad, fully outside dev-env's own path hierarchy — no test content ever touched the real
`brownm09/dev-env` remote):

1. **Repo-wide grep** confirming neither `research/SKILL.md` nor `journal-compose/SKILL.md`
   retains a direct write path to `~/.claude/skills/sources.md` — only the unchanged Pass-1
   **grep** reads remain, plus the new "do not write directly" warnings.
2. **Initial live dry run** of the first implementation's command sequence: fresh-creation path,
   `Edit` on the worktree's absolute path, commit, push, and a simulated second invocation
   confirming worktree reuse — all succeeded and confirmed not blocked by the live
   `pre-tool-use-canonical-mutate-guard.py` hook (`worktree`/`fetch`/`show-ref` aren't in its
   mutating-verb list; a `-C <worktree>` target resolves to the worktree's own root via
   `git rev-parse --show-toplevel`, not canonical, so its redirect-target check doesn't classify
   commits/pushes inside it as canonical-directed).
3. **Post-review re-verification of the corrected procedure**, run three separate times against
   fresh fixtures to specifically re-exercise every fixed finding:
   - Local-branch-first creation order, dedup-skip on a repeated entry, and the pathspec commit +
     verify step — all confirmed correct.
   - The stale-branch-after-sweep fix specifically: simulated a real squash-merge (`git merge
     --squash` + `push --delete`) against the fixture's queue branch, mirroring this repo's actual
     `gh pr merge --squash --delete-branch` convention. This caught **two additional bugs in my
     own fix**, not flagged by either review subagent: the originally-planned
     `merge-base --is-ancestor` check never fires after a squash merge (the squashed commit isn't
     a descendant of the original branch — dead code that looked like protection but wasn't), and
     a narrow single-branch `fetch --prune` silently fails to prune anything once that branch no
     longer exists remotely. Replaced with a push-then-delete signature (upstream tracking
     configured, but the remote-tracking ref gone after a full `fetch --prune`), then re-verified:
     a second invocation after the simulated sweep correctly detects it, deletes the stale local
     branch, and recreates fresh from the now-updated `origin/main`.
   - Confirmed throughout, across all three runs: the fixture's own "canonical" checkout stayed
     clean (no tracked-file changes), and — after cleanup — the real `brownm09/dev-env` repo has
     no stray `research-sources-queue` worktree and its canonical checkout is still clean.

## ADR-warrant

Warranted — new shared skill under `claude/skills/`, changes to two existing skills' documented
write behavior, and a workflow rule (worktree-isolate-not-canonical-write for `sources.md`)
future work should follow. New [ADR-101](docs/adr/101-source-library-writes-through-worktree.md),
citing ADR-013 (helper-skill pattern precedent), ADR-082 (why journal-compose's isolation doesn't
already cover this), ADR-071/ADR-024 (why neither existing guard hook already prevents this).
`docs/adr/INDEX.md` updated. Numbered 101, not 100 — `origin/main` advanced during this session's
investigation (PR #706 merged, claiming ADR-100 for an unrelated hook) before this branch was cut;
confirmed via `git log` against the fetched `origin/main` tip before writing the ADR.

## Doc reconciliation

`docs/REFERENCE.md`'s `/research` and `/journal-compose` entries updated in this PR (see Changes
above). No `README.md` Skills-table change — the new helper skill is internal-only, matching the
existing `sync-routine-worktree` precedent (also not listed there).

`/review` ([comment](https://github.com/brownm09/dev-env/pull/714#issuecomment-4940919430)) found
7 blocking + 4 non-blocking findings via two parallel Opus subagents (correctness/security +
reliability/maintainability). All fixed in commit `4bf0951`, with two further bugs in the fix
itself caught by empirical re-verification (see Test plan item 3):

**Blocking — all fixed:**
1. **Branch-resolution order bug** (`-b` hard-fails once a local branch exists — breaks every
   invocation from the second onward). **Fixed** — check local branch before remote.
2. **Command injection** in the untrusted-content Bash fallback. **Fixed** — values now passed
   via environment variables into a fixed `node -e` script, never string-interpolated.
3. **Weak liveness/orphan detection** (`.git` existing alone was insufficient). **Fixed** —
   requires `git rev-parse --show-toplevel` to also resolve, with cleanup before recreation.
4. **Dedup gap** (Pass-1 grep can't see queued-but-unmerged entries). **Fixed** — added a second
   dedup pass against the queue worktree's own copy.
5. **Unconditional success return** (no verification the commit landed). **Fixed** — added a
   verification step; SUCCESS is only ever returned after it passes.
6. **Unhandled concurrency** (shared worktree/index race with no guard). **Fixed** — a
   lock/collision failure retries the whole sequence once before reporting failure.
7. **Missing commit pathspec** (bare `git commit`, contradicting this repo's own `CLAUDE.md`
   convention for a shared checkout). **Fixed** — commit now carries
   `-- claude/skills/sources.md`.

**Non-blocking — all fixed:**
8. **`SECTION` exact-match fragility.** **Fixed** — caller must copy the heading verbatim; match
   is now case-/whitespace-insensitive.
9. **Commit-message type mismatch** (`research:` for a `journal-compose` caller). **Fixed** —
   changed to the neutral `chore:` type.
10. **`CALLER` interpolation, latent risk.** **Fixed** — Parameters table now documents `CALLER`
    must be a fixed literal.
11. **Stale-branch resurrection after a sweep.** **Fixed** — see Test plan item 3 above for the
    corrected detection signal and its own re-verification.

Closes #708


brownm09:
## Review: fix(skills): route source-library writes through a worktree, not canonical

### Summary

The core goal — never write `sources.md` through the canonical-resolving junction — is achieved
and independently confirmed safe against both mutate-guarding hooks. However, two parallel
subagent passes over the new `queue-source-library-entry/SKILL.md` procedure (the actual
executable bash sequence a future LLM invocation will follow, often fully unattended) found real
functional bugs in it: a branch-resolution ordering bug that permanently breaks the skill on its
second real-world run, a weak liveness check that mis-handles the exact orphan cases it claims to
cover, an unconditional "success" return with no verification the commit landed, a concurrency
race with no guard, a dedup gap that guarantees duplicate entries accumulate, and a command-injection
risk in the documented Bash fallback for untrusted web-sourced content. **Needs revision** before
merge — not a small-tweak PR, since several of these compound each other (the concurrency race and
the missing pathspec, in particular). None require reducing scope; all are fixable within the same
file.

---

### Blocking Findings

**[correctness]** `claude/skills/queue-source-library-entry/SKILL.md` Step 3
The branch-creation logic checks `refs/remotes/origin/$QUEUE_BRANCH` first and runs `worktree add
-b "$QUEUE_BRANCH" ...` whenever it exists — but `-b` creates a *new* local branch and hard-fails
if one by that name already exists locally.
After the first successful run, the local branch persists in the dev-env canonical `.git` even
after the worktree directory is later removed (`worktree prune` clears the worktree registration,
never the branch ref). Every subsequent invocation then hits this `-b` failure and, per this same
step's own "do not fall back to the direct write" rule, silently stops queuing entries until a
human manually deletes the local branch. This is the second-invocation-onward path, not a rare one.
**Fix:** Check `refs/heads/$QUEUE_BRANCH` first and attach with the no-`-b` form
(`worktree add "$QUEUE_WORKTREE" "$QUEUE_BRANCH"`) when it exists; use `-b ... origin/$QUEUE_BRANCH`
only when the local branch is absent but the remote has it; `-b ... origin/main` only when neither
exists. Use `git fetch --prune` so a deleted remote branch can't leave a stale ref.

**[security]** `claude/skills/queue-source-library-entry/SKILL.md` Step 5 (Edit-blocked fallback)
The fallback is specified only as "e.g. `node -e` reading, splicing, and rewriting the file ... or
a targeted `sed`/`py -3` invocation," with no concrete snippet.
`ENTRY_MARKDOWN` originates from a research subagent's web-search output — arbitrary page titles
and text, attacker-influenceable. An agent following this guidance literally will most naturally
interpolate that untrusted string directly into the `-e`/`sed` command string, so a crafted title
containing quotes, backticks, or `$(...)` yields shell/JS command injection the moment this
fallback path is actually exercised (reached whenever the invoking session's own cwd is a dev-env
sibling worktree — a real, recurring condition for dev-env development sessions running
`/research`).
**Fix:** Mandate passing `ENTRY_MARKDOWN`/`SECTION` via environment variables (`process.env.*` /
`os.environ`) or a temp file read from disk — never string-interpolated into the command — and
provide the concrete safe snippet instead of "e.g."

**[correctness]** `claude/skills/queue-source-library-entry/SKILL.md` Steps 2–3
Step 2 treats mere existence of `${QUEUE_WORKTREE}/.git` as proof the worktree is "live," but that
is weaker than the liveness check this repo's own `pre-tool-use-worktree-path-check.py` uses
(`git rev-parse --show-toplevel` resolving back to the worktree). The two orphan shapes Step 2
claims to cover aren't actually handled: a present-but-disconnected `.git` link routes to reuse and
then fails at the Step 6 commit ("not a git repository"); a missing `.git` link with the directory
still present makes Step 3's `worktree add` fail ("already exists") since `worktree prune` never
removes the directory itself.
**Fix:** Verify liveness with `git -C "$QUEUE_WORKTREE" rev-parse --show-toplevel` succeeding (not
just `.git` existing); on failure, `rm -rf "$QUEUE_WORKTREE"` before `worktree add` in Step 3.

**[correctness]** `claude/skills/queue-source-library-entry/SKILL.md` Scope boundary + both callers
Dedup relies entirely on the caller's Pass-1 grep of `~/.claude/skills/sources.md`, which resolves
to the canonical checkout (`origin/main`) — it structurally cannot see entries already sitting
unmerged on `chore/research-sources-queue`. Any source found again before a human sweeps the
branch passes the "not already present" check and gets re-appended.
**Fix:** Add a second dedup pass that greps the *queue worktree's own* `sources.md` before
appending, so an already-queued-but-unmerged entry is recognized and skipped.

**[reliability]** `claude/skills/queue-source-library-entry/SKILL.md` Step 8
Step 8 unconditionally "Return[s] SUCCESS," and no step defines failure handling for Steps 4–7 (only
Step 3 does). If Step 6's commit fails silently (e.g. "nothing to commit" from a no-op edit, or the
concurrency race below) the skill still reports "queued" to the user for an entry that was never
actually committed.
**Fix:** Verify the commit landed (e.g. `git -C "$QUEUE_WORKTREE" log -1` shows the expected new
commit) before returning SUCCESS; return an explicit FAILURE with reason otherwise — never a silent
false-positive SUCCESS.

**[reliability]** `claude/skills/queue-source-library-entry/SKILL.md` Steps 2–6
The design forces every caller onto one shared worktree + branch with no mutual exclusion, yet the
unattended daily `journal-compose` run can genuinely overlap an interactive `/research` invocation.
Two concurrent invocations racing the same fixed path/index can each fail differently (`worktree
add` "already exists," or `index.lock`/"nothing to commit" on the shared index) — for the
unattended producer this is silent, unrecoverable data loss of a found citation.
**Fix:** On a lock/already-exists collision, re-run the Behavior sequence from Step 2 once before
reporting FAILURE — don't retry indefinitely, and don't report SUCCESS on the losing side.

**[reliability]** `claude/skills/queue-source-library-entry/SKILL.md` Step 6
The commit is a bare `git commit -m ...` with no `-- <pathspec>`. dev-env's own `CLAUDE.md`
explicitly mandates an explicit pathspec on every commit into a file shared across concurrent
writers, precisely because a bare commit sweeps in whatever else happens to be staged — which is
exactly the shared-index scenario the concurrency finding above describes.
**Fix:** `git -C "$QUEUE_WORKTREE" commit -m "..." -- claude/skills/sources.md`.

---

### Non-Blocking Findings

**[maintainability]** `claude/skills/queue-source-library-entry/SKILL.md` Parameters / Step 5
`SECTION` must match an existing `## ${SECTION}` heading *exactly*, or a new section is created. A
near-miss from an LLM-chosen value at runtime (case, pluralization, trailing whitespace) silently
fragments `sources.md` into near-duplicate sections over time.
**Suggestion:** Require the caller pass the heading verbatim from its own Pass-1 grep output, and
make the heading match case-/whitespace-insensitive before deciding no section fits.

**[maintainability]** `claude/skills/queue-source-library-entry/SKILL.md` Step 6
The commit message hardcodes a `research:` conventional-commit type regardless of `CALLER` — when
`CALLER=journal-compose`, the queue branch's own log reads "research: ... (via journal-compose)."
**Suggestion:** Use the neutral `chore:` type matching the `chore/` branch prefix instead of a
type that's wrong for one of the two documented callers.

**[security]** `claude/skills/queue-source-library-entry/SKILL.md` Step 6 / Parameters
`${CALLER}` is interpolated into the commit message's `-m` value. Not exploitable today — both
documented callers pass fixed literals — but the skill's own "invoked by other skills" framing
invites future callers, and a dynamic `CALLER` would be a shell-injection vector.
**Suggestion:** State in the Parameters table that `CALLER` must be a fixed literal from this
document, never derived from external content.

**[reliability]** `claude/skills/queue-source-library-entry/SKILL.md` Scope boundary
The queue worktree is synced to `origin/main` only at creation, and Step 7 pushes unconditionally.
After a human sweeps the branch into a merged, deleted PR branch, the next invocation still reuses
the stale local branch and Step 7 re-creates the just-deleted remote branch carrying the
already-merged commits plus the new one.
**Suggestion:** Before reusing an existing local branch, check whether its tip is already an
ancestor of `origin/main` (`git merge-base --is-ancestor`); if so, treat it as already swept and
recreate fresh from `origin/main` instead of reusing.

---

### Findings Disposition

To be added after fixes are applied in this same PR.

---

*Reviewed by `claude-sonnet-5` via Claude Code — Step 5 parallel subagent analysis (correctness/security + reliability/performance/maintainability, both `claude-opus-4-8`)*

PR #722, opened 2026-07-11
fix(hooks): fail-closed gate crash guards — module-level + __main__ → exit 2 (Phase A)

brownm09:
Closes #718. Part of the hook-reliability initiative #717 (**Phase A — urgent hardening**).

## Problem

Both fail-closed PreToolUse gates crashed to fail-**OPEN**:

- `pre-auto-merge-checkpoint-gate.py` ran an **unguarded module-level** `importlib` `exec_module` of `pre-merge-findings-gate.py`. A broken sibling raised at import time — before `main()` — exiting 1.
- Neither gate had a `__main__` try/except, so any unhandled exception in `main()` also exited 1 (e.g. a malformed `gh pr view --json` whose `commits[-1]` is not a dict).

Claude Code reads any **non-2 exit as "the hook allowed the tool"**, so both paths silently ungated `gh pr merge --auto` / a same-day `/journal-compose` — the exact class #717 exists to close.

## Fix

- **`pre-auto-merge-checkpoint-gate.py`**: wrap the dependency load (`from _hookio import …` + the dynamic `exec_module` + attribute extraction) in `try/except → sys.exit(2)`; add a `__main__` `try/except → sys.exit(2)`. Both ASCII-sanitize the reason (cp1252-stderr).
- **`pre-tool-use-journal-compose-force-guard.py`**: add a `__main__` `try/except → sys.exit(2)` (JSON-reason channel, matching `_emit_block`). **Its top-level imports deliberately stay fail-OPEN** — it is registered on *every* Bash call, so a broken `_hookio` must not exit 2 and block **all** Bash; only a crash *inside* `main()` (reached solely once a same-day compose target has matched) fails closed.
- Both `__main__` guards **re-raise `SystemExit`** so the gate's own deliberate exit 0 (allow) / exit 2 (block) verdicts are never reclassified.
- **Docs**: new authoring invariant **#5 "declared fail direction"** in `docs/REFERENCE.md` → Hooks → Authoring rules; addenda to **ADR-083** and **ADR-096**; INDEX rows updated; global `claude/CLAUDE.md` Hook-Safety pointer de-stale-counted; dev-env `CLAUDE.md` Testing items 52/55 updated.

### Load-surface asymmetry (deliberate)
The auto-merge gate fails its **dependency load closed** (a broken sibling blocking all Bash is loud + CI-caught; silently ungating `--auto` is the worse outcome). The compose-force gate keeps its **imports fail-open** (broader per-call surface, far less severe ungated action). Documented in both ADR addenda and REFERENCE #5.

## Testing

| Gate | Result |
|---|---|
| Item 1 — hook-script syntax (`ast.parse`, all scripts) | **SYNTAX OK** |
| Item 39 — AST crude-command-substring check | Tests: **13 passed, 0 skipped, 0 failed** (0 unexpected/stale) |
| Item 52a — `test_pre_auto_merge_checkpoint_gate.py` (pure) | Tests: **33 passed, 0 skipped, 0 failed** |
| Item 52b — `test-auto-merge-checkpoint-gate.sh` (e2e) | **17 passed, 0 failed** (was 15; +`[16]` crash-after-trigger, +`[17]` corrupted sibling import) |
| Item 55 — `test_pre_tool_use_journal_compose_force_guard.py` | Tests: **45 passed, 0 skipped, 0 failed** (was 44; +`test_e2e_crash_after_trigger_fails_closed`) |

- **Coverage gate**: new crash-guard behavior is covered by the three new e2e cases above (each asserts the pre-fix fail-OPEN exit 1 is now a fail-CLOSED exit 2). File 2 needed a small env-gated, production-inert fault-injection seam because the gate is otherwise fully defensive (no natural runtime-crash vector); `_run_hook` pops that var from the inherited env so only the opt-in test trips it.
- **Suppression grep**: clean (no policy-listed suppressions). Note: the crash guards use `except Exception … # noqa: BLE001` — a *blind except is the correct, necessary shape for a crash guard* (catch anything, fail closed); it is not one of the policy's listed suppression forms and does not match the Rule-3 grep.
- **Test-integrity grep**: clean; no deleted test files; no lowered thresholds.
- Baseline-failure tracking: not enabled for dev-env (no `baseline_test_failure_tracking`); N/A.

## Risk-dimension audit
- **Testing** — covered (above). **Observability** — the fix *is* an observability improvement: crashes now emit an ASCII/JSON reason on the exit-2 stderr channel the model reads, instead of an invisible fail-open. **Security** — strengthens a safety gate (prevents ungated `--auto`); no authz/secret surface. **Resilience** — this is the failure-mode fix; revert restores prior fail-open. **Performance** — a `try/except` (zero happy-path cost) + one inert env-var check. **Data integrity/migrations** — N/A (no data/schema).

## ADR-warrant & doc-reconciliation
Amended ADR-083 + ADR-096 (no new ADR — the initiative ADR is PR2). REFERENCE authoring rule + INDEX + both CLAUDE.md files + Testing items 52/55 reconciled in this PR. README hook one-liners remain accurate ("fails closed" still holds).

Both findings from the posted [review comment](https://github.com/brownm09/dev-env/pull/722#issuecomment-4941735588) are resolved:

- **[reliability] `import _winsubp` outside the fail-closed guard** — **fixed in `4bf45dc`**: moved `import _winsubp` to the first line inside the dependency-load `try` block so all sibling imports fail closed uniformly. Re-verified after the fix: `test_pyw_stdio.py` 7 passed (its static `_winsubp` check is whitespace-tolerant), file-1 pure 33 passed, file-1 e2e 17 passed, benign smoke exit 0.
- **[maintainability] Step 2e language-scope note (Python/shell test files)** — **verified, no change needed**: the diff adds only new test cases (no skip markers, deleted blocks, or lowered thresholds); the pre-PR test-integrity grep was clean.


brownm09:
## Review: fix(hooks): fail-closed gate crash guards — module-level + __main__ → exit 2 (Phase A)

### Summary
**Merge with one minor fix.** A tight, well-tested hardening PR: it converts two fail-closed PreToolUse gates from crash-to-fail-OPEN (exit 1) to crash-to-fail-CLOSED (exit 2), guarding both the module-level dependency load and `main()`, with `SystemExit` correctly re-raised so the gates' own deliberate allow/block verdicts are never reclassified. The deliberate import-fail-open asymmetry for the globally-registered compose-force gate is sound and well-documented. The one substantive finding is a coverage gap in the PR's *own* thesis: file 1's `import _winsubp` sits just outside the new import guard. No blocking findings.

---

### Non-Blocking Findings

**[reliability]** `claude/scripts/pre-auto-merge-checkpoint-gate.py:44`
`import _winsubp` is the one *sibling*-module import left outside the new fail-closed dependency-load `try/except`, so a `_winsubp` import failure would raise before `main()` and exit 1 = fail-OPEN — the exact behavior this PR exists to eliminate. Real-world risk is ~nil (`_winsubp` is stdlib-only, and `_hookio` — imported inside the guard — re-imports it anyway), but leaving it out makes the guard's coverage inconsistent with its stated goal.
**Suggestion:** Move `import _winsubp` to the first line inside the `try` block. The `_winsubp` static check in `test_pyw_stdio.py` is leading-whitespace-tolerant (`^\s*(?:import\s+_winsubp...)`), so indenting it is safe. The stdlib imports correctly stay outside — they cannot fail.

**[maintainability]** `claude/scripts/tests/` (Step 2e language scope)
The two changed test files are Python/shell; the review's automated test-integrity scanners target JS/TS frameworks and cannot inspect them, so a clean automated result is not evidence for these files.
**Suggestion:** Confirmed manually — the diff adds only new test cases (no skip markers, deleted blocks, or lowered thresholds), and the pre-PR test-integrity grep was clean. No action needed beyond this note.

---

### Style Notes
The two crash guards deliberately use different ASCII-safety mechanisms — file 1 `.encode("ascii","replace")` via `_fail_closed`, file 2 `json.dumps(ensure_ascii=True)` via the `_emit_block`-matching channel. Each matches its own file's existing emit convention (the right call for a minimal hardening PR); PR2's `_hookout.ascii_sanitize()` will unify them.

---

### Findings Disposition
_To be recorded in the PR body before merge (both non-blocking findings fixed-in-PR or verified)._

---

PR #725, opened 2026-07-11
docs: document bare --force after rebase auto-closes open PRs

brownm09:
## Summary

- Adds `### Bare \`--force\` after rebase auto-closes any open PR on the target branch` to `claude/CLAUDE.md` → Git Workflow section
- Adds a matching detailed runbook entry to `docs/REFERENCE.md` → Git Workflow Runbooks (between the stacked-PR section and "Separate clones")

## What this documents

**Failure mode (incident: win11-init-tools PR #34 → PR #46, 2026-07-11):**

After `git rebase origin/main`, `git push --force-with-lease` rejects with `(stale info)` because the rebase rewrote local commits without refreshing the remote-tracking ref. Falling back to bare `git push origin HEAD:<branch> --force` succeeds at the network level — but GitHub interprets it as a branch-delete + branch-recreate, **auto-closing any open PR** targeting that branch (`mergedAt: null`, `mergeCommit: null`, `gh pr reopen` fails).

**Root cause:** `--force-with-lease` is working correctly — the tracking ref is genuinely stale. The fix is `git fetch origin` after the rebase to refresh it, then `git push --force-with-lease` again.

## Testing

No automated test suite. Manual verification:
- Parse check: both files are Markdown; no parse tool applies.
- Content review: confirmed both the `## Git Workflow` H3 and the REFERENCE.md runbook landed at the correct positions, with accurate symptom/fix/recovery information matching the incident.

Tests: N/A — documentation-only change.

## ADR-warrant check

This change adds a new documented failure mode and recovery runbook — no new rule, skill, hook, or settings change. No ADR warranted.

## Suppression/test-integrity check

Documentation-only. No suppressions, no test changes.

Closes #724

### DIFF
--- a/claude/CLAUDE.md
+++ b/claude/CLAUDE.md
@@ -148,6 +148,27 @@ gh api -X DELETE "repos/{owner}/{repo}/git/refs/heads/<branch>"   # pure REST re
 
 Confirmed as a **general git-worktree mechanic**, not a quirk of any one repo's worktree count: [dev-env#575](https://github.com/brownm09/dev-env/pull/575) (canonical correctly on `main`) and [lifting-logbook#664](https://github.com/brownm09/lifting-logbook/pull/664) (an idle worktree left squatting `main` by an earlier merge; that repo's CLAUDE.md "Standard Issue Workflow" step 8 carries its own copy of this same two-step pattern). Full detection, root cause, and non-destructive auto-correction (parking idle squatters instead of removing them): dev-env [ADR-058](../docs/adr/058-worktree-squatting-main-detection-correction.md); complete runbook: dev-env [`docs/REFERENCE.md` → Git Workflow Runbooks](../docs/REFERENCE.md#git-workflow-runbooks).
 
+### Bare `--force` after rebase auto-closes any open PR on the target branch
+
+**Pattern:** After `git rebase origin/main`, `--force-with-lease` rejects with `(stale info)` because the rebase rewrote local commits without refreshing the remote-tracking ref. Falling back to bare `git push origin HEAD:<branch> --force` (or `--force`) succeeds at the network level — but GitHub interprets it as a branch-delete + branch-recreate. **Any open PR targeting that branch is auto-closed** with `mergedAt: null`, `mergeCommit: null`, and `gh pr reopen` fails with "Could not open the pull request."
+
+**Symptom:** `! [rejected] HEAD -> <branch> (stale info)` from `--force-with-lease`, then a bare `--force` push succeeds, and the PR shows `state: CLOSED` with no merge commit.
+
+**Fix:** After any rebase, run `git fetch origin` *before* retrying `--force-with-lease` — never fall back to bare `--force`:
+```bash
+git rebase origin/main
+git fetch origin              # refreshes the stale remote-tracking ref
+git push --force-with-lease   # now succeeds without closing the PR
+```
+
+**Recovery (PR already auto-closed):** `gh pr reopen` fails with "Could not open the pull request." Create a replacement PR:
+```bash
+gh pr create --head <branch> --base main --title "..." \
+  --body "Replaces #N (auto-closed by bare --force after rebase; [context](https://github.com/.../pull/N))"
+```
+
+Motivating incident: [win11-init-tools PR #46](https://github.com/brownm09/win11-init-tools/pull/46) — PR #34 was auto-closed and replaced. Full runbook: [`docs/REFERENCE.md` → Git Workflow Runbooks](../docs/REFERENCE.md#git-workflow-runbooks).
+
 ---
 
 ## Dev-Env & Project Boards
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -876,6 +876,82 @@ Motivating incident: career-playbook [#587](https://github.com/brownm09/career-p
 (Step 4.7) / [#591](https://github.com/brownm09/career-playbook/pull/591) (Step 4.8, which superseded
 the orphaned #588).
 
+### Bare `--force` after rebase auto-closes any open PR on the target branch
+
+**Trigger.** A `git rebase origin/main` rewrites local commits but does **not** refresh the remote-tracking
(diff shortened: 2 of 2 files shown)


brownm09:
## Review: docs: document bare --force after rebase auto-closes open PRs

### Summary

The core guidance is correct — `git fetch origin` after rebase before retrying `--force-with-lease`, never fall back to bare `--force` — and the fix sequence is actionable. Two blocking correctness issues in the Diagnosis section: `timelineItems` is not a valid `gh pr view --json` field (the recovery diagnostic command would immediately error), and the `(stale info)` / `(fetch first)` message distinction is factually wrong for `--force-with-lease` pushes (both stale-tracking-ref and concurrent-push scenarios produce `(stale info)` when using `--force-with-lease`). Two non-blocking issues around the accuracy of the mechanism explanation. Fix the blocking items before merging.

---

### Blocking Findings

**[correctness]** `docs/REFERENCE.md:936`

`gh pr view <N> --json state,closedAt,mergedAt,mergeCommit,timelineItems` fails immediately — `timelineItems` is not a valid field for `gh pr view --json`. Verified: running the command against this PR produces `Unknown JSON field: "timelineItems"` and lists available fields. A user following the recovery steps would hit this error before they could confirm anything about the PR's state.

**Fix:** Remove `timelineItems` from the field list. The confirmation you need — "this was an auto-close, not an intentional close" — is fully captured by `state=CLOSED, mergedAt=null, mergeCommit=null`:
```bash
gh pr view <N> --json state,closedAt,mergedAt,mergeCommit
# state=CLOSED, mergedAt=null, mergeCommit=null confirms the auto-close
```

---

**[correctness]** `docs/REFERENCE.md:907`

The Diagnosis section states: "A concurrent push leaves `(fetch first)` in the rejection message; `(stale info)` means the *local* tracking ref is out of date."

This is factually wrong in the `--force-with-lease` context. `(fetch first)` is the rejection message for a regular `git push` (no force flag) when the remote has advanced non-fast-forward — it is never produced by `--force-with-lease`. For `git push --force-with-lease`, the lease compares the local remote-tracking ref against the actual remote state; if they differ, git always rejects with `(stale info)` regardless of why they differ, including a concurrent push. A user trying to distinguish "stale tracking ref after rebase" from "genuine concurrent push" cannot do so from the rejection message alone — both produce `(stale info)`.

The correct diagnostic after a `(stale info)` rejection from `--force-with-lease` is: run `git fetch origin`, then inspect `git log origin/<branch>..HEAD` (your commits not yet on remote) and `git log HEAD..origin/<branch>` (remote commits not in your local). If the second log is non-empty, someone else pushed; if only the first is non-empty, the tracking ref was simply stale (safe to push with `--force-with-lease` now).

**Fix:** Replace the `(stale info)` / `(fetch first)` distinction in the Diagnosis paragraph. For example:

> **Diagnosis.** The `(stale info)` rejection from `--force-with-lease` means the local remote-tracking ref differs from what the remote actually holds. This happens in two cases: (a) the tracking ref was never refreshed after the rebase — the most common case — or (b) someone else pushed to the branch since your last fetch. Both look identical at the rejection message level. After running `git fetch origin`, inspect `git log HEAD..origin/<branch>` — if it is non-empty, there are concurrent commits to reconcile before force-pushing; if empty, the stale tracking ref was the only issue and `git push --force-with-lease` will now succeed.

The existing "safe test" code block can stay; only the prose before it needs updating.

---

### Non-Blocking Findings

**[correctness]** `docs/REFERENCE.md:881-887` and `claude/CLAUDE.md:153-154`

The Trigger (REFERENCE.md) and Pattern (CLAUDE.md) sections assert that the `(stale info)` rejection is a natural consequence of rebasing because "the rebase rewrote local commits without refreshing the remote-tracking ref." This is incomplete. After a rebase where nobody else pushed, the remote-tracking ref and the actual remote both point to the same pre-rebase commit — the lease check passes and `--force-with-lease` succeeds without any rejection. For `(stale info)` to occur, the remote must have advanced since the last fetch. The rebase alone doesn't produce the error.

**Suggestion:** Clarify that the failure requires the remote to have advanced since the last fetch — e.g., "After `git rebase`, if the remote branch has advanced since your last `git fetch` (a previous push from another machine, or a concurrent collaborator push), `--force-with-lease` rejects with `(stale info)` because the lease is based on a stale tracking ref."

---

**[correctness]** `docs/REFERENCE.md:884-887`

"GitHub treats a force-push whose new tip does not descend from the old tip as a **branch-delete + branch-recreate**" is likely not the accurate mechanism. GitHub's documented behavior is that PRs remain open on force-push (even non-fast-forward); they are auto-closed only on branch deletion. The incident did produce a closed PR with `gh pr reopen` failing, which is the fingerprint of a branch-deletion close — but the specific mechanism that turned the force-push into a branch-delete is not established here or in the issue.

**Suggestion:** Hedge the mechanism claim until the exact trigger is known — e.g., "GitHub auto-closes PRs when their head branch is deleted; the precise mechanism by which this force-push triggered that close is not fully understood (see [dev-env#724](https://github.com/brownm09/dev-env/issues/724) for the incident details)." The prevention and recovery advice remain correct regardless of the exact mechanism.

---

### Style Notes

`docs/REFERENCE.md:942` — The recovery body template `"<same body>\n\nReplaces #N..."` uses a double-quoted shell string; `\n` in double-quotes is a literal backslash-n, not a newline. Users copying verbatim get `\n\n` characters in the PR body instead of a blank line. Either switch to `$'...\n\n...'` (ANSI-C quoting) or annotate that `\n\n` is a formatting placeholder to replace with an actual blank line.

---

### Findings Disposition

Per ADR-028, each finding below must be fixed in this PR (with the fixing commit noted) or filed as a tracked issue and linked here before merge:

| # | Finding | Required action |
|---|---------|-----------------|
| B1 | `timelineItems` invalid field in recovery command | Fix in PR |
| B2 | `(stale info)` / `(fetch first)` distinction wrong for `--force-with-lease` | Fix in PR |
| N1 | Trigger/Pattern overstate when `(stale info)` occurs | Fix in PR or file issue |
| N2 | "branch-delete + branch-recreate" mechanism uncertain | Fix in PR or file issue |

---

PR #730, opened 2026-07-11
ci: hook/script test-suite runner + windows-latest PR gate (PR4 of #717)

brownm09:
## What & why

**PR4 (Phase B) of the hook-reliability initiative** ([#717](https://github.com/brownm09/dev-env/issues/717)). PR3 ([#726](https://github.com/brownm09/dev-env/pull/726)) added three mechanical enforcement gates (`test_hook_output_contract`, `test_hook_safe_exit_guard`, `test_settings_hook_wiring`) but nothing *ran* them outside a local `## Testing` command an author has to remember. This PR makes the whole ~60-item suite an external gate on every PR.

- **`claude/scripts/run-hook-tests.py`** — glob-discovers every `claude/scripts/tests/test_*.py` plus the bash gates in `claude/scripts/tests/` and `claude/hooks/tests/`, runs each as a subprocess, reports pass/fail/skip, and exits non-zero iff any failed. Discovery is glob-based so a new test file is gated automatically — no hand-maintained list to drift from `## Testing`. `--list` / `--timeout N` supported.
- **`.github/workflows/hook-tests.yml`** — `windows-latest`, **`pull_request` trigger only** (never on `main` pushes, per the lockfile-gate scoping lesson in the dependency policy). Faithful to the real `py -3` / Git Bash / cp1252 runtime ([ADR-007](https://github.com/brownm09/dev-env/blob/main/docs/adr/007-hook-command-invocation.md)). Sets a throwaway git identity (many tests spin up fixture repos and commit); grants **no gh token**, so the network- and shellcheck-dependent gates self-skip.

### Runner-hostile test — documented skip, not weakened
Per plan gotcha #7, the runner **runner-skips only `test_pyw_stdio.py`** (it spawns real `pythonw.exe` to probe Windows-subsystem stdio under parent-supplied pipes — unhostable on a headless CI runner). The skip is in a small, unit-test-pinned `SKIP_TESTS` list with an in-place reason; the test is unchanged and still runs locally. A bash gate's own `SKIP:` self-exit (no shellcheck, unauthenticated gh) is treated as non-failing.

## Testing

Full suite via the new runner, locally on Windows (identical env to CI — `py -3`, Git Bash, cp1252):

```
Tests: 63 passed, 2 skipped, 0 failed (114.0s)
```

- **Skips (test-integrity Rule 2 justification):**
[...]

PR #733, opened 2026-07-11
fix(hooks): restore ntpath.isabs pre-3.13 semantics in cd/redirect path resolvers (dev-env#732)

brownm09:
## Summary

Python 3.13 changed `ntpath.isabs` so a rooted-but-driveless path (`/Git/dev-env`, `\Git\dev-env`) returns `False` where ≤3.12 returned `True` ([What's New in 3.13 → os.path](https://docs.python.org/3/whatsnew/3.13.html#os-path)). Three shared cd-chain / redirect resolvers decide absoluteness via `os.path.isabs` and join the target onto `cwd` when it reads as relative:

- `effective_merge_dir` (`_hookio.py`, ADR-067)
- `_effective_push_dir` (`pr-merge-reminder.py`, ADR-065)
- `_blockable_redirect_root` (`pre-tool-use-canonical-mutate-guard.py`, ADR-071)

On 3.13+ a forward-slash `cd`/`-C` target wrongly takes the relative branch — `normpath(join(cwd, "/Git/dev-env"))` → `\Git\dev-env` — silently **mis-scoping the ADR-065/067 merge/push hooks** and **fail-opening the canonical-mutate guard** (a `git -C /repo checkout` redirect into a canonical checkout resolves to a wrong dir, its `git rev-parse` fails, the redirect is skipped, the mutating command is no longer blocked).

**Latent today** — prod `py -3` is 3.12 (correct). Surfaced by #730's `windows-latest` CI (3.13+), which flagged 12 failures across `test_hookio.py` (5) and `test_pr_merge_reminder.py` (7). Part of the hook-reliability initiative #717.

## Fix

- New shared pure helper `_hookio.is_absolute_path(path)` = `path.startswith(("/", "\\")) or os.path.isabs(path)` — restores ≤3.12 semantics on every version by short-circuiting on the leading separator (drive-absolute / UNC still flow to `os.path.isabs`, unchanged).
- All three resolvers swap `os.path.isabs(...)` → `is_absolute_path(...)`. Behaviour-identical on 3.12, correct on 3.13+. One helper, three call sites.

**Audited the two other `isabs` users — not sensitive, unchanged:** `post-tool-use.py` `_canonical_root_from_common_dir` (git emits drive-absolute/relative on Windows, never driveless-rooted; its `test_post_tool_use.py:442` already passes on 3.13) and `pre-tool-use-worktree-path-check.py` `file_path` guard (always a drive-absolute Windows path from the Write/Edit tool contract). The `normpath`-only sites are separator normalizers, unaffected by the `isabs` change.

## Testing

Ran the affected suites on the local 3.12 runtime — all green:

```
test_hookio.py                  Tests: 91 passed, 0 skipped, 0 failed
test_pr_merge_reminder.py       Tests: 56 passed, 0 skipped, 0 failed
test_canonical_mutate_guard.py  Tests: 64 passed, 0 skipped, 0 failed
```

Also ran every `test_*.py` in `claude/scripts/tests/` (56 files pass — see the pre-existing note below). The **3.13 fix itself is validated by this PR's `windows-latest` CI** (`hook-tests.yml`, `py -3` → 3.13+) — exactly where the 12 failures showed.

**Two failure shapes fixed:** the absolute-path cases (`cd /Git/dev-env` → assert `== "/Git/dev-env"`) are fixed by the function change alone (no test edit); the relative-path cases (`assert os.path.isabs(out)` on a driveless `\base\sub\repo`) are fixed by swapping those two assertions to the version-agnostic `is_absolute_path(out)`.

**Coverage (ADR-022):** `test_hookio.py` gains 5 `is_absolute_path` cases including a **3.13 simulation** that patches `os.path.isabs` to return `False` and proves the `startswith` short-circuit is interpreter-independent — the only way to pin the 3.13 behaviour on a 3.12 runtime, where the two branches are indistinguishable for `/x`.

**Suppression / test-integrity greps (ADR-026/029):** both clean — the diff only adds tests and strengthens two assertions (no suppressions, skips, deletions, or lowered thresholds).

**Pre-existing, unrelated failure (not blocking):** running `test_pyw_stdio.py` directly reports `run-hook-tests.py` uses subprocess without importing `_winsubp`. That file is **untouched** by this PR (see diffstat), the failure predates this branch, and `test_pyw_stdio.py` is the documented *runner-skip* — the `run-hook-tests.py` CI suite skips it, so it does not gate this PR. Filed as a follow-up.

## Risk audit

- **Observability** — N/A (no runtime; hook output streams / exit codes unchanged — pure path-resolution fix).
- **Security** — *closes* a guard fail-open (a forward-slash `-C`/`cd` redirect escaping the canonical-mutate block on 3.13); no new surface, no input-validation change.
- **Resilience** — helper is total (any str, incl. `""`); fallback behaviour unchanged; single-commit revert.
- **Performance** — O(1) `startswith`; no new subprocesses.
- **Data integrity / migrations, Accessibility** — N/A.

## Docs

ADR-050 **Amendment 22** + `INDEX.md` row (date + tags); `## Testing` items 13 / 28 / 33 reconciled. No README/REFERENCE changes (no hook added / removed / renamed).

`/review` (`claude-opus-4-8`) posted [a clean review](https://github.com/brownm09/dev-env/pull/733#issuecomment-4947552086) — **0 blocking, 0 non-blocking findings**, nothing to dispose. Checkpoints: ADR-warrant `written` (ADR-050 Amendment 22), doc-reconciliation `not-applicable` (modify-only — no hook/script added, removed, or renamed).

Closes #732


brownm09:
## Review: fix(hooks): restore ntpath.isabs pre-3.13 semantics in cd/redirect path resolvers (dev-env#732)

### Summary
**Merge-ready — 0 blocking, 0 non-blocking findings.** A small (194-line), well-scoped forward-compat fix. The new `_hookio.is_absolute_path` helper is pure and total, and correctly restores ≤3.12 `ntpath.isabs` semantics on every version by short-circuiting on a leading `/`/`\` before falling through to `os.path.isabs` (drive-absolute / UNC targets still handled by `isabs`, unchanged).

Verified during review:
- **≤3.12 no-op.** For every input, `is_absolute_path` is behaviourally identical to `os.path.isabs` on ≤3.12 (a leading `/`/`\` is exactly the set ntpath ≤3.12 already called absolute), so the three call-site swaps change nothing on the current production runtime — the fix only *adds back* the driveless-rooted cases 3.13 dropped.
- **Security swap is a restore, not a behaviour change.** In `_blockable_redirect_root`, a driveless-rooted `-C`/redirect like `git -C /repo` now yields `resolved = "/repo"` (used as-is) instead of the pre-fix `normpath(join(cwd, "/repo"))` → `\repo`. `resolved = "/repo"` is what ≤3.12 already produced and is more faithful to what git itself does with `/repo` (OS resolves the drive), so the guard correctly re-detects the canonical checkout on 3.13 instead of fail-opening. No false-block risk introduced.
- **Monkeypatch hygiene.** The 3.13-simulation test patches the process-shared `os.path.isabs`, restores it in a `finally`, and re-asserts restoration afterward — so it can't leak into sibling tests in the single-process runner. Sequential runner ⇒ no race.
- **Imports** stay alphabetically sorted in all four touched files; the `is_absolute_path` docstring is a raw string (`r"""`), correct given the literal backslashes.

CI (`hook-tests.yml`, windows-latest → 3.13+) is green — the 12 formerly-failing cases now pass, which is the real validation of the fix.

### Findings
None.

### Checkpoints
- **ADR-warrant:** `written` — ADR-050 Amendment 22 + `INDEX.md` row are present in the diff (ADR-011 criterion 1: touches hooks/scripts under `claude/`).
- **Doc-reconciliation:** `not-applicable` — the change *modifies* existing scripts; it does not add / remove / rename a hook or script, so the Documentation Maintenance table's README.md / docs/REFERENCE.md Hooks-table triggers aren't met. The correct internal docs (the ADR-050 amendment and `## Testing` items 13 / 28 / 33) are updated in this PR.
- **Test coverage (ADR-022):** satisfied — 5 new `is_absolute_path` cases incl. the 3.13 simulation.
- **Test integrity (ADR-029):** satisfied — Python-only diff (JS/TS scanners N/A); manually verified no skip markers, deleted tests, or lowered thresholds; PR body carries the `Tests: N passed, 0 skipped, 0 failed` summary lines.

---

PR #735, opened 2026-07-11
fix: allow stub-write pushes to draft/<today> without a compose marker

brownm09:
## Summary

- `segment_targets_today_compose` matched `draft/<today>` tokens in any non-message scan position, so `git push -u origin draft/<today>` from the engineering-journal canonical checkout (a stub-write) was incorrectly gated and blocked
- Adds `_is_compose_op(command, cwd, today)` as a secondary check in `main()` — returns True only when `compose-<today>` appears in the cwd path (session is inside the compose worktree) or in a non-message scan token (e.g. a `-C compose-worktree` path)
- Stub-writes — which only ever reference `draft/<today>`, never a compose worktree path — exit 0 (allowed) after this check; all existing compose-op patterns (worktree add with compose dest, `-C compose-<today> commit/push`) still block

Closes #728

## How it works

`_is_compose_op` uses two fast-path checks before re-parsing segments:

1. If `compose-<today>` is absent from both the cwd and the raw command string → definitely a stub-write, return False immediately (no segment parse needed)
2. If `compose-<today>` is in the cwd path → compose worktree session, return True
3. Otherwise → scan non-message tokens for `compose[-/]<today>` using the already-split segments; return True if found

The crash-guard fault-injection seam (from the 2026-07-10 addendum) is moved to after both checks so it still fires only when a genuine compose op is confirmed.

## Testing

Full suite after /review fixes: **64 passed, 2 skipped, 0 failed**

```
Tests: 64 passed, 2 skipped, 0 failed
```

2 skipped (pre-existing, environmental — not introduced by this PR):
- `test_pyw_stdio.py` — requires interactive `pythonw.exe`; unreliable on non-interactive runner (no attached console)
- `run-shellcheck.sh` — `shellcheck` not found in PATH on this machine

Suppression/test-integrity greps: clean.

11 new tests added:
- 8 Layer 1 `_is_compose_op` unit tests: compose cwd (hyphen + slash), `-C` compose path, stub push simple, stub push refspec, compose token in message prose (no match), empty cwd, worktree-add with compose dest
- 4 Layer 2 e2e: stub push allowed without marker; refspec push allowed without marker; push from compose cwd still blocked; push with `-C compose-<today>` still blocked
- 1 Layer 2 e2e (added in /review follow-up): chained `git add && git commit && git push -u origin draft/<today>` allowed without marker (production Stub file workflow step 7 form)

## ADR

Updated ADR-096 with a 2026-07-11 addendum documenting the false-positive, the fix, and the testing additions.

Review comment: https://github.com/brownm09/dev-env/pull/735#issuecomment-4949641470

- **[documentation] REFERENCE.md trigger condition stale** — fixed in fe1e519: extended Trigger column to document `_is_compose_op` secondary check and note stub-write pass-through
- **[maintainability] Docstring typo in `_command_has_compose_token`** — fixed in fe1e519: `compose[-/<today>` → `compose[-/]<today>`
- **[maintainability] No test for chained stub-write command** — fixed in fe1e519: added `test_e2e_chained_stub_write_allows_without_marker`
- **[maintainability] 2 skipped tests unexplained** — addressed above in Testing section


brownm09:
## Code Review — PR #735

**fix: allow stub-write pushes to `draft/<today>` without a compose marker**

---

### Pre-checks

- **Test coverage gate:** ✅ Bug fix includes 8 new Layer 1 unit tests + 4 new e2e tests — good coverage.
- **Test integrity (Step 2e):** Python source files changed; JS-only skip-marker scan is not applicable. Manual inspection finds no `pytest.mark.skip` / `pytest.skip()` / equivalent in the diff. `Tests: 63 passed, 2 skipped, 0 failed` is present in PR body. (Non-blocking note on the 2 skips below.)
- **ADR-warrant:** ✅ ADR-096 addendum present in diff — warranted and written.
- **doc-reconciliation:** ❌ `claude/scripts/pre-tool-use-journal-compose-force-guard.py` changed; `docs/REFERENCE.md` does not appear in the diff. The hook's Trigger condition column now describes stale behavior.

---

### Blocking

**[documentation] `docs/REFERENCE.md` hook trigger condition is now inaccurate**

The Hooks table row for `pre-tool-use-journal-compose-force-guard.py` currently ends its Trigger column with:

> "…reference a `draft/<today>` or `compose[-/]<today>` target — 'today' is this process's own `datetime.date.today()`, never anything the command claims"

After this fix, that description is incomplete: it implies that any `git push -u origin draft/<today>` would be gated. In fact, the hook now applies a secondary `_is_compose_op` check and only gates commands where `compose[-/]<today>` appears in the cwd path or in a non-message scan token. Stub-write pushes that only reference `draft/<today>` are allowed through. A reader of REFERENCE.md today would still believe their engineering-journal stub push would be blocked.

**Fix:** Extend the Trigger column text to note the secondary condition, e.g.:

> "…reference a `draft/<today>` or `compose[-/]<today>` target — 'today' is this process's own `datetime.date.today()`, never anything the command claims — **and** the command or cwd identifies a compose worktree context (`compose[-/]<today>` in the resolved cwd or in a non-message scan token via `_is_compose_op`); plain stub-write pushes that reference only `draft/<today>` pass through unchanged (dev-env#728)"

---

### Non-blocking

**[maintainability] Typo in `_command_has_compose_token` docstring**

The docstring reads: `"…contains compose[-/<today>."` — the character-class closing bracket `]` is missing (should be `compose[-/]<today>`). Minor but misleading to a future reader of the docstring.

**Fix:** `compose[-/<today>` → `compose[-/]<today>` in the `_command_has_compose_token` docstring.

---

**[maintainability] No e2e test for the chained stub-write command**

The actual production form (Stub file workflow, step 7) is a chained command: `git add ... && git commit -m "draft: 2026-07-11 session N" -- ... && git push -u origin draft/<today>`. The e2e tests cover the bare `git push` and the refspec form, but not the chained form that the Bash tool actually receives. The logic is correct (fast-path eliminates non-compose commands regardless of chaining), but the coverage gap means a future refactor to `split_top_level` or chaining-parse behavior could silently break this scenario with no failing test.

**Fix (optional):** Add a `test_e2e_chained_stub_write_allows_without_marker` e2e test covering the full chained command.

---

**[maintainability] 2 skipped tests unexplained individually in PR body**

The PR body has `Tests: 63 passed, 2 skipped, 0 failed` ✅, but the Test Integrity policy (Rule 2) requires the PR body to justify each skip individually when the count is non-zero. The 2 skips are known environmental ones (`test_pyw_stdio.py` — no attached console; `run-shellcheck.sh` — shellcheck not found), but they aren't named in the PR body.

**Fix (optional):** Add a line under the Testing section naming the 2 skips and why they're acceptable (e.g., "2 skipped: `test_pyw_stdio.py` (requires interactive pythonw.exe), `run-shellcheck.sh` (shellcheck not in PATH)").

---

### Summary

```
review-findings: 1 blocking, 3 non-blocking
adr_warrant: written
doc_reconciliation: missing
```

PR #746, opened 2026-07-12
fix(journal-compose): generalize Step 10.5 conflict recovery to diff-and-replay

brownm09:
## Summary

Closes #742. Generalizes journal-compose's Step 10.5 conflict-recovery path from a
hand-maintained allowlist of "known deletable file classes" to a genuine diff-and-replay
against the actual changes the draft branch made since it diverged from `origin/main`.

- Computes `git diff --no-renames --name-status $MERGE_BASE $PREV` (both already available in
  Step 10.5: `$MERGE_BASE` from the ADR-080 conflict probe, `$PREV` as the draft branch's
  pre-recovery tip) scoped to the composed project's `sessions/<project>/` tree plus `README.md`
  (or, in multi-project mode, every composed project's tree).
- Replays `A`/`M` as `git checkout $PREV -- <path>`, `D` as `git rm`. Any unexpected status
  prints a warning instead of being silently dropped.
- Removes the old explicit checkout list (composed journal, both READMEs, `open-prs.jsonl` /
  `open-prs/`) and the "2b" step that hand-re-derived and hand-removed the exact PR shard
  numbers Step 9.5 had already reconciled.
- Updates the Step 9.5 carry-forward note and the Step 11 post-merge shard-leak check comment
  to reflect the new mechanism.

dev-env#684 (closed 2026-07-09) fixed the identical symptom — a reconciled open-PR shard
resurrected by the old allowlist — but only for that one file class, via a one-off manual
recovery. This PR generalizes the mechanism itself, so a future file class (e.g. a
`sessions/<project>/reports/` deletion) is handled automatically with no SKILL.md edit.

## Verification

`claude/skills/journal-compose/SKILL.md` is a markdown skill with embedded bash — no dedicated
`claude/scripts/*.py` test target applies (see ADR-104's Trade-offs section). Verified the
diff-and-replay logic directly by simulating a full Step 10.5 recovery in a throwaway git repo:
a draft branch with a stub added-then-deleted (net-zero, correctly not replayed), a composed
journal added, a project README modified, and one of two open-PR shards reconciled-away —
against a `main` that advanced with an unrelated conflicting change. The replay correctly
reconstructed the composed journal, the modified README, and deleted only the reconciled shard
(leaving the still-open one in place) — reproducing the dev-env#684 fix generically instead of
via a hardcoded case.

## Testing

Ran the full hook test suite per dev-env's `## Testing`:
```
py -3 claude/scripts/run-hook-tests.py
```
Tests: 64 passed, 2 skipped, 0 failed (98.3s). No new failures; the 2 skips are pre-existing
(`test_pyw_stdio.py` runner-skip, `run-shellcheck.sh` — shellcheck not installed locally).

Suppression grep: clean. Test-integrity grep: clean. No `package.json` touched (lockfile policy N/A).

## ADR-warrant check

This change modifies a documented skill's conflict-recovery mechanism and has rationale git log
won't recover on its own — added [ADR-104](docs/adr/104-journal-compose-conflict-recovery-diff-and-replay.md)
and updated `docs/adr/INDEX.md`.

## Doc reconciliation

No skill/hook/script/routine was added, removed, or renamed, and the skill's invocation syntax
(`/journal-compose [YYYY-MM-DD] [--force]`) is unchanged — no README/REFERENCE.md updates
required.

## Risk-dimension audit (Pass 3)

1. **Testing** — N/A — markdown skill with no dedicated test harness; verified via a real
   throwaway-repo simulation (see Verification above) plus the full hook suite as a regression
   check.
2. **Observability** — N/A — this skill has no runtime beyond the user-visible bash it prints;
   the new `WARNING:` catch-all in the replay loop's `case` statement surfaces an unexpected
   diff status instead of silently mis-replaying it.
3. **Security** — N/A — no permission/authz/isolation boundary touched.
4. **Resilience / failure modes** — Improved: the diff-and-replay is a full reconstruction from
   known-good trees (never a patch application), so it cannot fail on stale context the way a
   `git apply`-based approach could (see ADR-104's Alternatives Considered).
5. **Performance** — N/A — a single `git diff --name-status` plus a small file-count loop,
   equivalent in cost to the checkouts it replaces.
6. **Data integrity & migrations** — This *is* the data-integrity fix: prevents resurrecting
   content (open-PR shards, and now any other file class) that a session legitimately deleted.

Per `/review` (0 blocking, 1 non-blocking):
- **[reliability]** A/M replay always takes `$PREV`'s content wholesale (not new behavior, but
  worth documenting explicitly) — **fixed in `8e4baf6`** (added a clarifying bullet to ADR-104's
  Trade-offs section).

## Tiles spawned this session

None — this PR fully resolves dev-env#742; no further out-of-scope follow-ups identified during
this work.


brownm09:
## Review: fix(journal-compose): generalize Step 10.5 conflict recovery to diff-and-replay

### Summary
Ready to merge. This is a clean, well-scoped fix: it replaces a hand-maintained allowlist in journal-compose's Step 10.5 conflict-recovery path with a genuine `git diff --name-status $MERGE_BASE..$PREV` replay, closing the class of bug dev-env#684 only patched for one file type (open-PR shards). The mechanism was verified directly against a throwaway repo simulating the exact recovery scenario, the accompanying ADR-104 is thorough (including three alternatives considered with concrete rejection reasons), and the diff is appropriately scoped — no unrelated changes.

---

### Non-Blocking Findings

**[reliability]** `claude/skills/journal-compose/SKILL.md` (both diff-and-replay blocks)
The replay unconditionally takes `$PREV`'s version for any `A`/`M` path via `git checkout "$PREV" -- "$FILEPATH"`, discarding whatever `origin/main` independently holds at that path (most relevant for the shared top-level `README.md`, which multiple projects' composes could touch on the same day). This is not a regression — the pre-existing allowlist did the identical unconditional `checkout "$PREV" -- ... README.md` — but the new diff-and-replay's docstring frames itself as a general "reconcile against what changed" mechanism, which could read as safer than it is on this specific point.
**Suggestion:** Add a one-line note next to the diff-and-replay block (or in ADR-104's Trade-offs section) stating explicitly that an `M` status always takes `$PREV`'s content wholesale — same semantics as the old allowlist, not new behavior — so a future reader doesn't assume this mechanism reconciles a concurrent `origin/main` edit to the same path.

---

### Findings Disposition

Non-blocking documentation note above — fix in-PR (a one-line addition to ADR-104's existing Trade-offs section) or file-and-link per ADR-028.

---

PR #748, opened 2026-07-13
Block worktree-squatting of engineering-journal's draft/YYYY-MM-DD branch

### LINKED ISSUE #630: Scheduled-task worktree provisioning hijacks the engineering-journal canonical checkout's HEAD — reproduced 2 days running

brownm09:
## Summary

Tile-spawned sessions whose own primary repo is a different project (finishing real work
elsewhere, then writing an engineering-journal stub as a side effect) have been creating their
own throwaway worktree for engineering-journal and checking out `draft/YYYY-MM-DD` **inside it**
instead of via `-C` against the canonical, as the documented Stub file workflow requires. Git
only allows one worktree to hold a branch at a time, so this locks the shared branch away from
the canonical and every other concurrent session. Confirmed live and reproduced twice on
2026-07-12: `.claude/worktrees/stub-823-120134` held `draft/2026-07-11`, `.claude/worktrees/stub-829-165612`
held `draft/2026-07-12` (that day's, actively blocking every other session) — simultaneously.

- New blocking `PreToolUse(Bash)` hook (`pre-tool-use-journal-draft-worktree-guard.py`) — rejects
  `git worktree add`/`checkout`/`switch` of a `draft/YYYY-MM-DD`-shaped branch anywhere except the
  canonical, with an `ALLOW_JOURNAL_DRAFT_WORKTREE=1` override.
- Generalizes `_worktree_topology.py`'s existing `main`-squat "park, don't delete" pattern
  (ADR-058) to an arbitrary branch pattern (`DRAFT_BRANCH_RE`, `non_canonical_worktrees_matching`,
  `pattern_squat_action`).
- Wires auto-park/remove into `prune-merged-worktrees.py`: an idle, clean, fully-pushed squatter
  is parked *and* removed; a dirty or not-provably-pushed one is parked only, contents left
  untouched for review.
- Corrects four places (`CLAUDE.md`'s mutate-guard exemption bullet, the mutate-guard script's
  own comment, ADR-093's References, root `README.md`'s hooks table) that cited dev-env#346 as if
  it covered this — it doesn't; #346 is narrowly scoped to `biweekly-retro`'s own report-writing
  step.
- Adds a comment on the live incident to closed #630 (not reopened — its own root-cause
  conclusion still holds) with fresh evidence the canonical-hijack pattern also hits regular
  interactive session provisioning, not just scheduled tasks.

See [ADR-105](docs/adr/105-draft-branch-worktree-squat-guard.md) for the full design, judgment
calls, and the five `/review`-found gaps fixed before merge (see Review findings disposition
below).

Closes #747.

## Testing

Full suite run via `py -3 claude/scripts/run-hook-tests.py`:

```
Tests: 65 passed, 2 skipped, 0 failed (81.5s)
```

The 2 skips are pre-existing and environment-related (`test_pyw_stdio.py` needs a real console;
`run-shellcheck.sh` needs shellcheck installed locally) — unrelated to this change.

`test_journal_draft_worktree_guard.py` (new, 25 cases — pure extraction-function coverage +
subprocess end-to-end coverage against real throwaway git repos, mirroring
`test_canonical_mutate_guard.py`'s two-layer convention). `test_worktree_topology.py` extended
(+8 cases for `DRAFT_BRANCH_RE`/`non_canonical_worktrees_matching`/`pattern_squat_action`).
`test_prune_merged_worktrees.py` extended (+4 cases for the park-and-remove / park-only /
remove-fails / remove-times-out integration paths, with call-recording assertions added after
`/review` found the original two cases didn't verify `git worktree remove` was actually invoked).

Also ran and confirmed clean: hook-script syntax check (item 1), suppression grep, test-integrity
grep, `settings.json` JSON validation. No deleted tests, no lockfile changes (no `package.json`
touched).

Every `/review` finding was independently re-verified by sabotaging the underlying logic and
confirming the relevant test actually failed, then restoring the fix and confirming it passed
again — not just re-reading the diff. See Review findings disposition below for the full list.

## Live recovery (separate from this PR)

The actual 2026-07-12 incident — both squatting worktrees, plus this session's own orphaned
worktree registration (a related but distinct dev-env#630-class issue, commented separately) —
was already resolved directly against engineering-journal's canonical, before this PR. Not part
of this diff (draft-branch commits in that repo don't go through PR review).

`/review` ran two parallel passes (correctness/security; reliability/performance/maintainability)
against the initial diff. All findings fixed in this PR, in `38c1c6d`:

1. **[correctness]** Tautological test for the hook's one legitimate case (`-C <journal>` via a
   Windows backslash path silently defeated `shlex.split(posix=True)`, so the test passed
   regardless of whether the canonical-matching logic was correct) — fixed in `38c1c6d` (render
   the `-C` path with forward slashes, matching the documented production convention).
2. **[correctness]** A leading `cd` fully exempted `git worktree add ... draft/YYYY-MM-DD` from
   the block, reproducing the exact incident this PR fixes — fixed in `38c1c6d` (removed the
   cd-scope guard from `find_worktree_add_blocks`, which needs no cwd at all; kept correctly on
   `find_checkout_candidates`).
3. **[correctness]** The park-and-remove test asserted only `(pruned, skipped)` counts, identical
   to park-only's, so a regression that stopped removing entirely would pass — fixed in `38c1c6d`
   (call-recording dispatcher + two new tests for remove failing/timing out after a successful
   park).
4. **[correctness]** A trailing `git checkout <branch> --` with no pathspec still switches
   branches but was wrongly exempted as a file restore — fixed in `38c1c6d`.
5. **[correctness]** `git worktree add --detach <draft-branch>` was a false-positive over-block
   (a detached checkout holds no branch ref) — fixed in `38c1c6d`.
6. **[maintainability]** `non_canonical_worktrees_matching` was imported but never used in
   `prune-merged-worktrees.py` (the check was reimplemented inline) — fixed in `38c1c6d`
   (precomputed once before the loop, mirroring `main_squatter`).
7. **[performance]** The new hook's redirect resolution dropped the sibling hook's
   `toplevel_cache` memoization (a dev-env#576/PR#584 hardening fix) — fixed in `38c1c6d`.
8. **[maintainability/reliability]** No test coverage for the park-and-remove failure/timeout
   path — fixed in `38c1c6d` (folded into finding 3's new tests).
9. **[documentation]** Root `README.md`'s hooks table had the same dev-env#346 mis-citation
   already corrected in three other places, and was missing an entry for the new hook — fixed in
   `38c1c6d`.

No findings filed-and-deferred; everything addressed in-PR.


brownm09:
## Review: Block worktree-squatting of engineering-journal's draft/YYYY-MM-DD branch

### Summary

Well-constructed fix for a confirmed, reproduced bug. Two parallel adversarial passes (correctness/security; reliability/performance/maintainability) found 8 real, non-overlapping gaps — none were architectural, all were fixable in-PR. The most severe: the single test protecting the hook's most important invariant (the legitimate `-C <canonical>` case) was tautological due to a Windows-backslash/`shlex` interaction, giving zero regression protection for the one thing that must never break. All 8 findings were fixed and independently re-verified via sabotage-then-reconfirm (break the logic, confirm the relevant test fails, restore, confirm it passes again) before this comment was posted — see the PR body's "Review findings disposition" section for exact commit references. Diff size (1506 lines) is appropriately scoped for one cohesive fix (hook + shared-module generalization + wiring + tests + docs), not a candidate for splitting.

---

### Blocking Findings

**[correctness]** `claude/scripts/tests/test_journal_draft_worktree_guard.py` (`test_main_allows_redirect_at_journal_canonical`)
The test built its `-C <journal>` command using `str(Path(...))`, which renders Windows backslashes. `_tokenize()`'s `shlex.split(rest, posix=True)` treats backslash as an escape character and silently strips every one, mangling the path into a non-existent directory — `_resolve_git_toplevel` then returns `None` and the hook fails open (exit 0), the same exit code the test expects for the *correct* reason. Sabotaging `_is_journal_canonical` to always return `False` — which in production would block the documented daily workflow for every session — still passed the full 21-case suite. This is the one test protecting the hook's single most important invariant (distinguish the legitimate case from every squat), and it gave zero regression protection.
**Fix:** Render the `-C` path with forward slashes, matching the documented production convention (`C:/Users/brown/Git/engineering-journal`).

**[correctness]** `claude/scripts/pre-tool-use-journal-draft-worktree-guard.py` (`find_worktree_add_blocks`)
The `cd`-takes-command-out-of-scope guard was applied identically to both extractors, but `find_worktree_add_blocks` needs no cwd resolution at all — it blocks purely on the branch-name token. `cd C:/Users/brown/Git/engineering-journal && git worktree add .claude/worktrees/foo draft/2026-07-12` bypassed the block entirely (confirmed: exit 0), reproducing the exact incident this PR exists to prevent, and a matching test actively encoded the gap as expected behavior.
**Fix:** Remove the `cd`-scope guard from `find_worktree_add_blocks` specifically; keep it on `find_checkout_candidates`, which genuinely can't resolve a target after a `cd`.

**[correctness]** `claude/scripts/tests/test_prune_merged_worktrees.py` (`test_draft_branch_squat_park_and_remove`) / `claude/scripts/prune-merged-worktrees.py` (the new draft-squat branch's `except subprocess.TimeoutExpired` / `if rm.returncode != 0` paths)
The park-and-remove test asserted only `(pruned_count, skipped_count)`, which are identical for park-and-remove and park-only (both append to `pruned`; only the primary is skipped in the happy path). Sabotaging `pattern_squat_action` to always return park-only still passed the "park and remove" test. Separately, the `git worktree remove` failure/timeout degradation paths (new in this PR) had no coverage at all.
**Fix:** Record every dispatched subprocess call and assert `git worktree remove` was (park-and-remove) or was not (park-only) actually invoked; add dedicated tests for the removal failing (non-zero exit) and timing out after a successful park.

**[correctness]** `claude/scripts/pre-tool-use-journal-draft-worktree-guard.py` (`find_checkout_candidates`)
`git checkout <branch> --` with nothing after the `--` still switches branches (confirmed against real git behavior) but was exempted as a file restore, since the code only checked for `--`'s presence, not whether a real pathspec followed it.
**Fix:** Only exempt when `--` is not the last token (i.e., a real pathspec follows).

**[correctness]** `claude/scripts/pre-tool-use-journal-draft-worktree-guard.py` (`_worktree_add_target`)
`git worktree add <path> --detach draft/2026-07-12` was blocked, but `--detach` checks out a detached HEAD — it holds no branch ref at all, so it can never actually be a squat. Over-blocking is the safe direction and no current flow uses this pattern, so this was latent rather than live.
**Fix:** Treat `--detach` as yielding zero candidates.

**[documentation]** `README.md` (PreToolUse hooks table)
The existing row for `pre-tool-use-canonical-mutate-guard.py` stated the engineering-journal carve-out is "temporary... pending dev-env#346" — the same inaccuracy already corrected in three other places by this PR (`claude/CLAUDE.md`, the mutate-guard script's own comment, ADR-093's References), but missed here. The table was also missing an entry for the new hook, unlike its closest sibling.
**Fix:** Correct the citation; add a row for `pre-tool-use-journal-draft-worktree-guard.py`.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/prune-merged-worktrees.py` (import block)
`non_canonical_worktrees_matching` was imported from `_worktree_topology` but never called — the draft-squat branch re-implemented the same check inline via `DRAFT_BRANCH_RE.match(branch or "")`, inconsistent with how the parallel main-squatter feature precomputes `main_squatter(worktrees)` once before the loop.
**Suggestion:** Precompute the draft-squatter set via `non_canonical_worktrees_matching(worktrees, DRAFT_BRANCH_RE)` before the loop, mirroring `main_squatter`.

**[performance]** `claude/scripts/pre-tool-use-journal-draft-worktree-guard.py` (`_resolve_checkout_target`)
The copied low-level git-parsing helpers are byte-identical to the sibling hook (verified, no drift), but the sibling's `toplevel_cache` memoization — added as a dev-env#576/PR#584 review fix for a documented "15 subprocesses for one command" worst case — was not carried over into this hook's own redirect-resolution wrapper, so a command with several repeated draft-branch checkout segments would re-spawn `git rev-parse` once per segment instead of once per distinct target.
**Suggestion:** Thread a shared `toplevel_cache` dict through `_resolve_checkout_target` across all candidates in one command, mirroring the sibling's `_blockable_redirect_root`.

---

### Findings Disposition

All 8 findings fixed in-PR, commit `38c1c6d`. See the PR body's "Review findings disposition" section for the itemized list and exact fix description per finding. No findings filed-and-deferred.

---

PR #752, opened 2026-07-13
feat(hooks): add heartbeat/liveness ledger for all wired hooks (PR8 of #717)

brownm09:
## Summary

Phase D (PR8) of the #717 hook-reliability initiative. Sub-issue: #745.

- Adds `_hookutil.record_heartbeat(hook_name)` — writes `~/.claude/scratch/hook-heartbeat/<hook_name>.ts` (current Unix timestamp) via a tmp-file + `os.replace` atomic swap, best-effort (swallows I/O errors). Called as the **unconditional first statement of `main()`** by all 41 currently-wired hook scripts, plus the new `hook-liveness-check.py` itself (42 call-sites total) — never at module level, so a test file's bare `import` of a hook module never pollutes the ledger.
- Adds `claude/scripts/hook-liveness-check.py` (new `UserPromptSubmit` hook): parses `claude/settings.json`'s `hooks` block to discover the currently-wired set (`wired_hook_events`), reads each non-exempt hook's heartbeat file (`stale_hooks`), and warns via `_hookout.emit_advisory("UserPromptSubmit", ..., audience="model")` when one or more have gone quiet longer than `DEFAULT_CADENCE_DAYS = 7`. Exempt: hooks wired **exclusively** to `PostCompact`/`Notification` (rare-firing events where long silence is expected). **Self-checks its own wiring and is debounced to once per session** — see Review findings disposition below.
- This is the mechanism that would have surfaced `post-tool-use.py`'s months-long silent death (dev-env#377) and `usage-snapshot.py`'s 8-day gap (dev-env#355) — the output-contract gates (ADR-103) verify a hook's code is *correct*; this verifies it is *running*.
- **New ADR-106** documents the design (call-site placement rationale, exemption-by-event-set, the deliberate non-goal of consolidating settings.json-parsing with `tests/_hook_wiring.py` — reserved for Phase E). Amended with a dated addendum documenting the review-driven refinements below.
- Doc-reconciliation: README.md + REFERENCE.md Hooks sections/tables updated; new Testing items 67-68 (`test_hook_liveness_check.py`, `test_hook_heartbeat_guard.py`); item 27 extended for `record_heartbeat` coverage.

## Testing

Ran `py -3 claude/scripts/run-hook-tests.py` three times (before opening the PR, after the doc-reconciliation edits, and again after the review-fix commit) — all green, final run:

```
Tests: 67 passed, 2 skipped, 0 failed (86.8s)
```

The 2 skips are pre-existing, documented environment skips (`test_pyw_stdio.py` needs a real Windows console; `run-shellcheck.sh` needs the `shellcheck` binary) — unrelated to this change.

New/extended coverage:
- `test_hookutil.py` — 6 new tests for `record_heartbeat`: writes a parseable recent timestamp, creates the heartbeat dir if absent, overwrites on a second call, leaves no leftover tmp file, swallows errors when the directory can't be created, and the default dir is `SCRATCH / "hook-heartbeat"`.
- `test_hook_liveness_check.py` (39 tests) — pure-helper coverage for `hook_name_from_command`, `wired_hook_events` (incl. malformed-structure degradation), `exempt_hooks` (incl. the mixed-event non-exemption case), `stale_hooks` (incl. the exact-cadence boundary), `_age_desc`, `format_warning`, and `format_self_check_failure` (ASCII-safety pinned); a cross-check against `tests/_hook_wiring.py` on the real settings.json; and 6 end-to-end subprocess tests covering the healthy path, a stale-hook warning, both self-check-failure paths, and the once-per-session debounce (same session silenced, different session still fires).
- `test_hook_heartbeat_guard.py` (new, 11 tests) — a structural AST gate asserting every wired hook calls `_hookutil.record_heartbeat("<own-name>")` as the literal first statement of `main()`, with an empty allowlist from day one.

Suppression grep, test-integrity grep, deleted-test-file check, and crude-command-substring AST gate (item 39, part of the full suite above): all clean.

## Notes for reviewers

- The 41-file sweep (one `_hookutil.record_heartbeat("<name>")` line + `import _hookutil` where missing) was done via a small one-off AST-based transformation script, then reviewed file-by-file via the full diff before committing. One bug was caught and fixed during that review: the script's alphabetical sibling-import heuristic misclassified `from __future__ import annotations` as a "sibling" module (since `"__future__".startswith("_")`), misplacing the new `import _hookutil` line ahead of `import _winsubp` in 3 files (`awake-blocker.py`, `journal-shard-write-advisory.py`, `turn-count-hook.py`). Fixed by hand before commit — the final diff places `import _hookutil` correctly in all 41 files.
- No existing hook's behavior, output, or exit code changes — verified by the full offline suite passing unchanged, including the AST safe-exit-guard and output-contract gates (ADR-103 PR3), which treat the added heartbeat line as inert.

`/review` ([comment](https://github.com/brownm09/dev-env/pull/752#issuecomment-4954774132)) found correctness/security clean and 4 reliability/performance/maintainability findings (1 blocking, 3 non-blocking). All fixed in-PR, commit [`a697663`](https://github.com/brownm09/dev-env/commit/a697663):

1. **[reliability, blocking]** `hook-liveness-check.py` could go silently dark on its own settings.json parse failure (its heartbeat call already ran, so its own ledger entry stayed fresh regardless) — fixed via a self-check that emits a distinct advisory when this hook's own name is missing from its parsed wired-hook set, instead of a silent exit. 2 new e2e tests.
2. **[maintainability]** No structural enforcement that a wired hook calls `record_heartbeat` correctly — fixed via a new AST gate, `test_hook_heartbeat_guard.py` (11 tests, empty allowlist from day one), which immediately caught `hook-liveness-check.py`'s own first draft using a variable instead of a literal in its heartbeat call.
3. **[maintainability]** No regression guard against `hook-liveness-check.py`'s settings.json parser drifting from `tests/_hook_wiring.py`'s near-identical one — fixed via a new cross-check test running both against the real `claude/settings.json`.
4. **[performance]** The full check (settings.json parse + ~40 heartbeat-file reads) ran on every prompt — fixed via a once-per-session debounce, matching `reconcile-open-prs.py`'s established sentinel pattern. 2 new e2e tests.

ADR-106 amended with a 2026-07-13 addendum documenting all four. Full suite re-verified green after the fix commit (see Testing above).

Closes #745


brownm09:
## Review: feat(hooks): add heartbeat/liveness ledger for all wired hooks (PR8 of #717)

### Summary

Well-scoped and appropriately large for what it does (51 files, but 41 are a mechanical 1-3 line sweep; the real logic lives in 5 files). Correctness and security are clean — the atomic heartbeat write, all 42 heartbeat call-site placements, the settings.json parsing, the boundary math, and the `_hookout` channel usage were all directly verified against the real files, not just the diff. Reliability/performance/maintainability review surfaced one real gap worth calling **blocking**: the new liveness-checker hook could itself go silently dark on a broken settings.json parse, with its own heartbeat staying "fresh" throughout — an instance of the exact failure class this PR exists to close, reintroduced in the tool meant to catch it. Three non-blocking findings (a missing regression guard against parser drift, no structural enforcement of the heartbeat-call invariant going forward, and no discussion of added per-prompt latency) round out the review. **All four have already been fixed in a follow-up commit on this branch, verified by 50 new/extended tests, before this review was posted** — see Findings Disposition below. Diff size and scope: appropriately sized, no split needed.

---

### Blocking Findings

**[reliability]** `claude/scripts/hook-liveness-check.py:161-172` (original `main()`)

The hook called `_hookutil.record_heartbeat("hook-liveness-check")` first (correct), then on a `SETTINGS_PATH.read_text()`/`json.loads()` failure silently `sys.exit(0)`'d with no distinct signal. Because the heartbeat call already ran, this hook's own ledger entry stayed perfectly fresh even while the mechanism it implements was doing nothing — exactly the "silently dead, discovered only by accident" failure class this whole PR exists to close (`post-tool-use.py` dead for months, dev-env#377; `usage-snapshot.py` dark for 8 days, dev-env#355), reintroduced in the one tool meant to catch it, in the one place least able to self-detect it. Compounding this, `wired_hook_events` degrading to an empty or wrong result (e.g. from an unexpected but valid-JSON settings.json shape) would also silently no-op with the same symptom.

**Fix:** Add a self-check floor — after parsing, verify this hook's own name appears in the parsed wired-hook set (or some other sanity floor), and fail loudly (a distinct advisory) rather than silently, mirroring `run-hook-tests.py`'s existing `suite_discovery_error` precedent for exactly this shape of gap.

---

### Non-Blocking Findings

**[maintainability]** `claude/scripts/hook-liveness-check.py:66-92` (`wired_hook_events`) vs. `claude/scripts/tests/_hook_wiring.py:58-115` (`wired_script_events`)

These are two structurally identical settings.json walkers — the same nested `isinstance` defensive chain, differing only in whether the regex capture includes the `.py` suffix. The ADR's deferral of consolidation to Phase E is a real, stated decision, but nothing catches the two silently drifting apart in the meantime (e.g. a future settings.json schema change applied to one and not the other).

**Suggestion:** At minimum, add a test asserting `hook-liveness-check.wired_hook_events()` and `_hook_wiring.wired_script_events()` agree (modulo the `.py` suffix) against the same input, so drift is caught even before Phase E lands.

---

**[maintainability]** PR body / diff (no file — process gap)

The PR body states the 41-file sweep was done via "a small one-off AST-based transformation script," but no such script appears anywhere in the diff, and neither the ADR nor the PR body documents a repeatable procedure for wiring a 42nd hook into this pattern correctly. That leaves the invariant ("every wired hook heartbeats correctly") as tribal knowledge to be reinvented — or done by risky manual copy-paste — the next time a hook is added.

**Suggestion:** Either commit the transformation script, or (stronger, and consistent with how this repo already enforces the safe-exit-guard and output-contract shapes structurally rather than by convention) add an AST-based lint mirroring `test_hook_safe_exit_guard.py` that fails when a wired hook's `main()` doesn't call `_hookutil.record_heartbeat` with its own correct name as the first statement.

---

**[performance]** `claude/settings.json` (new `hook-liveness-check.py` registration) + `claude/scripts/hook-liveness-check.py:105-135` (`stale_hooks`)

This adds a 13th script to the `UserPromptSubmit` group, each paying its own `pyw -3` interpreter startup on every single prompt in every session (a standing concern this repo tracks separately, dev-env#715). `stale_hooks` additionally opens and reads one small file per non-exempt wired hook (~40 today) on top of the settings.json parse, all synchronously on the critical path before Claude begins processing the prompt. Neither the PR body nor the ADR discusses this added per-prompt cost or considers debouncing.

**Suggestion:** Debounce to once per session (or once per N minutes) via a sentinel file, matching `disk-space-check.py` / `reconcile-open-prs.py`'s established per-session-marker pattern — the 7-day staleness cadence has no meaningful per-prompt resolution anyway.

---

### Findings Disposition

All four findings were fixed in-PR, in a follow-up commit (`a697663`) pushed to this branch before this review was posted:

1. **[reliability] Self-check gap** — fixed in `a697663`. Added a self-check after the settings.json parse: if `"hook-liveness-check"` is not present in the parsed `wired_hook_events()` result, emits a distinct `format_self_check_failure()` advisory (`audience="model"`) instead of a silent exit. Covers both the JSON-parse-failure sub-case and the parse-succeeds-but-degrades sub-case. Two new end-to-end subprocess tests (`test_e2e_unreadable_settings_emits_self_check_failure`, `test_e2e_own_wiring_missing_emits_self_check_failure`) drive both paths via a new `HOOK_LIVENESS_SETTINGS_PATH` test seam.
2. **[maintainability] Parser drift guard** — fixed in `a697663`. Added `test_wired_hook_events_agrees_with_hook_wiring_module`, which runs both `wired_hook_events()` and `_hook_wiring.wired_script_events()` against the **real** `claude/settings.json` and asserts agreement (modulo the `.py`-suffix convention).
3. **[maintainability] No structural enforcement / undocumented sweep** — fixed in `a697663`, via the stronger option. New `claude/scripts/tests/test_hook_heartbeat_guard.py` (11 tests), an AST gate mirroring `test_hook_safe_exit_guard.py`'s pattern, asserting every currently-wired hook calls `_hookutil.record_heartbeat("<own-name>")` as the literal first statement of `main()`. Ships with an empty allowlist from day one (no pre-existing debt — this PR made everything compliant in the same change). This gate immediately caught a real regression in its own introducing commit: `hook-liveness-check.py`'s first draft used a module-level `OWN_HOOK_NAME` variable instead of a literal in its own heartbeat call, inconsistent with all 41 other hooks — fixed on the spot.
4. **[performance] No debounce** — fixed in `a697663`. Added a once-per-session debounce via a `_hookutil` sentinel file, matching `reconcile-open-prs.py`'s established `_already_ran(session_id)` / `_mark_done(session_id)` pattern (with the same synthetic-fallback-key handling for a missing `session_id`). Two new end-to-end tests confirm a second call in the same session is silenced while a different session's call still fires.

ADR-106 was amended with a dated addendum (2026-07-13) documenting all four fixes and their rationale. Full offline suite re-verified green after the fix commit: `py -3 claude/scripts/run-hook-tests.py` → 67 passed, 2 pre-existing environment skips, 0 failed.

---

